Issue #26, opened 2026-07-31
Pre-publish review of the whole repository at 74d2647 — index


JPHutchins:
## Status 2026-08-30

Everything the review at `74d2647` found is shipped and merged except the items below. The five segfaults are fixed and pinned, the release pipeline has run through 50 wheels and an sdist, and the cp315 wheels are built and tested in CI while held out of the PyPI release until 3.15.0rc1 (#33's ruling).

## Still open

| # | what | why it stays open |
| --- | --- | --- |
| #13 | pickle and copy | JPH: not wanted yet; this index closes when it lands |
| #2 | `object.__setattr__` on a frozen struct, 3.10–3.12 | design ruling 2026-08-30: keep open, decide later |
| #1 | version `0.0.0` and the PyPI Trusted Publisher | release-time; the Trusted Publisher half is the web UI |
| #3 | `kw_only` | declined |
| #36 | PEP 487 `__init_subclass__` keywords | in flight — PR #157 |

---

<details>
<summary>Original index (74d2647)</summary>

Tracking issue for the whole-repository review run at `74d2647`, before any PyPI upload. It found **five reproducible segfaults** and **a release pipeline that has never run and cannot succeed as written**.

Nothing here is fixed. Every finding was reproduced by hand before it was filed; the ones that could not be reproduced say so.

## Work order

Grouped into one-PR batches, **simplest first** rather than by severity. `src/meta.c` is the contention point — ten issues touch it — so the batches that do must be serialized against each other and against **#27** (in flight, @Guflly's fix for #11). Everything outside `src/*.c` runs in parallel with anything.

| # | batch | issues | touches | note |
| --- | --- | --- | --- | --- |
| 1 | Stub declarations | #34 | `salix/__init__.pyi`, `tests/typing/` | smallest thing here |
| 2 | Two crash fixes | #6, #30 | `hash.c`, `compare.c` | ~7 lines, two segfaults gone; no `meta.c` |
| 3 | Metadata and repo files | #17, #25 | `pyproject.toml`, new root files, `bench/` | unblocks batch 10 |
| 4 | Release pipeline | #16, #22, #35 | `.github/`, `tools/check_tag.py` | **#16 blocks #1** |
| 5 | Class-creation bugs | #8, #10, #21 | `meta.c`, `fields.c` | **after #27**; #8 and #21 edit the same function |
[...]

PR #39, opened 2026-07-31
fix: two segfaults reachable from ordinary Python

### LINKED ISSUE #6: hash() of a self-referential struct segfaults: tp_hash has no recursion guard
`hash()` of a struct that contains itself **segfaults the interpreter**. There is no recursion guard on `tp_hash`, so the C stack overflows before Python's recursion limit can fire. A frozen dataclass in the same shape raises `RecursionError`.

```console
$ python -c '
import salix
class N(salix.Struct):
    child: object = None
n = N(); salix.set_field(n, "child", n)
hash(n)'
Segmentation fault (core dumped)      # exit 139
```

The cycle is reachable through the *supported* API — `set_field` in `__post_init__` is the documented way to derive a frozen field — and a two-node cycle crashes identically. `repr()` is guarded and `==` is guarded; `hash()` is the one unguarded entry point.

| operation | self-referential struct | frozen dataclass |
| --- | --- | --- |
| `repr(n)` | `N(child=...)` | `N(child=...)` |
| `n == m` | `RecursionError` | `RecursionError` |
| `hash(n)` | **SIGSEGV** | `RecursionError` |

<details>
<summary><b>Why it crashes, and why the other two do not</b></summary>

`Struct_hash` builds a tuple of the field values and hands it to `PyObject_Hash` — [src/hash.c:16](https://github.com/JPHutchins/salix/blob/74d2647/src/hash.c#L16):

```c
Py_hash_t Struct_hash(PyObject * const self) {
	StructType const * const type = struct_type_of(self);
	PyObject * const values = PyTuple_New(type->struct_field_count);
	...
	Py_hash_t const hash = PyObject_Hash(values);
```

When a value *is* `self`, `tuplehash` calls `PyObject_Hash` on it, which dispatches straight back to `Struct_hash`. Nothing on that path touches a Python frame, so `Py_EnterRecursiveCall` is never reached and the recursion limit never applies. It is a pure C recursion that runs until the stack is gone.

The other two protocols are guarded, which is why they degrade instead of crashing:

- `repr` calls `Py_ReprEnter` / `Py_ReprLeave` — [src/repr.c:15](https://github.com/JPHutchins/salix/blob/74d2647/src/repr.c#L15). That is what produces the `...`.
- `==` goes through `PyObject_RichCompareBool`, and CPython's `do_richcompare` enters the recursive call itself, so the interpreter raises before the stack does.

A dataclass escapes because its `__hash__` is a *Python* function: the frame push is the guard.

Measured on the 3.14 build in this tree; reproducible on every run.

</details>

<details>

`Py_EnterRecursiveCall` / `Py_LeaveRecursiveCall` around the body of `Struct_hash` converts the crash into the `RecursionError` a dataclass gives. It is two lines and needs no new state:

```c
if (Py_EnterRecursiveCall(" while hashing a struct") != 0) {
	return HASH_ERR;
}
...
Py_LeaveRecursiveCall();
```

Note it has to wrap the `PyObject_Hash(values)` call, not just the tuple build — the tuple build is not where the recursion happens.

Worth checking the same question for any future protocol that walks field values without going through a guarded CPython entry point. Today `repr`, `==`/ordering and `hash` are the complete set, and only `hash` is unguarded.

A regression test belongs in `tests/test_protocols.py`: build the cycle with `set_field` and assert `pytest.raises(RecursionError)`. It cannot be written as a crash test — a segfault takes the test runner with it — so the assertion has to be that the guarded error arrives.

</details>

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for a whole-repository code review in preparation for publishing salix to PyPI, with the findings filed as issues for future agents to resolve. This crash was found by probing the extension for recursion handling and reproduced three times.

### LINKED ISSUE #30: ordering_result reuses borrowed slot pointers across user code: use-after-free segfault
`ordering_result` takes two **borrowed** slot pointers, calls into arbitrary Python, and then uses those pointers again. If the user code drops the slot, the second use is a read of freed memory. **This segfaults on an ordinary GIL build** — no free threading required.

```python
import salix

class Evil:
    def __eq__(self, other):
        salix.set_field(holder, "a", 0)     # releases the value the C code is holding
        return False
    def __lt__(self, other): return True

class S(salix.Struct, order=True):
    a: object
    b: int

holder = S(Evil(), 1)
holder < S(Evil(), 2)
```

```console
comparing
Segmentation fault (core dumped)
```

Equality survives the same trick — `values_equal` returns as soon as the comparison answers and never touches the pointers again. Ordering is the one path that reuses them.

<details>

[src/compare.c:73-85](https://github.com/JPHutchins/salix/blob/74d2647/src/compare.c#L73-L85):

```c
	for (Py_ssize_t i = 0; i < self_type->struct_field_count; ++i) {
		PyObject * const mine = struct_slot_or_none(self_type, self, i);
		PyObject * const theirs = struct_slot_or_none(other_type, other, i);
		int const equal = PyObject_RichCompareBool(mine, theirs, Py_EQ);

		if (equal == COMPARISON_ERROR) {
			return NULL;
		}

		if (equal == COMPARISON_UNEQUAL) {
			return PyObject_RichCompare(mine, theirs, op);
		}
	}
```

`struct_slot_or_none` returns a borrowed reference ([src/types.h:64](https://github.com/JPHutchins/salix/blob/74d2647/src/types.h#L64)) — it reads the slot and hands back the pointer without incrementing. `PyObject_RichCompareBool` on line 76 can run any Python code, including code that rebinds the slot. When it returns `COMPARISON_UNEQUAL`, line 83 dereferences `mine` and `theirs` — which may now be the only-reference-dropped-and-freed objects.

Nothing about this depends on the GIL. The write path used above is `set_field`, but a mutable struct reaches it with a plain `self.a = 0`, and `__del__` on a field's value can do it too.

</details>

<details>
<summary><b>Scope: which paths are exposed</b></summary>

| function | takes borrowed slots | calls user code | reuses them after | crashes |
| `ordering_result` ([compare.c:73](https://github.com/JPHutchins/salix/blob/74d2647/src/compare.c#L73)) | yes | yes | **yes** | **yes** |
| `values_equal` ([compare.c:120](https://github.com/JPHutchins/salix/blob/74d2647/src/compare.c#L120)) | yes | yes | no | no |
| `Struct_hash` ([hash.c:24](https://github.com/JPHutchins/salix/blob/74d2647/src/hash.c#L24)) | yes | no — `Py_NewRef` into the tuple first | n/a | no |
| `field_repr` ([repr.c:60](https://github.com/JPHutchins/salix/blob/74d2647/src/repr.c#L60)) | yes | yes, `PyObject_Repr` | no | no |

`Struct_hash` is the model: it takes a strong reference into the tuple *before* anything can run. The fix for `ordering_result` is the same discipline — hold `mine` and `theirs` across the loop body, which `PY_OWNED` in [src/owned.h](https://github.com/JPHutchins/salix/blob/74d2647/src/owned.h) already expresses:

```c
		PY_OWNED(mine, Py_NewRef(struct_slot_or_none(self_type, self, i)));
		PY_OWNED(theirs, Py_NewRef(struct_slot_or_none(other_type, other, i)));
```

Two increments per field on the ordering path only. Equality and repr are safe as written but are one refactor from not being, so a comment recording *why* they are safe is worth as much as the fix.

Under a free-threaded build the exposure is wider — another thread can drop the slot between the read and any later use, with no user code involved. That is tracked in #7; this issue is the part that reproduces with the GIL on.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Closes #6. Closes #30.

Two crashes on a stock GIL build, both reproduced at exit 139 before the fix and gone after it.

| | before | after |
| --- | --- | --- |
| `hash(n)` where `n.child is n` | `Segmentation fault` (139) | `RecursionError: Stack overflow (used 8144 kB) while hashing a struct` |
| `<` where a field's `__eq__` rewrites the slot | `Segmentation fault` (139) | `True` |

### #6 — `tp_hash` had no recursion guard

`Struct_hash` builds a tuple and hands it to `PyObject_Hash`; when a value *is* the struct, `tuplehash` comes straight back in. Nothing on that path pushes a Python frame, so the interpreter's depth limit never sees the descent and the C stack goes first. A frozen dataclass survives the same shape only because its `__hash__` is a Python function — the frame push is its guard.

`Py_EnterRecursiveCall` around `PyObject_Hash`, not around the tuple build, which is not where the recursion is.

### #30 — ordering reused borrowed slots across user code

`ordering_result` took two borrowed pointers, ran `PyObject_RichCompareBool` on them, and then read both again on the unequal branch. The probe runs arbitrary Python; that Python can drop the slot. `PY_OWNED` holds both for the length of an iteration — two increments per field, ordering path only.

<details>
<summary><b>The neighbouring paths, checked rather than assumed</b></summary>

`ordering_result` is the only one of four that reuses the pointers:

| function | borrowed slots | runs user code | reuses after | crashes |
| `ordering_result` | yes | yes | **yes** | **yes** |
| `values_equal` | yes | yes | no | no |
| `Struct_hash` | yes | no — `Py_NewRef` into the tuple first | n/a | no |
| `field_repr` | yes | yes | no | no |

`values_equal` now carries a comment saying why its borrow is sound, because it is one refactor away from the bug next door.

</details>

<details>
<summary><b>Why the tests assert what they assert</b></summary>

Neither can assert a crash — a segfault takes pytest with it — so each asserts the guarded outcome:

- `test_hash_of_a_struct_that_contains_itself` builds the cycle with `set_field` and expects `RecursionError`. It asserts the type, not the message: 3.12+ reports `Stack overflow (used N kB)` where earlier versions say `maximum recursion depth exceeded`, and this passes on 3.10 through 3.15 plus 3.14t.
- `test_a_field_that_rewrites_its_own_slot_mid_comparison` is the #30 reproducer with an assertion on the answer.

A regression in either takes the test session down rather than failing a case. That is the cost of testing a use-after-free in-process, and it is loud.

`hash.c` also moves its tuple to `PY_OWNED`, which is what removes the second `Py_DECREF` the new early return would otherwise need.

`camas check` green at 25/25 — six interpreters, `clang-tidy`, `gcc -fanalyzer`, and the nix C tests.

</details>

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for the pre-publish review's issues to be resolved in batches, simplest first; this is batch 2 of 12, and #26 indexes the rest.

### DIFF
--- a/src/compare.c
+++ b/src/compare.c
@@ -2,6 +2,7 @@
 
 #include "compare.h"
 #include "mixin.h"
+#include "owned.h"
 #include "types.h"
 
 /* The tri-state PyObject_RichCompareBool speaks, named. */
@@ -52,6 +53,9 @@ static PyObject * equality_result(PyObject * const self, PyObject * const other,
  * compare in. Both classes have to have asked for ordering, because a class
  * that did not is not orderable, and being compared against one that did does
  * not change that.
+ *
+ * The values are owned rather than borrowed: the equality probe runs arbitrary
+ * Python, which can rebind either slot before the unequal branch reads them.
  */
 static PyObject * ordering_result(PyObject * const self, PyObject * const other, int const op) {
 	StructType const * const self_type = struct_type_of(self);
@@ -71,8 +75,8 @@ static PyObject * ordering_result(PyObject * const self, PyObject * const other,
 	}
 
 	for (Py_ssize_t i = 0; i < self_type->struct_field_count; ++i) {
-		PyObject * const mine = struct_slot_or_none(self_type, self, i);
-		PyObject * const theirs = struct_slot_or_none(other_type, other, i);
+		PY_OWNED(mine, Py_NewRef(struct_slot_or_none(self_type, self, i)));
+		PY_OWNED(theirs, Py_NewRef(struct_slot_or_none(other_type, other, i)));
 		int const equal = PyObject_RichCompareBool(mine, theirs, Py_EQ);
 
 		if (equal == COMPARISON_ERROR) {
@@ -111,6 +115,8 @@ static enum comparison names_equal(
 	);
 }
 
+/* Borrowed slots are sound here and not in ordering_result: the comparison is
+ * the last use of either pointer. */
 static enum comparison values_equal(
 	StructType const * const self_type,
 	PyObject * const self,
--- a/src/hash.c
+++ b/src/hash.c
@@ -1,6 +1,7 @@
 #include <Python.h>
 
 #include "hash.h"
+#include "owned.h"
 #include "types.h"
 
 /* CPython reserves -1: a real hash of -1 is remapped to -2. */
@@ -12,10 +13,13 @@ enum : Py_hash_t {
  * A struct hashes as the tuple of its values, so `hash(p) == hash((1.0, 2.0))`
  * and structs interoperate with tuple keys. Building the tuple is the whole
  * cost; PyObject_Hash does the rest.
+ *
+ * A struct that holds itself re-enters here through tuplehash without pushing
+ * a Python frame, so the interpreter's own depth limit never sees the descent.
  */
 Py_hash_t Struct_hash(PyObject * const self) {
 	StructType const * const type = struct_type_of(self);
-	PyObject * const values = PyTuple_New(type->struct_field_count);
+	PY_OWNED(values, PyTuple_New(type->struct_field_count));
 
 	if (values == NULL) {
 		return HASH_ERR;
@@ -25,8 +29,13 @@ Py_hash_t Struct_hash(PyObject * const self) {
 		PyTuple_SET_ITEM(values, i, Py_NewRef(struct_slot_or_none(type, self, i)));
 	}
 
+	if (Py_EnterRecursiveCall(" while hashing a struct") != 0) {
+		return HASH_ERR;
+	}
+
 	Py_hash_t const hash = PyObject_Hash(values);
-	Py_DECREF(values);
+
+	Py_LeaveRecursiveCall();
 
 	return hash;
 }
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -11,7 +11,7 @@ import weakref
 import pytest
 from values import EVERY, HASHABLE, identify
 
-from salix import Struct
+from salix import Struct, set_field
 
 
 class Plain(Struct):
@@ -169,6 +169,21 @@ class TestOrder:
 
         with pytest.raises(TypeError):
             _ = Holder(object()) < Holder(object())
+
+    def test_a_field_that_rewrites_its_own_slot_mid_comparison(self):
+        class Fickle:
+            def __eq__(self, other):
+                set_field(holder, "x", 0)
+                return False
+
+            def __lt__(self, other):
+                return True
+
+            __hash__ = None
+
+        holder = Ordered(Fickle(), 1)
+
+        assert holder < Ordered(Fickle(), 2)
 
     def test_it_is_inherited(self):
         class Child(Ordered):
--- a/tests/test_protocols.py
+++ b/tests/test_protocols.py
@@ -2,7 +2,7 @@
 
 import pytest
 
-from salix import Struct
+from salix import Struct, set_field
 
 
 class Point(Struct):
@@ -70,6 +70,17 @@ def test_equal_structs_share_a_hash_bucket():
 def test_an_unhashable_field_makes_the_struct_unhashable():
     with pytest.raises(TypeError, match="unhashable"):
         hash(Nested([1]))
+
+
+def test_hash_of_a_struct_that_contains_itself():
+    class Node(Struct):
+        child: object = None
+
+    node = Node()
+    set_field(node, "child", node)
+
+    with pytest.raises(RecursionError):
+        hash(node)
 
 
 def test_repr_round_trips_through_eval():


JPHutchins:
Force-pushed a comment trim. No behaviour change — the two fixes, the tests and the measurements are unchanged; `camas check` is green at 25/25.

What I had written into `hash.c` and `compare.c` was a second copy of the commit message. What is left is only the part the code cannot show:

```c
 * A struct that holds itself re-enters here through tuplehash without pushing
 * a Python frame, so the interpreter's own depth limit never sees the descent.
```

```c
 * The values are owned rather than borrowed: the equality probe runs arbitrary
 * Python, which can rebind either slot before the unequal branch reads them.
```

Both test docstrings are gone — `test_hash_of_a_struct_that_contains_itself` and `test_a_field_that_rewrites_its_own_slot_mid_comparison` already say what they assert, and the surrounding suite uses one short line or nothing.

For the record on the previous round: the review workflow run for `c34c0fd` (`30656998473`) failed with no step-level failure and no log, so the `comment` job was skipped and nothing landed here. This push starts a fresh CI and review rather than re-running a round that produced no output.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for a comment-style cleanup across the open pull requests, folded into each one's next round rather than pushed on its own.

Issue #32, opened 2026-07-31
-Werror ships in the sdist's compile flags, where a user cannot override it


JPHutchins:
The compile flags baked into the distribution include `-Werror`. A user installing from the sdist compiles salix with warnings-as-errors on **their** compiler, which nobody here has tested, and cannot turn it off.

Straight out of the published sdist, `salix-0.0.0/build_config.py`:

```python
    c_flags=(
        "-std=c2x",
        "-O2",
        "-Werror",
        "-Wdouble-promotion",
        "-Wall",
        "-Wextra",
        "-Wno-unused-parameter",
    ),
```

`setup.py` passes these as `extra_compile_args`, which land **after** anything in `CFLAGS`, so a user cannot override them — `CFLAGS="-Wno-error"` loses to a later `-Werror`.

Any warning from any compiler salix has not been built with turns `pip install salix` into a hard failure. GCC and Clang add warnings in every release; a new one in a future GCC breaks installation of an already-published sdist, on a machine that has no wheel.

<details>
<summary><b>Who is actually exposed</b></summary>

Not most people. 50 wheels cover CPython 3.10–3.15 on Linux, macOS and Windows, x86-64 and arm64, including free-threaded — so a normal `pip install` takes a wheel and never compiles. The sdist is what serves:

- musl/Alpine, which has no musllinux wheel in the matrix;
- any architecture outside the six built (ppc64le, s390x, riscv64);
- distribution packagers, who build from sdist by policy;
- `--no-binary` builds and hermetic build systems.

Those are the users who get a build with someone else's `-Werror`.

</details>

<details>
<summary><b>Shape of a fix</b></summary>

The flags exist for a real reason. `build_config.py` is described as the single source of truth precisely so *"a flag or a new translation unit lands in both"* the local build and the nix cross builds, and `-Werror` is load-bearing for CI — [tasks.py:55](https://github.com/JPHutchins/salix/blob/74d2647/tasks.py#L55) notes the flags carry it so gcc holds the build to a second compiler.

What is wanted is for it to be load-bearing *in this repo* and absent from the artifact. `BuildConfig` is already a `NamedTuple` with a `c_flags` field, so splitting it costs almost nothing:

```python
class BuildConfig(NamedTuple):
    sources: tuple[str, ...]
    c_flags: tuple[str, ...]          # what any build needs
    strict_flags: tuple[str, ...]     # -Werror and friends, development only
```

[...]

Issue #25, opened 2026-07-31
Repository hygiene before publishing: CHANGELOG, CONTRIBUTING, SECURITY, and a stale header comment


JPHutchins:
Small, entirely additive gaps between "a repository that works" and "a repository someone else can contribute to". Grouped because each is a few minutes and none is worth its own issue.

**Absent from the repository:**

| file | why it matters at publish time |
| --- | --- |
| `CHANGELOG.md` | The first release has nothing to point at. `Project-URL: Changelog` in #17 needs a target. |
| `CONTRIBUTING.md` | The build is `nix develop` + `uv run camas check` and nothing else works. That is unusual enough that a drive-by contributor will not guess it — the README's two lines are the whole of it today. |
| `SECURITY.md` | GitHub surfaces its absence on the repository page. For a C extension a one-line "open a private advisory" is enough. |
| `.github/dependabot.yml` | Nothing keeps the pinned actions current — and #22 proposes SHA-pinning them, which needs Dependabot to be maintainable. |
| issue templates | Optional. Worth it only if this attracts contributors. |

**Smaller items found while reading:**

`build_config.py:31` sets `-std=c2x`. C23 is final and both GCC 14+ and Clang 18+ accept `-std=c23`; `c2x` is the pre-ratification spelling and will eventually be dropped. Only worth changing once every compiler in the matrix — including the zig-bundled clang used for the cross builds — accepts the new name, so this needs checking rather than assuming.

Two `static` arrays of unknown size are declared at file scope as forward declarations — `src/meta.c:71` and `src/mixin.c:13`:

```c
static PyGetSetDef StructMeta_getset[];
```

C23 6.9.2 makes a tentative definition with internal linkage and an incomplete type a constraint violation. GCC and Clang accept it as an extension and the build does not use `-Wpedantic`, so nothing fires today — but CLAUDE.md asks for C23 and for compile-time provability, and a `static PyGetSetDef X[N];` or a prototype-free reordering costs nothing.

`bench/bench.py:158` imports `records` unconditionally although the dependency stops at 3.11, so the benchmark cannot run on a newer interpreter without it installed.

`src/salix.c`'s header comment opens with a stale identity and one stale claim:

```c
/* struct — a minimal, C-backed, inheritable ``Struct`` base class.
 *
 * This is the seed implementation copied from the feasibility prototype
 * (JPHutchins/record-type PR #1). ...
 *   - per-type vectorcall for instance creation (fields written straight to
[...]

Issue #7, opened 2026-07-31
set_field is not free-threading safe and segfaults on 3.14t


JPHutchins:
`set_field` writes a live instance's slot with a plain read-modify-write. Under a free-threaded interpreter that is a data race, and it **crashes**. salix ships 15 free-threaded wheels, so this is a promise the code does not keep.

```console
$ python3.14t -X gil=0 race.py        # 4 threads set_field, 4 threads read the same struct
gil enabled: False
Segmentation fault (core dumped)      # exit 139, reproduced on every run
```

The same load through CPython's own member descriptor survives — swap `set_field(shared, "value", v)` for `shared.value = v` on a `frozen=False` struct and the identical script prints `survived`. That isolates the race to `set_field` alone.

It also means the safety claim in [src/salix.c:41-47](https://github.com/JPHutchins/salix/blob/74d2647/src/salix.c#L41-L47) stopped being true at `74d2647`:

> Everything shared between threads is written once and then only read [...] **What writes a live instance is CPython's own member descriptor**, so a free-threaded build's guarantees there are inherited rather than reimplemented.

`set_field` is now a second writer, and it inherits nothing.

<details>
<summary><b>The measurement</b></summary>

Both runs on `cpython-3.14.6+freethreaded`, against the in-tree build, 8 threads × 200 000 iterations on one shared instance:

| write path | result |
| --- | --- |
| `salix.set_field(shared, "value", [tag, i])` | **SIGSEGV**, core dumped — 3/3 runs |
| `shared.value = [tag, i]` (`frozen=False`) | `survived` — exit 0 |

Readers in both runs do `v = shared.value` and `len(v)`.

</details>

<details>
<summary><b>Why one races and the other does not</b></summary>

salix writes the slot with `Py_XSETREF` — [src/construct.c:251](https://github.com/JPHutchins/salix/blob/74d2647/src/construct.c#L251):

```c
case FIELD_LOOKUP_FOUND:
	Py_XSETREF(*struct_slot(type, self, found.index), Py_NewRef(value));
```

`Py_XSETREF` is a plain load, a plain store and a decref (CPython 3.14, `Include/cpython/object.h:372`):

```c
#define Py_XSETREF(dst, src) \
    do { \
        PyObject **_tmp_dst_ptr = _Py_CAST(PyObject**, &(dst)); \
        PyObject *_tmp_old_dst = (*_tmp_dst_ptr); \
        PyObject *_tmp_src = _PyObject_CAST(src); \
[...]

PR #40, opened 2026-07-31
chore: give the distribution and the repository their metadata

### LINKED ISSUE #17: PyPI metadata has no classifiers, no project URLs and no keywords
The generated `METADATA` has no classifiers, no project URLs and no keywords. On the day salix is published, its PyPI page will carry no link back to this repository and no indication of which interpreters it supports.

Taken verbatim from the built wheel (`nix build .#packages.x86_64-linux."wheel-3.14-manylinux-x86_64"`):

```
Metadata-Version: 2.4
Name: salix
Version: 0.0.0
Summary: A C-backed, inheritable Struct base class for Python — fast import, fast type creation
Author-email: JP Hutchins <jphutchins@gmail.com>
License-Expression: MIT
Requires-Python: >=3.10
Description-Content-Type: text/markdown
License-File: LICENSE
Dynamic: license-file
```

That is the whole of it. No `Classifier:`, no `Project-URL:`, no `Keywords:`.

<details>
<summary><b>What is missing and why each one matters</b></summary>

| missing | consequence |
| --- | --- |
| `Project-URL: Repository` / `Issues` / `Changelog` | The sidebar is empty. A reader on PyPI has no way to reach the source, and the README's references to `src/salix.c`, `tasks.py` and `uv run camas benchmark` are unreachable — see the README issue. `nix/wheel.nix` already knows the URL. |
| `Programming Language :: Python :: 3.10` … `3.15` | PyPI's version facets and the "Meta" panel show nothing. This is the standard signal that a package supports a given interpreter. |
| `Programming Language :: Python :: Implementation :: CPython` | **The functional one.** salix depends on `PyHeapTypeObject`'s layout and on `__attribute__((cleanup))`. PyPy and GraalPy satisfy `requires-python = ">=3.10"`, find no matching wheel, fall through to the sdist, and fail to compile — the same reason `tasks.py:147` excludes MSVC. A classifier does not *stop* them, but it is the declared signal, and it is what tooling reads. |
| `Programming Language :: Python :: Free Threading :: 3 - Stable` | 15 of the 50 wheels are free-threaded. This is the classifier that advertises it. |
| `Typing :: Typed` | `py.typed` and the stub ship in every wheel; nothing says so on the page. |
| `Development Status ::` | Sets expectations for a 0.x. |
| `Intended Audience :: Developers`, `Topic ::`, `Keywords` | Discoverability. Nobody searching for "dataclass", "struct", "slots" or "fast import" will find this. |
| `License :: OSI Approved :: MIT License` | Deliberately absent, and correctly so — PEP 639 `License-Expression: MIT` supersedes it and PyPI rejects both together. No change wanted here. |

`version = "0.0.0"` is tracked separately in #1. Worth noting it is not free: `tasks.py:164` documents that a permanent `0.0.0` is what forces `--no-cache` on the `wheel_test` leg, because uv keys its cache on name and version.

</details>

<details>
<summary><b>What is already correct</b></summary>

Verified against the built artifacts, so the fix is additive only:

- `Metadata-Version: 2.4` with `License-Expression: MIT` and `License-File: LICENSE` — modern PEP 639, correct, and `setuptools>=77` is pinned for it.
- The wheel ships `salix/py.typed`, `salix/__init__.pyi` and `dist-info/licenses/LICENSE`.
- Zip entries are stamped `1980-01-01 00:00`, so the wheels are reproducible.
- Tag `cp314-cp314-manylinux_2_17_x86_64` matches the payload, and the payload has **zero non-Python undefined symbols** (`nm -D --undefined-only` returns 60 symbols, all `Py*`/`_Py*`), so the `manylinux_2_17` claim is satisfied with room to spare.
- The sdist builds from source with only `setuptools`, and contains every `.c` and `.h` it needs.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Closes #17. Partially addresses #25 — three of its items belong to batches that own those files; see below.

### PyPI metadata (#17)

`METADATA` had no `Classifier`, no `Project-URL` and no `Keywords`. On publication day the page would have carried no link to the source and no statement of which interpreters this supports. Generated from this branch:

```
Project-URL: Repository, https://github.com/JPHutchins/salix
Project-URL: Issues, https://github.com/JPHutchins/salix/issues
Project-URL: Changelog, https://github.com/JPHutchins/salix/blob/main/CHANGELOG.md
Keywords: struct,dataclass,slots,frozen,immutable,import-time,free-threading,c-extension
Classifier: Development Status :: 3 - Alpha
...
Classifier: Programming Language :: Python :: Implementation :: CPython
Classifier: Programming Language :: Python :: Free Threading :: 3 - Stable
Classifier: Typing :: Typed
```

Fifteen classifiers, each checked against the `trove-classifiers` package rather than recalled — an invalid one is rejected at upload, and two here are recent enough to doubt (`3.15`, and the free-threading one).

`Development Status :: 3 - Alpha` is the one judgement call; it is a one-line change if you would rather say Beta.

### Repository files (#25)

`CHANGELOG.md`, `CONTRIBUTING.md`, `SECURITY.md`, `.github/dependabot.yml`, and the `bench/` import fix.

<details>
<summary><b>Why the changelog is a scaffold and not a history</b></summary>

Nothing has been released, so there is no baseline to describe changes against. Reconstructing one commit by commit would produce a document that claims to be a release history and is not. It says that, and says where entries go once there is a version to hang them on. `Project-URL: Changelog` now has a target either way.

</details>

<details>
<summary><b>The bench fix, both directions run</b></summary>

`bench/bench.py:158` imported `records` unconditionally, although `record-type` is declared `python_version>='3.11'` and has no wheel for the newest interpreters. In an environment without it:

```
  File "bench.py", line 158, in build_constructors
    from records import record
ModuleNotFoundError: No module named 'records'
```

— before printing a single row. After, in the same environment, the row simply is not there:

```
construct                 import ms   us/type   inst ns
-------------------------------------------------------
salix (this project)          0.151       8.6      54.2
msgspec.Struct                9.926       9.6      61.2
typing.NamedTuple             1.602      70.6     137.6
dataclass (frozen+slots)      9.918     346.4     232.9
```

and with it installed, all five rows are back. Every row is a rival to measure against, not a requirement, so the filter is general rather than special-cased to `records`.

</details>

<details>
<summary><b>What is left in #25, and why it is not here</b></summary>

| item | where it goes | why |
| --- | --- | --- |
| `-std=c2x` → `-std=c23` | with #32 | Needs every compiler in the matrix to accept the new spelling, the zig-bundled clang included. That is a cross-build question, not a one-word edit. |
| tentative definitions at `src/meta.c:71`, `src/mixin.c:13` | the batch that owns those files | `meta.c` is where PR #27 is, and the struct-identity work touches both. |
| `src/salix.c` stale header comment | with #7 | The same comment carries a second stale claim — the free-threading one, which is #7. Fixing half a comment twice is worse than fixing it once. |

</details>

`camas check` green at 25/25.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for the pre-publish review's issues to be resolved in batches, simplest first; this is batch 3 of 12, and #26 indexes the rest.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# #22 pins the actions by commit SHA, which is the one pin nobody updates by
+# hand. One group, because a pull request per action is noise at this size.
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    groups:
+      actions:
+        patterns:
+          - "*"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,10 @@
+# Changelog
+
+The format is [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and
+salix follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+Nothing released yet; everything in `main` is the first release.
+
+[Unreleased]: https://github.com/JPHutchins/salix/commits/main/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,71 @@
+# Contributing
+
+## The environment
+
+```sh
+nix develop
+```
+
+That is not a convenience. `clang-tidy`, `gcc -fanalyzer`, `nixfmt` and
+`jphfmt` all come from the dev shell, the shell deliberately carries no Python
+so that `uv` supplies every interpreter the matrix names, and CI installs Nix
+and enters the same shell. A checkout without it can build the extension and
+little else.
+
+## The loop
+
+```sh
+uv run camas check      # the default task
+uv run camas ci         # exactly what CI runs
+```
+
+`camas` is one definition of the task tree shared by this repository and its
+workflows, so `ci` locally is the same tree GitHub runs, not an approximation
+of it. `uv run camas --list` names the rest; the ones you are most likely to
+want alone are `test`, `type_check`, `analyze` and `format`.
+
+`check` is the inner loop: six CPython versions plus free-threaded 3.14, the
+linters, the analyzers and the three type checkers. `ci` adds `nix flake
+check`, which builds the wheels and takes minutes.
+
+Run `uv run camas format` before committing, or `check` will tell you to.
+
+## What the tests are
+
+- `tests/` is pytest, and every supported interpreter runs all of it.
+- `tests/typing/accepted.py` and `tests/typing/rejected.py` are assertions
+  about mypy, pyright and ty rather than about runtime behaviour. `rejected.py`
+  asserts by inversion: each `# type: ignore` is a claim that the checker
+  reports something there, and `--warn-unused-ignores` makes a suppression that
+  stops being needed an error.
+- `README.md` is collected by pytest as doctests, so its examples are run and
+  not proof-read.
+- `src/*.c` carry in-file C tests, built and run by `nix build .#c-tests`.
+
+A fix for a crash needs a test that asserts the guarded outcome — an
+exception, or the right answer — because a segfault takes the test runner with
+it and proves nothing.
+
+## Style
+
+Follow the code already in the file you are editing. The C is formatted by
+`jphfmt` and there is no Python formatter on purpose; `ruff` is configured to
+find defects, not to have opinions about layout.
+
+The C targets C23 and builds with `-Werror -Wall -Wextra -Wdouble-promotion`.
+
+## Commits and pull requests
+
+Conventional prefixes — `feat:`, `fix:`, `build:`, `refactor:`, `style:`,
+`chore:`. The body is where reasoning belongs: what was measured, what was
+tried and rejected, and why. That is context the source cannot carry.
+
+Every pull request is reviewed automatically once CI finishes. The review runs
+from the default branch and comments on the pull request.
+
+Work authored by an LLM agent must say so, in the commit trailer and in
+anything it posts:
+
+```
+Co-Authored-By: <model-id> <noreply@<provider>>
+```
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,24 @@
+# Security
+
+## Reporting
+
+Report a vulnerability privately through GitHub's
+[security advisories](https://github.com/JPHutchins/salix/security/advisories/new).
+Please do not open a public issue for one.
+
+## Scope
+
+salix is a C extension. A crash reachable from ordinary Python — a segfault, a
+use-after-free, a read past the end of a field table — is in scope whether or
+not it is exploitable, because the memory unsafety is the bug and the
+exploitability is a detail. Report it here rather than as an issue if you are
+unsure which it is.
+
+Out of scope: passing deliberately hostile arguments to `ctypes`, to
+`object.__setattr__`, or to anything else that reaches an instance's memory
+without going through the API. `salix/__init__.pyi` is the API.
+
+## Supported versions
+
+Nothing has been released yet. Once it has, fixes land on the latest minor
+version.
--- a/bench/bench.py
+++ b/bench/bench.py
@@ -53,6 +53,13 @@ class Construct(NamedTuple):
     ctor: Callable[[], Callable[[int, int, int], object]]
 
 
+class Row(NamedTuple):
+    label: str
+    dep_ms: float
+    per_type_us: float
+    instantiate_ns: float
+
+
 def _struct(i: int) -> str:
     return f"class C{i}(Struct):\n    a: int\n    b: int\n    c: int\n"
 
@@ -126,8 +133,10 @@ def _record_type_ctor() -> Callable[[int, int, int], object]:
     return C  # type: ignore[no-any-return]
 
 
+# An unbuilt salix/ still holds py.typed and the stub, which makes it a
+# namespace package: find_spec answers, and the module it names is empty.
 def _installed(dep: str | None) -> bool:
-    return dep is None or find_spec(dep) is not None
+    return dep is None or ((spec := find_spec(dep)) is not None and spec.loader is not None)
 
 
 # A rival to measure against, not a requirement: record-type carries a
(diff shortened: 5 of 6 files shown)


JPHutchins:
Both findings taken, and they were the same defect: I put a general filter at the `CONSTRUCTS` level and a special case inside `build_constructors()`. Two places to remember is what was actually wrong, so the fix is structural rather than a third guard.

The constructor now lives on the `Construct` record, beside the dependency it needs:

```python
class Construct(NamedTuple):
    key: str
    label: str
    dep: str | None
    header: str
    body: Callable[[int], str]
    ctor: Callable[[], Callable[[int, int, int], object]]


def build_constructors() -> dict[str, Callable[[int, int, int], object]]:
    return {c.key: c.ctor() for c in CONSTRUCTS}
```

Each `ctor` does its own import when called, so a construct that did not survive the filter is never constructed and its dependency is never imported. One list, one filter, one place to add the next rival.

Finding 1 also noted the summary would raise `StopIteration` even if the import were guarded — correct, and that is fixed too. `rows` is now keyed by construct and the summary takes the intersection with the three it wants, so a filtered construct cannot reappear.

<details>
<summary><b>Run three ways</b></summary>

Everything installed — five rows, unchanged:

```
salix (this project)          0.163       9.0      55.8
msgspec.Struct               15.258       8.9      59.5
typing.NamedTuple             1.608      69.7     135.0
dataclass (frozen+slots)      8.863     347.9     235.9
record-type (@record)        10.055      95.8     239.9
```

Neither `record-type` nor `msgspec` — degrades instead of crashing, which is the case finding 1 says was still broken:

```
salix (this project)          0.142       9.2      58.1
typing.NamedTuple             1.647      71.4     139.3
dataclass (frozen+slots)      8.929     352.5     235.4

  N=1      salix    0.151 ms   typing.NamedTuple    1.718 ms
```

The summary line drops `msgspec` rather than dying on it.

</details>

Finding 2 (the redundant `_installed("records")`) disappears with the same change — there is no second call left.

Also folded in: a comment trim across this branch. What I wrote was commit-message prose in the source. The reasoning belongs in the commit, and a comment earns its place only by recording something the code cannot show.

`camas check` green at 25/25.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked that review findings and a separate comment-style cleanup be addressed in one round rather than two, so this force-push carries both.

JPHutchins:
Both taken. `camas check` green at 25/25.

**1 — `_installed("salix")` and the namespace package.** Confirmed exactly as described before fixing. `salix/` carries `py.typed` and the stub whatever happens, so an unbuilt checkout is a PEP 420 namespace package:

```
spec is not None -> True
spec.loader     -> None
AttributeError: module 'salix' has no attribute 'Struct'
```

The predicate now requires a loader, which is what "installed" was always meant to mean.

One consequence worth naming rather than burying: on an unbuilt checkout the benchmark now runs **without its own subject** instead of dying on it. Strictly better than a traceback, and it is what the finding asked for — but a missing salix row in salix's own benchmark is quiet. A hard error naming the build command is one line if that reads better; I did not add it because inventing a special case for salix is what the previous round told me not to do.

**2 — positional row tuple.** Taken. `Row` NamedTuple, matching `Construct` in the same file:

```python
class Row(NamedTuple):
    label: str
    dep_ms: float
    per_type_us: float
    instantiate_ns: float
```

`total()` reads `r.dep_ms`/`r.per_type_us` now, so a field added or reordered fails instead of silently changing the numbers.

Run both ways again: five rows with everything installed, and `gen_salix` dropping out cleanly on a checkout where the extension is not built.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, answering the automated review on this PR as part of a batch of pre-publish fixes she asked to be iterated with that reviewer.

JPHutchins:
All three taken. `camas check` green at 25/25.

**1 — `find_spec` proves findability, not importability.** This was the real one, and the reviewer's named case reproduces exactly: PyPI's `records` is a SQL library with no `record` in it.

```console
$ printf 'SQL_FOR_HUMANS = True\n' > records.py
$ PYTHONPATH=. python bench/bench.py
ImportError: cannot import name 'record' from 'records'
```

Not one row printed. Two mechanisms knew whether a construct was usable and only one of them was right — which is the same shape as the defect round 1 fixed, so the fix is again structural rather than a third guard. **Building the constructor is the availability test**, because there is no cheaper question with the correct answer:

```python
def build_constructors() -> dict[str, Callable[[int, int, int], object]]:
    built = ((c.key, _constructor(c)) for c in CONSTRUCTS)

    return {key: ctor for key, ctor in built if ctor is not None}
```

`_installed` and the `find_spec` import are gone; `CONSTRUCTS` is now the plain, unfiltered list of rivals.

One place I did not follow the recommendation: only `ImportError` is caught, not `SyntaxError`. "Not here" and "here and wrong" are different answers, and a rival that is installed and broken should stay loud rather than quietly vanish from a benchmark that is meant to be comparing against it. Happy to widen it if that reads better.

<details>
<summary><b>Run four ways</b></summary>

Everything installed — five rows:

```
salix (this project)          0.164       8.8      52.1
msgspec.Struct               15.447       8.9      59.5
typing.NamedTuple             1.641      71.6     137.6
dataclass (frozen+slots)      8.697     358.0     235.1
record-type (@record)         9.894      94.6     239.8
```

Under the `records` shim above — `record-type` drops instead of tracebacking:

```
salix (this project)          0.153       8.8      51.6
msgspec.Struct               15.576       9.0      58.4
typing.NamedTuple             1.683      71.8     137.3
dataclass (frozen+slots)      8.994     346.3     235.4
```

On a checkout where `salix/` holds only `py.typed` and the stub — the namespace-package case round 2 fixed, which this keeps:

```
msgspec.Struct               15.552       9.0      60.8
typing.NamedTuple             1.650      71.3     136.9
dataclass (frozen+slots)      8.686     351.9     236.7
record-type (@record)         9.812      94.7     239.0
```

And both at once, leaving three rows.

</details>

**2 — the lazy-import comment was false for one of five.** Correct: `typing` is stdlib and `Construct` and `Row` already import it at module scope. The comment now says that instead of asserting an invariant that holds four times out of five.

**3 — `label.split()[0]` renamed `typing.NamedTuple` to `typing`.** Taken, via the field rather than the parse: `Construct` and `Row` carry a `short` alongside `label`, so the summary column is declared rather than recovered from prose written for a human. The headline reads `salix / NamedTuple / msgspec` again.

> [!WARNING]
> **LLM Disclosure**
>

Issue #1, opened 2026-07-29
Release: bump the version off 0.0.0 and configure PyPI Trusted Publishing


JPHutchins:
The machinery is in place as of dfcb9f1; what remains are the two inputs only the author can supply.

## What exists

- `nix build .#release` produces the 50 wheels plus an sdist.
- `tools/check_tag.py` refuses a tag that does not name the version in `pyproject.toml`.
- `.github/workflows/release.yaml` fires on a `v*` tag only, re-runs `nix flake check --all-systems` against the tagged source, and publishes through PyPI Trusted Publishing (no API token).

## What is outstanding

1. **Bump `version` off `"0.0.0"`** in `pyproject.toml`. It is currently a placeholder, and it is load-bearing in one non-obvious way: uv keys its cache on name+version, so a rebuilt `0.0.0` wheel is indistinguishable from one built months ago. That is the only reason the `wheel_test` task carries `--no-cache`, and a real version retires the flag.
2. **Configure the PyPI Trusted Publisher** for this repository and workflow, and create the `pypi` GitHub environment the workflow names.

Until (2) is done the release workflow will fail at the publish step, which is the correct failure — everything before it still runs and verifies.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to build out the remaining substantial work in this repo and then open issues for whatever was left outstanding. This issue records the two release inputs I could not supply myself.

PR #41, opened 2026-07-31
fix: make the release path something that could actually run

### LINKED ISSUE #16: release.yaml runs 'nix flake check --all-systems', which d713085 removed as unrunnable on a runner
`.github/workflows/release.yaml` still runs the command `d713085` removed for being unrunnable on a GitHub runner. The release path has never executed — there are no tags and no runs of this workflow — so the first `git push --tags` will fail at step 4.

[.github/workflows/release.yaml:32](https://github.com/JPHutchins/salix/blob/74d2647/.github/workflows/release.yaml#L32):

```yaml
      - run: nix flake check --all-systems
```

`d713085` ("fix: stop asking CI to build a darwin check on an x86_64 runner") replaced exactly this in `tasks.py`, and its message says what happens:

> `294c463` added `--all-systems` on the belief that it evaluates the other systems. It does not: it makes nix **build** them. An x86_64 runner cannot build the aarch64-darwin check, so CI has been red since that commit went out — `"error: Cannot build ... Required system: 'aarch64-darwin'"`.

`release.yaml` was written at `dfcb9f1`, which predates the fix, and was never revisited. It is the only file in the repo that still carries the flag:

```console
$ grep -rn "flake check" .github/workflows/ tasks.py
.github/workflows/release.yaml:32:      - run: nix flake check --all-systems
tasks.py:121:flake_check = Sequential(flake_evaluates, Task("nix flake check", when=NIX_INPUTS))
```

<details>
<summary><b>Confirmation that this has never run</b></summary>

```console
$ git tag | wc -l
0
$ gh run list --workflow=release.yaml --limit 5
(no output)
```

The `systems` list in `flake.nix` includes `aarch64-darwin`, and `checks` is populated for every system in it — 50 wheel derivations plus `nixfmt`, `c-tests` and `wheels-verified` each. Those aarch64-darwin derivations have to be *built*, and `runs-on: ubuntu-latest` cannot build them.

</details>

<details>
<summary><b>The comment above the step is also wrong</b></summary>

```yaml
      # The same check CI runs, on the same source: every wheel built and
      # verified, the payloads matched to their tags, twine --strict over all of
      # them and the sdist, and the native wheel installed and tested.
      - run: nix flake check --all-systems
```

CI does not run this. CI runs `camas ci`, which is:

```python
ci = Parallel(flake_check, free_threaded, format_check, lint, analyze, type_check)
```

where `flake_check` is the two-step form. So the release differs from CI in two ways, not one: the flag, and the five other tasks CI runs that the release does not.

Whether the release *should* re-run lint, analyze and type_check is a separate call — a tag usually points at a commit CI already proved. But nothing currently enforces that either: `on: push: tags: ["v*"]` fires for a tag on any commit, including one that never went through CI. Worth deciding alongside the fix.

</details>

<details>

Replace the step with what `tasks.py` settled on, ideally by calling camas so there is one definition rather than two:

```yaml
      - run: nix develop --command uv sync --no-install-workspace
      - run: nix develop --command uv run --no-sync camas flake_check
```

`tasks.py` is described in its own first line as *"the single source of truth for local and CI"*, and this file is the one place that restates a task instead of invoking it. That is the actual defect; the stale flag is the symptom.

Given that the whole path is unproven, it is worth exercising it end to end before trusting it — a `v0.0.0-rc1`-style tag against TestPyPI, or `act`, or a temporary `workflow_dispatch` trigger. `salix` is currently unregistered on both PyPI and TestPyPI (both return 404), so a TestPyPI dry run is available and free.

</details>

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for a whole-repository code review in preparation for publishing salix to PyPI, with the findings filed as issues for future agents to resolve. This one blocks issue #1.

### LINKED ISSUE #22: release.yaml pulls an action from a mutable branch while holding id-token: write
`release.yaml` runs with `id-token: write` — the credential that mints a PyPI Trusted Publishing token — and pulls an action from a **mutable branch**. Anyone who can move that branch gets code execution in the job that can publish to PyPI as salix.

[.github/workflows/release.yaml:20](https://github.com/JPHutchins/salix/blob/74d2647/.github/workflows/release.yaml#L20):

```yaml
      - uses: DeterminateSystems/nix-installer-action@main
```

`@main` is not a version. It is whatever that repository's default branch points at when the job starts.

| action | ref | mutable? |
| --- | --- | --- |
| `DeterminateSystems/nix-installer-action` | `@main` | **yes — a branch** |
| `pypa/gh-action-pypi-publish` | `@release/v1` | **yes — a branch** |
| `nix-community/cache-nix-action` | `@v7` | yes — a tag can be moved |
| `actions/checkout` | `@v7` | yes — a tag can be moved |

Same `@main` appears in `ci.yaml:22`, where the blast radius is smaller but the exposure is the same.

<details>
<summary><b>Recommended: pin to full commit SHAs</b></summary>

```yaml
      - uses: DeterminateSystems/nix-installer-action@<40-hex-sha>  # vX.Y.Z
```

This is [GitHub's own hardening guidance](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) for third-party actions, and PyPA's own recommendation for `gh-action-pypi-publish` when maximum security matters — their `release/v1` branch exists for convenience, not for a workflow holding `id-token: write`.

Dependabot keeps SHA pins current if `.github/dependabot.yml` includes the `github-actions` ecosystem, which is also currently absent.

</details>

<details>
<summary><b>Other hardening in the same pass</b></summary>

**No `permissions:` block on `ci.yaml`.** The job inherits whatever the repository default is, which for many repositories is read/write on `contents`. `ci.yaml` needs only `contents: read`; setting it explicitly at the workflow level is one line and removes the dependency on a setting that lives outside the repo.

**The tag is interpolated into a `run:` shell.** [.github/workflows/release.yaml:27](https://github.com/JPHutchins/salix/blob/74d2647/.github/workflows/release.yaml#L27):

```yaml
      - run: nix develop --command python tools/check_tag.py "${{ github.ref_name }}"
```

Only someone who can push a tag to this repository controls that string, so it is not an outside vector — but it is in the job holding `id-token: write`, and the fix is free: pass it through `env:` and reference `"$TAG"` in the script, so the value never reaches the shell as text.

**No `concurrency:` group on `release.yaml`.** `ci.yaml` has one. Two tags pushed together would race two publish jobs.

**No `timeout-minutes:` on any job.** The default is 6 hours. The wheels job cross-compiles 50 targets, so it is not fast, but 6 hours is not the number.

**Nothing ties the tag to a tested commit.** `on: push: tags: ["v*"]` fires for a tag on any commit, including one CI never saw. `tools/check_tag.py` checks the tag against the declared version — it does not check that the commit is green. Enforcing "this SHA has a successful CI run" is a real option; so is deciding that the `pypi` environment's required reviewers are the gate. Worth writing down which.

**Attestations.** `pypa/gh-action-pypi-publish` produces PEP 740 attestations by default on recent releases, which is the right default for a first publish — worth confirming rather than assuming once the action is SHA-pinned to a known version.

</details>

<details>
<summary><b>Assessed and not a problem</b></summary>

`cache-nix-action` restoring a cache into the release job looked worth checking. GitHub scopes caches by branch, and a cache written on a PR branch is not readable from another branch, so a fork PR cannot seed the release job's cache. Beyond that, the cache holds nix store paths, which are content-addressed — a substituted path that does not hash correctly is rejected by nix itself. Noted so the question is not reopened.

</details>

> [!WARNING]
> **LLM Disclosure**
>

### LINKED ISSUE #35: check_tag.py compares raw strings, so a tag can pass while PyPI publishes a different version
`check_tag.py` compares raw strings. PEP 440 normalises versions before they reach a filename, so a tag can pass the gate while PyPI publishes a different string — which is exactly the disagreement the gate exists to catch.

Its own docstring states the job:

> The tag triggers the release and pyproject.toml decides what gets uploaded under it. Nothing reconciles the two, and a disagreement is a typo in one of them -- caught here, before an upload that cannot be taken back.

Measured, with the current comparison:

```
declared      tag           check_tag  PyPI version
1.0.0-rc1     v1.0.0-rc1    PASS       1.0.0rc1   <-- gate passes, PyPI publishes a different string
1.0.0rc1      v1.0.0rc1     PASS       1.0.0rc1
1.0           v1.0          PASS       1.0
01.0.0        v01.0.0       PASS       1.0.0   <-- gate passes, PyPI publishes a different string
1.0.0.RELEASE v1.0.0.RELEASE PASS      INVALID   <-- gate passes, PyPI publishes a different string
```

Tag `v1.0.0-rc1` releases a distribution called `1.0.0rc1`. Someone reading the tag list and someone reading PyPI see different versions, and `pip install salix==1.0.0-rc1` — the string in the tag — resolves through normalisation but the tag no longer names the artifact literally.

The last row is worse: `1.0.0.RELEASE` is not a valid PEP 440 version at all. The gate passes it; the build fails later, or produces something unexpected.

<details>
<summary><b>The comparison</b></summary>

[tools/check_tag.py:34-40](https://github.com/JPHutchins/salix/blob/74d2647/tools/check_tag.py#L34-L40):

```python
    declared = declared_version(Path("pyproject.toml"))
    named = tag.removeprefix("v")

    if named != declared:
        raise SystemExit(f"tag {tag} names {named}, but pyproject.toml declares {declared}")
```

String equality, and nothing validates that either side is a version.

</details>

<details>

Normalise both sides and compare the results, which also rejects an unparseable version for free:

```python
from packaging.version import InvalidVersion, Version

    try:
        named, declared = Version(tag.removeprefix("v")), Version(declared_text)
    except InvalidVersion as invalid:
        raise SystemExit(f"{invalid}") from None

    if named != declared:
        ...
```

`packaging` is a dependency to acquire — the script currently imports only `tomllib` and runs under `nix develop --command python`. Two ways round that: add `packaging` to the release environment, or implement the small subset by hand (strip a leading `v`, lowercase, collapse the pre-release separators). The first is honest; the second avoids a dependency in a gate whose whole value is that it cannot itself fail.

Stricter and simpler still: **require the tag to be exactly `v` + the declared version, and require the declared version to already be in normalised form.** Then a maintainer cannot write `1.0.0-rc1` in `pyproject.toml` in the first place, and the string comparison becomes correct rather than merely usually-correct.

Two smaller things in the same file, worth handling in the same pass:

- `tag.removeprefix("v")` is a no-op for a tag with no `v`. The workflow only fires on `v*`, so this cannot arrive from the trigger, but the script is also documented as runnable by hand.
- The workflow interpolates the tag into a shell: `run: nix develop --command python tools/check_tag.py "${{ github.ref_name }}"`. Only someone who can push a tag controls that string, so it is not an outside vector, but it is in the job holding `id-token: write` — see #22. Passing it through `env:` instead of interpolation removes the question.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Closes #16. Closes #22. Closes #35.

The release workflow has never executed — no tags, no runs — and as written it could not have finished. Three things, all in the job that holds `id-token: write`.

### It ran a command that cannot succeed (#16)

Step 4 was `nix flake check --all-systems`, the exact command `d713085` removed from `tasks.py` for being unrunnable on a runner. `--all-systems` does not evaluate the other systems; it builds them, and an x86_64 box cannot build the `aarch64-darwin` check. `release.yaml` predates that fix and was the only file still carrying the flag.

The flag is the symptom. The defect is that this file **restated** a task instead of invoking the one definition, which is how it went stale. It now runs `camas ci`.

<details>
<summary><b>Why <code>camas ci</code> and not <code>camas flake_check</code> — the one judgement call here</b></summary>

The comment above the old step claimed "the same check CI runs". That was wrong twice: the flag, and the five other tasks in `ci` that the release skipped.

`on: push: tags` fires for a tag on **any** commit, including one CI never saw, so the release cannot assume the checks passed somewhere else. And `ci` is `Parallel(flake_check, free_threaded, format_check, lint, analyze, type_check)` where `nix flake check` dominates at ~240s — so making the claim true costs almost nothing in wall-clock.

If you would rather the release trust CI and only build the artifacts, it is a one-word change back to `camas flake_check`.

</details>

### It pulled actions from mutable branches (#22)

`id-token: write` mints the PyPI Trusted Publishing token. The job pulled `nix-installer-action@main` — a branch, so whatever that repo's default points at when the job starts — and `gh-action-pypi-publish@release/v1`, also a branch. `@v7` tags are no better; a tag moves.

Every action is now pinned to a commit SHA resolved through the GitHub API, annotated tag objects dereferenced to their commits, with the version in a trailing comment. `ci.yaml` took the same treatment. `.github/dependabot.yml` is what keeps them current — it is in #40.

Also in this pass: explicit `permissions: contents: read` on `ci.yaml`, which had none; a `concurrency` group on `release.yaml` keyed by workflow rather than ref, because two tags are two refs and would otherwise publish at once; `timeout-minutes`, because the default is six hours; and the tag passed through `env:` so it reaches the shell as a value rather than as interpolated text.

### The version gate compared raw strings (#35)

Measured, old script against new, same synthetic `pyproject.toml`:

| declared | tag | old | new | what PyPI publishes |
| `1.0.0` | `v1.0.0` | PASS | PASS | `1.0.0` |
| `1.0.0rc1` | `v1.0.0rc1` | PASS | PASS | `1.0.0rc1` |
| `1.0` | `v1.0` | PASS | PASS | `1.0` |
| `1.0.0-rc1` | `v1.0.0-rc1` | PASS | **refuse** | `1.0.0rc1` — tag names something else |
| `01.0.0` | `v01.0.0` | PASS | **refuse** | `1.0.0` — tag names something else |
| `1.0.0.RELEASE` | `v1.0.0.RELEASE` | PASS | **refuse** | not a version at all |
| `1.0.0+local` | `v1.0.0+local` | PASS | **refuse** | rejected on upload |

The gate exists to catch that disagreement and passed all four of them.

<details>
<summary><b>Canonical form, and why there is no <code>packaging</code> dependency</b></summary>

The rule is now: the declared version must already be in PEP 440 **canonical** form, and the tag must be exactly `v` + that string. That makes the comparison correct rather than usually correct, and rejects an unparseable version on the way — the strictest of the three shapes #35 suggested, and the only one where the tag and the published distribution can never name different things.

Adding `packaging` to a gate whose whole value is that it cannot itself fail is a poor trade, so the canonical form is a regex. It was not written from memory — it was checked against `str(Version(v)) == v` over **4620 generated strings**, every combination of epoch, release, pre, post and dev segment this produces:

```
checked 4620 non-local strings, 0 disagreements: []
```

Local versions (`1.0+ubuntu.1`) are the only place the regex and `packaging` differ, and they are excluded deliberately: PyPI refuses to accept one.

`tests/test_check_tag.py` runs the gate the way the release runs it — subprocess, synthetic `pyproject.toml` — over both tables. It skips below 3.11, because the script reads `pyproject.toml` with `tomllib` and the release invokes it through `uv run`, which takes 3.14, the first interpreter `.python-version` names. That skip is itself a finding: the previous arrangement would have silently depended on whichever `python` the runner's `PATH` happened to expose inside `nix develop`.

</details>

Verified that `.#release` still evaluates (`salix-release`) and that all four workflow files parse. `camas check` green at 25/25.

This unblocks #1, which is the remaining half of publishing and only @JPHutchins can do: set the version, and register the Trusted Publisher on PyPI.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for the pre-publish review's issues to be resolved in batches, simplest first; this is batch 4 of 12, and #26 indexes the rest.

### DIFF
--- a/.github/actions/code-review-setup/action.yml
+++ b/.github/actions/code-review-setup/action.yml
@@ -14,8 +14,8 @@ runs:
     # as the runner and cannot take a write lock in that store -- `nix develop` dies with
     # `opening lock file ...-source.lock: Permission denied` on a cache hit.
     #
-    # Pinned by digest, not the @main CI takes: this step runs in the job that holds the model key,
-    # so what it installs has to be a thing that cannot change under us.
+    # Pinned by digest, and CI now pins the same one: this step runs in the job that holds the model
+    # key, so what it installs has to be a thing that cannot change under us.
     - uses: DeterminateSystems/nix-installer-action@ef8a148080ab6020fd15196c2084a2eea5ff2d25 # v22
     # Digest-pinned for the same reason, so the pin above is not theatre -- both of these run before
     # the agent, in the same job, with the same key in the environment.
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,6 +9,11 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+# Set explicitly: the repository default lives outside this file and can change
+# without a commit.
+permissions:
+  contents: read
+
 jobs:
   # Every wheel for every interpreter, cross-compiled on one runner, which also
   # emits the fan-out for the legs below. The legs are the only other jobs, so
@@ -18,12 +23,12 @@ jobs:
     outputs:
       coverage: ${{ steps.emit.outputs.matrix }}
     steps:
-      - uses: actions/checkout@v7
-      - uses: DeterminateSystems/nix-installer-action@main
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      - uses: DeterminateSystems/nix-installer-action@ef8a148080ab6020fd15196c2084a2eea5ff2d25 # v22
       # Keyed on what determines the closure. A source-only change misses the
       # key but still restores through the prefix, so the interpreter archives
       # -- the expensive part -- survive and only the compiles re-run.
-      - uses: nix-community/cache-nix-action@v7
+      - uses: nix-community/cache-nix-action@7df957e333c1e5da7721f60227dbba6d06080569 # v7.0.2
         with:
           primary-key: nix-${{ runner.os }}-${{ hashFiles('flake.lock', 'flake.nix', 'nix/**') }}
           restore-prefixes-first-match: nix-${{ runner.os }}-
@@ -35,7 +40,7 @@ jobs:
       - run: nix develop --command uv sync --no-install-workspace
       - run: nix develop --command uv run --no-sync camas ci
       - run: nix build .#default --out-link result-wheels
-      - uses: actions/upload-artifact@v7
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
           name: wheels
           path: result-wheels/*.whl
@@ -54,9 +59,9 @@ jobs:
       matrix: ${{ fromJSON(needs.wheels.outputs.coverage) }}
     runs-on: ${{ matrix.OS }}
     steps:
-      - uses: actions/checkout@v7
-      - uses: astral-sh/setup-uv@v9.0.0
-      - uses: actions/download-artifact@v8
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      - uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
         with:
           name: wheels
           path: result-wheels
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,6 +20,11 @@ jobs:
   # this is what keeps the runner count down.
   wheels:
     runs-on: ubuntu-latest
+    # cache-nix-action purges here, and deleting a cache is a REST call the
+    # workflow-level `contents: read` would leave without a scope.
+    permissions:
+      contents: read
+      actions: write
     outputs:
       coverage: ${{ steps.emit.outputs.matrix }}
     steps:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,29 +7,47 @@ on:
   push:
     tags: ["v*"]
 
+# Two tags are two refs, so grouping by ref would let both publish at once.
+# Never cancelled: a half-finished upload is the state worth avoiding.
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
 jobs:
   release:
     runs-on: ubuntu-latest
+    # 50 cross builds is not fast; six hours is still not the number.
+    timeout-minutes: 120
     environment: pypi
     permissions:
+      contents: read
       # Trusted Publishing: PyPI takes a short-lived OIDC token from this
-      # workflow, so there is no API token to hold, rotate or leak.
+      # workflow, so there is no API token to hold, rotate or leak. It is also
+      # why every action below is pinned to a commit: a tag can be moved and a
+      # branch always is, and this is the job that can publish as salix.
       id-token: write
     steps:
-      - uses: actions/checkout@v7
-      - uses: DeterminateSystems/nix-installer-action@main
-      - uses: nix-community/cache-nix-action@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      - uses: DeterminateSystems/nix-installer-action@ef8a148080ab6020fd15196c2084a2eea5ff2d25 # v22
+      - uses: nix-community/cache-nix-action@7df957e333c1e5da7721f60227dbba6d06080569 # v7.0.2
         with:
           primary-key: nix-${{ runner.os }}-${{ hashFiles('flake.lock', 'flake.nix', 'nix/**') }}
           restore-prefixes-first-match: nix-${{ runner.os }}-
           gc-max-store-size-linux: 5G
 
-      - run: nix develop --command python tools/check_tag.py "${{ github.ref_name }}"
+      - run: nix develop --command uv sync --no-install-workspace
 
-      # The same check CI runs, on the same source: every wheel built and
-      # verified, the payloads matched to their tags, twine --strict over all of
-      # them and the sdist, and the native wheel installed and tested.
-      - run: nix flake check --all-systems
+      # Through env: so the tag reaches the shell as a value, never as text.
+      - env:
+          TAG: ${{ github.ref_name }}
+        run: nix develop --command uv run --no-sync python tools/check_tag.py "$TAG"
+
+      # Invoked, not restated -- restating it is how the old step went stale.
+      # A tag can point at a commit CI never saw, so this cannot be skipped.
+      - run: nix develop --command uv run --no-sync camas ci
 
       - run: nix build .#release --out-link result-release
 
@@ -40,4 +58,4 @@ jobs:
           cp -L result-release/*.whl result-release/*.tar.gz dist/
           ls -l dist
 
-      - uses: pypa/gh-action-pypi-publish@release/v1
+      - uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # v1.14.2
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,8 +7,10 @@ on:
   push:
     tags: ["v*"]
 
-# Two tags are two refs, so grouping by ref would let both publish at once.
-# Never cancelled: a half-finished upload is the state worth avoiding.
+# Two tags are two refs, so grouping by ref would let both publish at once. A
+# running release is never cancelled -- a half-finished upload is the state
+# worth avoiding -- but GitHub holds only one run pending per group, so a third
+# tag pushed while one waits replaces it.
 concurrency:
   group: ${{ github.workflow }}
   cancel-in-progress: false
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -15,9 +15,6 @@ concurrency:
   group: ${{ github.workflow }}
   cancel-in-progress: false
 
-permissions:
-  contents: read
-
 jobs:
   release:
     runs-on: ubuntu-latest
@@ -39,6 +36,9 @@ jobs:
           primary-key: nix-${{ runner.os }}-${{ hashFiles('flake.lock', 'flake.nix', 'nix/**') }}
           restore-prefixes-first-match: nix-${{ runner.os }}-
           gc-max-store-size-linux: 5G
+          # Restore-only. Saving is a REST call needing actions: write, which
+          # this job deliberately does not have; CI owns the cache.
+          save: false
 
       - run: nix develop --command uv sync --no-install-workspace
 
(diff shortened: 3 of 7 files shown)


JPHutchins:
The first push turned CI red and this fixes it. `wheel-smoke` runs `pytest ${./tests}` against a copy of `tests/` in the nix store — with no `tools/` beside it — so `tests/test_check_tag.py` was calling a script at a path that does not exist, and 26 cases failed.

```
❌ checks.x86_64-linux.wheel-smoke
   > FAILED test_check_tag.py::...[1.0] - assert 2 == 0
   > 26 failed, 579 passed, 5 skipped in 9.80s
```

The test now skips when `tools/` is not beside it. `wheel-smoke` asks whether the built wheel works, not whether this repository's release tooling does.

Verified: 27 passed on 3.14, 27 skipped on 3.10, `nix build .#checks.x86_64-linux.wheel-smoke` green, `camas check` 25/25.

Also folded in a comment trim across this branch. The comments I added were re-arguing the diff rather than recording a *why* the code cannot show — the reasoning belongs in the commit message, which already carries it.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked that fixes be folded into the same round rather than pushed separately, so this force-push carries both the CI fix and a comment cleanup she called for.

JPHutchins:
Four of five taken. `camas check` green at 25/25.

**1 — cache permissions (major).** Taken, but on the argument rather than on your evidence, because the evidence was inconclusive both ways. On [run 30657816200](https://github.com/JPHutchins/salix/actions/runs/30657816200) the purge ran its *list* call, found nothing scoped to `refs/pull/41/merge`, and never reached a delete; the save skipped because the key already existed:

```
Purging cache(s) created before ... scoped to "refs/pull/41/merge" ...
No cache(s) found.
Not purging.
```

So neither privileged path executed. On `main` both would. One detail cuts against the benign reading: the list call *succeeded* under `contents: read`, which means either the `actions` scope was not really restricted, or the call failed silently and reported "no caches" — and the second is worse than the finding. `actions: write` on the `wheels` job, the only one that purges. `release.yaml` does not purge and stays at `contents: read`.

**2 — concurrency comment.** Taken. "Never cancelled" was false for a pending run.

**4 — error message.** Taken, and it caught a second bug for free: the test asserted `"v1.0.0" in stderr`, which only ever passed because the old message happened to contain that substring. The message now names both versions and the test asserts both.

**5 — missing cases.** Taken, all five verified against `packaging` before being written down:

| declared | `packaging` | regex |
| --- | --- | --- |
| `1.0.0rc01` | `1.0.0rc1` | refused |
| `1.0.0.post01` | `1.0.0.post1` | refused |
| `1.0.0.dev01` | `1.0.0.dev1` | refused |
| `1.0.0.dev2.post1` | `InvalidVersion` | refused |
| `1.0.0.post1.rc1` | `InvalidVersion` | refused |

### 3 — declined, and replaced with something stronger

The finding's own argument is that `packaging` is present **transitively**, via setuptools, pytest and camas. That is the fragility the regex exists to avoid, not an answer to it: if camas drops it, the gate fails at the one moment it must not, and the gate's entire value is that it cannot itself fail. Depending on it deliberately means *declaring* it — which is the new dependency the finding says is unnecessary.

You are right about the weaker half, though. A hand-rolled PEP 440 pattern should not rest on a measurement I ran once in a scratch buffer. So that measurement is now a test:

```python
def test_the_pattern_accepts_exactly_what_packaging_calls_canonical():
    ...
    assert [v for v in generated
            if bool(check_tag.CANONICAL_VERSION.fullmatch(v))
            != packaging_calls_it_canonical(v)] == []
```

3,500 generated versions — every combination of epoch, release, pre, post and dev segment, plus malformed heads — checked on every run. `packaging` is now a declared **test** dependency; the gate still imports only the standard library. If it ever diverges from `packaging`, a test fails instead of a release shipping under the wrong name.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, answering the automated review on this PR as part of a batch of pre-publish fixes she asked to be iterated with that reviewer.

Issue #37, opened 2026-07-31
84% of every wheel is DWARF, and 14 of the 50 are never loaded before shipping


JPHutchins:
**84% of every published wheel is DWARF debug information**, and 14 of the 50 wheels are never loaded by anything before they ship.

Measured on `salix-0.0.0-cp314-cp314-manylinux_2_17_x86_64.whl`:

```console
$ ls -l payload.so
144712
$ size --format=sysv payload.so | grep debug
.debug_loc        20688
.debug_abbrev      7207
.debug_info       62375
.debug_ranges      3712
.debug_str         6406
.debug_line       16355
$ strip payload.so && ls -l payload.so
23160
```

144,712 bytes shipped, 23,160 bytes of actual code and data. For a project whose pitch is a tiny `.so` and a sub-millisecond import, six times the necessary bytes in every artifact is worth one line in `nix/wheel.nix`.

<details>
<summary><b>Second half: what the matrix never executes</b></summary>

The 50 wheels get two kinds of check before publication:

- **`tools/check_wheel.py`** verifies every one of them statically — RECORD hashes and sizes, the WHEEL tag against the filename, and that the payload's container and machine match the platform tag. That is real coverage and it catches the failures that matter most.
- **Actually importing one** happens for a much smaller set. `wheel-smoke` in `flake.nix` runs only on `x86_64-linux` with 3.14, and the CI `coverage` matrix samples ten legs: ubuntu across all six interpreters, plus macOS and Windows at the oldest and newest.

What that leaves never loaded: **macOS x86_64** and **Windows arm64**, both cross-compiled by zig and neither runnable on any runner in the matrix. GitHub now offers `macos-13` for Intel and `windows-11-arm` for ARM64, so at least the first is straightforwardly addressable.

Both are the same category of gap — an artifact that goes to PyPI having been inspected but never run — which is why they are in one issue.

</details>

<details>
<summary><b>Shape of a fix</b></summary>

**Stripping** is a build-time flag in the cross compile: `-Wl,-s`, or `--strip-all` as a separate step, in [nix/wheel.nix](https://github.com/JPHutchins/salix/blob/74d2647/nix/wheel.nix). Worth confirming per platform: `strip` on a Mach-O and on a PE has different flags and different failure modes, and zig's linker may want `-Wl,-S` rather than `-s` on some targets. `tools/check_wheel.py` will keep verifying the payload afterwards, since it reads magic numbers and machine fields that stripping does not touch.

[...]

PR #42, opened 2026-07-31
build: what ships, and what ships in it

### LINKED ISSUE #23: Per-wheel flake outputs cannot be built by name: nix splits the attribute path on the version dot
The per-wheel flake outputs cannot be built by name. Nix splits an installable's attribute path on `.`, and every wheel name embeds the Python version's dot:

```console
$ nix build .#wheel-3.14-manylinux-x86_64
error: flake 'git+file:///home/jp/repos/jp-struct' does not provide attribute
'packages.x86_64-linux.wheel-3.14-manylinux-x86_64',
'legacyPackages.x86_64-linux.wheel-3.14-manylinux-x86_64' or 'wheel-3.14-manylinux-x86_64'
```

The attribute is there — `nix flake show` lists all 50 — it is `wheel-3` `.` `14-manylinux-x86_64` to the parser.

The working form is the fully qualified, quoted one:

```console
$ nix build '.#packages.x86_64-linux."wheel-3.14-manylinux-x86_64"'
/nix/store/933zwqpm8h1091c34gg844cavfg8m0fh-salix-wheel-3.14-manylinux-x86_64-0.0.0
```

Low impact — `nix build .#default` builds all 50 and is the path CI takes — but it makes the per-target outputs effectively undiscoverable for anyone debugging one cross build, which is the only reason they are exposed individually.

<details>
<summary><b>Where the name is built</b></summary>

[flake.nix:105](https://github.com/JPHutchins/salix/blob/74d2647/flake.nix#L105):

```nix
          named = lib.mapAttrs' (name: wheel: lib.nameValuePair "wheel-${name}" wheel) wheels;
```

and `name` comes from [flake.nix:86](https://github.com/JPHutchins/salix/blob/74d2647/flake.nix#L86):

```nix
              lib.nameValuePair "${pythonMinor}-${platformName}" (
```

where `pythonMinor` is `"3.14"` straight from `nix/python-targets.nix`.

Substituting the dot at the point the output name is formed — `wheel-py314-manylinux-x86_64`, or `3_14` — makes `nix build .#wheel-py314-manylinux-x86_64` work unquoted. `pythonMinor` itself must keep its dot: it is compared against `.python-version` by the assertion at [flake.nix:135](https://github.com/JPHutchins/salix/blob/74d2647/flake.nix#L135), and it feeds the wheel tags.

The free-threaded names (`3.14t`) have the same problem and the same fix.

Alternatively, leave the names and note the quoting in a comment — but a name that cannot be typed is a name that will be mistyped.

</details>

> [!WARNING]
> **LLM Disclosure**
>

### LINKED ISSUE #32: -Werror ships in the sdist's compile flags, where a user cannot override it

### LINKED ISSUE #33: cp315 wheels are built against CPython 3.15.0b4, a beta
The `cp315` and `cp315t` wheels are built against **CPython 3.15.0b4** — a beta. Publishing them under a `cp315` tag tells pip they are compatible with 3.15 final, and nothing has established that.

`nix/python-targets.nix`:

```nix
    "3.15" = {
      version = "3.15.0b4";
...
    "3.15t" = {
      version = "3.15.0b4";
```

That is 10 of the 50 wheels (3.15 and 3.15t across five platforms).

<details>
<summary><b>Why the tag is a promise</b></summary>

`cp315-cp315-manylinux_2_17_x86_64` means "compatible with the CPython 3.15 ABI". A beta's ABI is provisional: CPython freezes it at the first release candidate, and structure layouts, macro definitions and slot semantics can and do change up to that point. salix reads `PyHeapTypeObject`'s layout directly — [`struct_heap_type_members`](https://github.com/JPHutchins/salix/blob/74d2647/src/types.h#L88) computes a member array address from `tp_basicsize` — so it is precisely the kind of extension a layout change breaks.

If b4's ABI differs from 3.15.0 final, a user on 3.15.0 installs a wheel pip believes is compatible and gets a crash or a silent misread on import. There is no version of that failure that is easy to diagnose.

The exposure is a matter of timing rather than certainty: if 3.15.0 final ships before salix does, this resolves itself by regenerating the pins. If it does not, the choice has to be made deliberately.

</details>

<details>

1. **Do not publish `cp315` until 3.15.0 is final.** Build and test against the beta in CI — that is genuine early-warning value — and hold the wheels back from the upload. `nix/python-targets.nix` is generated by `tools/update_python_targets.py`, so the mechanism to include or exclude a minor already exists; the change is in what `flake.nix`'s `release` output collects, not in the matrix.
2. **Publish and re-publish.** Ship the beta-built wheels, then push a new release built against final. Works, but the first upload cannot be withdrawn — a yanked file stays yanked, and anyone who pinned in between has a bad artifact.
3. **Publish deliberately, and say so.** A `Development Status` classifier and a README line. Weakest, since pip will not read either before installing.

Option 1 is the conservative one and costs nothing but a later release.

Worth checking in the same pass: **`.python-version` lists 3.15 and the flake asserts the two agree**, so removing 3.15 from the release without removing it from the test matrix needs the release output filtered rather than the matrix trimmed. The assertion at [flake.nix:135](https://github.com/JPHutchins/salix/blob/74d2647/flake.nix#L135) is what enforces that, and it should stay.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Closes #23. Closes #32. Closes #33. Half of #37, which stays open.

Four things about the built artifacts. None of them touches the extension.

### The wheels got 77% smaller (#37, first half)

`-g0` on the cross builds, so the payload never carries DWARF rather than being stripped after the fact:

| | before | after |
| --- | --- | --- |
| payload `.so` | 144,712 B | **27,512 B** |
| wheel | 57,824 B | **13,443 B** |
| `.debug_*` sections | 6 | **0** |

27,512 and not the 23,160 that `strip` gives, because `-g0` drops the debug information and keeps the symbol table — which is what a backtrace needs.

### `-Werror` no longer ships to strangers (#32)

`setup.py` passes these as `extra_compile_args`, which land *after* `CFLAGS`, so `CFLAGS="-Wno-error"` loses. Anyone compiling from the sdist got warnings-as-errors on a compiler nobody here has tested — and one new warning in a future GCC breaks `pip install salix` for precisely the people who have no wheel: musl, ppc64le, s390x, riscv64, distribution packagers, `--no-binary`.

Three flag sets now, instead of one:

```
c-flags                     -std=c23 -O2 -Wdouble-promotion -Wall -Wextra -Wno-unused-parameter
c-flags --strict            ... -Werror
c-flags --strict --shipped  ... -Werror -g0
```

`STRICT` is opt-in through `SALIX_STRICT`, which camas sets and a stranger's `pip` does not. Verified both directions: `setup.py` emits no `-Werror` without the variable and does with it.

### Per-wheel outputs were unbuildable by name (#23)

Nix splits an attribute path on `.`, so `nix build .#wheel-3.14-manylinux-x86_64` looked for `wheel-3`. The 50 individually-exposed outputs exist to debug one cross build, and none of them could be named. `nix build .#wheel-314-manylinux-x86_64` now works — the dot is dropped from the attribute only, and `314` is the spelling the wheel tag already uses.

### cp315 wheels are held back until rc1 (#33)

Per @JPHutchins's ruling. CPython freezes the ABI at the first release candidate, and salix computes a member array address from `tp_basicsize`, so a `cp315` tag on a beta-built wheel is a promise it cannot keep. PyPI does not take an upload back.

<details>
<summary><b>What the predicate is, and what it holds back</b></summary>

The test is the *pinned version*, not a hard-coded `3.15`, so regenerating the pins is the whole of releasing them:

```nix
abiIsFrozen = version: !(lib.hasInfix "a" version || lib.hasInfix "b" version);
```

The release output goes from 51 paths to 40 — 39 wheels and the sdist. 11 held back, not the 10 #33 estimated: six for `3.15` and five for `3.15t`, which has no `windows-aarch64` target.

The build matrix is untouched and the `.python-version` assertion still holds, which is what #33 asked for: the early warning from building and testing against the beta is the reason it is in the matrix at all.

</details>

<details>
<summary><b>Also: <code>-std=c2x</code> → <code>-std=c23</code>, the deferred item from #25</b></summary>

#25 said this needed checking rather than assuming, because of the zig-bundled clang. Checked against every compiler in the matrix:

| compiler | version | `-std=c23` |
| --- | --- | --- |
| gcc | 15.3.0 | accepted |
| clang-tidy | LLVM 21.1.8 | accepted |
| zig cc | 0.16.0 | accepted |

</details>

### What is left in #37

The second half — `macos-x86_64` and `windows-arm64` wheels that ship having been inspected by `check_wheel.py` but never actually imported. Closing it means two new CI legs on `macos-13` and `windows-11-arm`, which cannot be exercised locally. That follows in its own PR rather than risking a red run on top of work that is measured.

`camas check` green at 25/25.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for the pre-publish review's issues to be resolved in batches, simplest first, and ruled on the cp315 question directly; this is batch 6 of 12, and #26 indexes the rest.

### DIFF
--- a/build_config.py
+++ b/build_config.py
@@ -28,15 +28,23 @@ BUILD: Final = BuildConfig(
     # -Wno-unused-parameter: CPython slot signatures are fixed by the API and
     # routinely ignore an argument.
     c_flags=(
-        "-std=c2x",
+        "-std=c23",
         "-O2",
-        "-Werror",
         "-Wdouble-promotion",
         "-Wall",
         "-Wextra",
         "-Wno-unused-parameter",
     ),
 )
+
+# setup.py passes these after CFLAGS, so an sdist build cannot override them.
+# One new warning in a future compiler would then turn `pip install salix` into
+# a failure, on the machines that have no wheel and must compile.
+STRICT: Final = ("-Werror",)
+
+# 84% of a published payload was DWARF that nothing on the far end reads. Not
+# in c_flags, so a local build stays debuggable.
+SHIPPED: Final = ("-g0",)
 
 
 if __name__ == "__main__":
@@ -47,5 +55,11 @@ if __name__ == "__main__":
             print("\n".join(BUILD.sources))
         case ["c-flags"]:
             print("\n".join(BUILD.c_flags))
+        case ["c-flags", "--strict"]:
+            print("\n".join(BUILD.c_flags + STRICT))
+        case ["c-flags", "--strict", "--shipped"]:
+            print("\n".join(BUILD.c_flags + STRICT + SHIPPED))
         case _:
-            raise SystemExit("usage: build_config.py {sources|c-flags}")
+            raise SystemExit(
+                "usage: build_config.py {sources|c-flags [--strict [--shipped]]}"
+            )
--- a/build_config.py
+++ b/build_config.py
@@ -27,8 +27,12 @@ BUILD: Final = BuildConfig(
     ),
     # -Wno-unused-parameter: CPython slot signatures are fixed by the API and
     # routinely ignore an argument.
+    # -std=c2x, not -std=c23: they select the same language mode -- both give
+    # __STDC_VERSION__ 202311 -- but c2x is accepted by GCC 9+ and Clang 9+
+    # while c23 needs GCC 14+ or Clang 18+. An sdist has to compile on whatever
+    # the machine has, and Ubuntu 24.04 LTS still ships GCC 13.
     c_flags=(
-        "-std=c23",
+        "-std=c2x",
         "-O2",
         "-Wdouble-promotion",
         "-Wall",
@@ -59,6 +63,8 @@ if __name__ == "__main__":
             print("\n".join(BUILD.c_flags + STRICT))
         case ["c-flags", "--strict", "--shipped"]:
             print("\n".join(BUILD.c_flags + STRICT + SHIPPED))
+        case ["c-flags", "--shipped", *_]:
+            raise SystemExit("build_config.py: --shipped only follows --strict")
         case _:
             raise SystemExit(
                 "usage: build_config.py {sources|c-flags [--strict [--shipped]]}"
--- a/flake.nix
+++ b/flake.nix
@@ -48,7 +48,10 @@
       # PyHeapTypeObject's layout directly, so a cp3XX wheel built against an
       # alpha or a beta is a promise it cannot keep -- and PyPI does not take an
       # upload back.
-      abiIsFrozen = version: !(lib.hasInfix "a" version || lib.hasInfix "b" version);
+      # Matched at the pre-release suffix rather than by looking for the letters
+      # anywhere, so a version string that happens to contain one is not read as
+      # an alpha. `rc` is frozen and deliberately absent.
+      abiIsFrozen = version: builtins.match "[0-9.]+(a|b)[0-9]+" version == null;
 
       releasableWheelNames = map ({ pythonMinor, platformName }: "${pythonMinor}-${platformName}") (
         lib.filter ({ pythonMinor, ... }: abiIsFrozen matrix.pythons.${pythonMinor}.version) wheelIds
--- a/flake.nix
+++ b/flake.nix
@@ -44,6 +44,16 @@
         )
       ) pinnedPythons;
 
+      # CPython freezes the ABI at the first release candidate, and salix reads
+      # PyHeapTypeObject's layout directly, so a cp3XX wheel built against an
+      # alpha or a beta is a promise it cannot keep -- and PyPI does not take an
+      # upload back.
+      abiIsFrozen = version: !(lib.hasInfix "a" version || lib.hasInfix "b" version);
+
+      releasableWheelNames = map ({ pythonMinor, platformName }: "${pythonMinor}-${platformName}") (
+        lib.filter ({ pythonMinor, ... }: abiIsFrozen matrix.pythons.${pythonMinor}.version) wheelIds
+      );
+
       # Only what a wheel is built from, so editing the README or the tests does
       # not invalidate every cross build.
       buildSource = lib.fileset.toSource {
@@ -102,7 +112,11 @@
             paths = lib.attrValues wheels;
           };
 
-          named = lib.mapAttrs' (name: wheel: lib.nameValuePair "wheel-${name}" wheel) wheels;
+          # Nix splits an attribute path on `.`, so `wheel-3.14-*` reaches the
+          # parser as `wheel-3` and cannot be built by name.
+          named = lib.mapAttrs' (
+            name: wheel: lib.nameValuePair "wheel-${lib.replaceStrings [ "." ] [ "" ] name}" wheel
+          ) wheels;
 
           jphfmt = pkgs.callPackage ./nix/jphfmt.nix { };
 
@@ -110,11 +124,12 @@
 
           sdist = pkgs.callPackage ./nix/sdist.nix { src = buildSource; };
 
-          # What a release uploads: every wheel and the one sdist, in the shape
-          # `twine upload` wants.
+          # What a release uploads: every releasable wheel and the one sdist, in
+          # the shape `twine upload` wants. A pre-release interpreter's wheels
+          # are still built and still tested, just never published.
           release = pkgs.symlinkJoin {
             name = "salix-release";
-            paths = lib.attrValues wheels ++ [ sdist ];
+            paths = map (name: wheels.${name}) releasableWheelNames ++ [ sdist ];
           };
         in
         {
--- a/flake.nix
+++ b/flake.nix
@@ -48,10 +48,11 @@
       # PyHeapTypeObject's layout directly, so a cp3XX wheel built against an
       # alpha or a beta is a promise it cannot keep -- and PyPI does not take an
       # upload back.
-      # Matched at the pre-release suffix rather than by looking for the letters
-      # anywhere, so a version string that happens to contain one is not read as
-      # an alpha. `rc` is frozen and deliberately absent.
-      abiIsFrozen = version: builtins.match "[0-9.]+(a|b)[0-9]+" version == null;
+      # Matched against the shapes whose ABI *is* frozen -- a final release, or
+      # an rc, which is where CPython freezes it -- so anything unrecognised is
+      # held back rather than published. The other direction fails open, and an
+      # upload to PyPI is not something a later commit can undo.
+      abiIsFrozen = version: builtins.match "[0-9]+\\.[0-9]+\\.[0-9]+(rc[0-9]+)?" version != null;
 
       releasableWheelNames = map ({ pythonMinor, platformName }: "${pythonMinor}-${platformName}") (
         lib.filter ({ pythonMinor, ... }: abiIsFrozen matrix.pythons.${pythonMinor}.version) wheelIds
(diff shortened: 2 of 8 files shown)


JPHutchins:
Verification I owed this PR: #32's claim is about the **sdist**, and I had changed `build_config.py` and `setup.py` without ever building one. Done now, end to end.

```console
$ nix build .#sdist                                    # salix-0.0.0.tar.gz, 24171 bytes
$ pip install --no-binary :all: salix-0.0.0.tar.gz     # no SALIX_STRICT, isolated env
$ python -c "import salix; ..."
built from sdist -> P(x=1, y='two') | fields ('x', 'y')
```

And the flags the stranger's compiler actually received, counted from the verbose build:

```
20  -Wall
10  -std=c23
 0  -Werror
 0  -g0
```

Ten translation units, `-std=c23` on each, and `-Werror` absent — which is the whole of #32. `-g0` is absent too, correctly: it belongs to the cross builds that produce published wheels, not to someone compiling on their own machine, where debug information is theirs to want.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for the pre-publish issues to be resolved in batches; this verifies a claim in this PR that I had asserted without measuring.

JPHutchins:
Six of seven taken. `camas check` green at 25/25.

**1 (major) — `-std=c23` breaks sdist builds on GCC < 14.** Right, and it undoes something this PR did. Confirmed the two spellings are the same language mode, on GCC 15:

```
-std=c2x   __STDC_VERSION__ = 202311
-std=c23   __STDC_VERSION__ = 202311
```

`-std=c2x` is accepted by GCC 9+ and Clang 9+; `-std=c23` needs GCC 14+ or Clang 18+. So the rename bought nothing and cost `pip install salix` from an sdist on Ubuntu 24.04 LTS — the exact population the `-Werror` opt-in **in this same PR** exists to protect. Reverted rather than probed: a build-time compiler probe is machinery in service of a flag spelling.

#25 asked for the rename. #25 did not know this; I will note it there.

**2 — `SALIX_STRICT` truthiness.** Real: `SALIX_STRICT=0` turned `-Werror` **on**. Compared against `"1"` now, and verified `SALIX_STRICT=0` yields `()`.

**3 — `--shipped` alone gave a generic usage line.** Now `build_config.py: --shipped only follows --strict`.

**5 — `compile_flags.py` is unconditionally strict.** The asymmetry is deliberate — that database feeds clangd, clang-tidy and the analyzer, which are this repository's own tools and never a stranger's build — and it was undocumented, which was the finding. Written down now.

**6 — `abiIsFrozen` substring matching.** Taken, matched at the pre-release suffix. I checked the two predicates against every pinned interpreter rather than assuming the release set was unchanged:

```
  3.10   3.10.20   old=True  new=True  AGREE      3.14   3.14.6    old=True  new=True  AGREE
  3.11   3.11.15   old=True  new=True  AGREE      3.14t  3.14.6    old=True  new=True  AGREE
  3.12   3.12.13   old=True  new=True  AGREE      3.15   3.15.0b4  old=False new=False AGREE
  3.13   3.13.14   old=True  new=True  AGREE      3.15t  3.15.0b4  old=False new=False AGREE
  3.13t  3.13.14   old=True  new=True  AGREE
```

Zero disagreements, so the 40-path release output is unchanged. `rc` is frozen and stays out of the pattern.

**7 — `STRICTLY`.** Renamed to `STRICT_FLAGS`.

<details>
<summary><b>4 — declined: <code>tuple(["-Werror"])</code> to survive a lost trailing comma</b></summary>

The risk is real but already caught, and at both use sites. Removing the comma:

```
build_config.py:59: error: No overload variant of "__add__" of "tuple" matches argument type "str"  [operator]
build_config.py:61: error: No overload variant of "__add__" of "tuple" matches argument type "str"  [operator]
```

`camas check`'s `tooling` leaf runs `mypy --strict` over `build_config.py` on every check, so the comma is *proved* rather than defended, and the idiomatic single-element tuple stays. `tuple([...])` would trade the idiom for a guard the type checker already provides.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
All three taken. `camas check` green at 25/25.

**3 (the nit, and the one that mattered most) — `abiIsFrozen` fails open.** Correct, and the finding's own example proves it. `3.16.0b1.post1` does *not* match `[0-9.]+(a|b)[0-9]+` — the trailing `.post1` breaks it — so the old predicate called it frozen and would have **published a beta to PyPI**. The round-1 version I replaced had the same hole.

Inverted to match the shapes that *are* frozen — a final release, or an rc, which is where CPython freezes the ABI — so anything unrecognised is held:

```
  3.14.6          publish      3.16.0a1         HELD
  3.15.0rc1       publish      3.16.0b1.post1   HELD
  3.16.0rc2       publish      weird            HELD
  3.15.0b4        HELD         (empty)          HELD
```

All nine pinned interpreters answer exactly as before — 3.15 and 3.15t held, the rest published — so the release set is unchanged and only the unknown-format direction moved.

**1 — an all-pre-release matrix degrades to an sdist-only release.** Taken, and closed at the derivation rather than in the workflow. An sdist alone is not a release: it makes every user on every platform compile. The `cp` glob failing is the right *outcome* but a terrible explanation, so:

```nix
assert lib.assertMsg (
  releasableWheelNames != [ ]
) "no releasable wheels: every pinned interpreter is a pre-release, so nothing has a frozen ABI";
```

`nix build .#release` now names the reason instead of the workflow reporting a glob that matched nothing.

**2 — `base-wheel.nix` lost `-Werror`.** Right, and worth keeping even though the payload is discarded: the compile still happens, so it is a free check on the host compiler that this branch silently dropped when it made `-Werror` opt-in. `SALIX_STRICT = "1"` on the derivation, with the reasoning recorded there.

> [!WARNING]
> **LLM Disclosure**
>

Issue #21, opened 2026-07-31
Two class-creation calls discard a -1 error return and continue with an exception set


JPHutchins:
Two calls in the class-creation path discard a `-1` error return and carry on. Neither can fire today — both operate on `str` keys in a plain dict — but both leave an exception set for an unrelated later call to trip over, and one of them diverges from the pattern the same codebase already uses correctly elsewhere.

**`inherits_field`** — [src/fields.c:161](https://github.com/JPHutchins/salix/blob/74d2647/src/fields.c#L161):

```c
	for (Py_ssize_t i = 0; i < inherited_count; ++i) {
		if (PyUnicode_Compare(field_name, PyTuple_GET_ITEM(base->struct_field_names, i)) == 0) {
			return true;
		}
	}
```

`PyUnicode_Compare` returns `-1` both for "less than" and for failure. A failure here is read as "not equal", the name is appended as a new field, and `StructMeta_new` can return a class with an exception still set — which CPython reports at whatever unrelated point next checks, or as `returned a result with an exception set`.

`find_field` in [src/construct.c:276](https://github.com/JPHutchins/salix/blob/74d2647/src/construct.c#L276) handles the identical call correctly:

```c
		if (compared == -1 && PyErr_Occurred()) {
			return (struct field_lookup){.tag = FIELD_LOOKUP_ERROR};
		}
```

so the codebase already knows the shape. This call site is the one that diverges.

**`drop_class_variables`** — [src/meta.c:392](https://github.com/JPHutchins/salix/blob/74d2647/src/meta.c#L392):

```c
		if (
			PyDict_Contains(namespace, field_name) == 1 &&
			PyDict_DelItem(namespace, field_name) < 0
		) {
			return RESULT_ERROR;
		}
```

`PyDict_Contains` returning `-1` is not `1`, so the `&&` short-circuits and the loop continues with the error set. Same consequence.

<details>
<summary><b>Why neither reproduces, and why they are still worth fixing</b></summary>

Both operate on interned `str` keys in a dict with no user-defined `__eq__` or `__hash__` in play, so the failure paths are unreachable from Python today. The claim here is not that a user can trigger them — it is that the code is one refactor away from being wrong in a way that surfaces somewhere else entirely, which is the most expensive kind of bug to find in a C extension.

[...]

Issue #10, opened 2026-07-31
A body-defined __eq__ does not clear __hash__, so equal structs hash differently


JPHutchins:
A class body that defines `__eq__` should lose `__hash__`. `bind_hash` writes `__hash__` into the namespace before `type.__new__` ever sees the body, so the rule never fires and salix produces a class where **equal values hash differently**.

```python
>>> class Custom(Struct):
...     x: int
...     def __eq__(self, other): return True

>>> Custom(1) == Custom(2)
True
>>> hash(Custom(1)) == hash(Custom(2))
False
>>> len({Custom(1), Custom(2)})
2
```

Two values that compare equal occupy two slots in a set. Any dict or set built from these is wrong, and no error is raised anywhere.

The plain-Python equivalent is protected:

```python
>>> class Plain:
...     def __eq__(self, o): return True
>>> Plain.__hash__ is None
True
```

<details>
<summary><b>Root cause</b></summary>

[src/meta.c:372](https://github.com/JPHutchins/salix/blob/74d2647/src/meta.c#L372):

```c
static enum result bind_hash(PyObject * const namespace, struct options const options) {
	if (PyDict_GetItemString(namespace, "__hash__") != NULL) {
		return RESULT_OK;
	}

	if (options.eq && !options.frozen) {
		return PyDict_SetItemString(namespace, "__hash__", Py_None) == 0 ? RESULT_OK : RESULT_ERROR;
	}

	static char const * const hash_name[] = {"__hash__"};

	return rebind(namespace, hash_name, options.eq);
}
```

The guard checks for `__hash__` in the body, not for `__eq__`. A body that defines `__eq__` and no `__hash__` falls through to `rebind`, which binds `_StructMixin.__hash__`. By the time `type.__new__` runs, `__hash__` is present in the namespace, so CPython's rule in `type_new_set_attrs` — *a class that defines `__eq__` without `__hash__` gets `__hash__ = None`* — sees nothing to do.

Verified:

```python
>>> Custom.__dict__["__hash__"]
<slot wrapper '__hash__' of 'salix._StructMixin' objects>
```

The `options.eq` flag is about the *generated* equality. A body-defined `__eq__` is a third state that neither `eq=True` nor `eq=False` describes, and the code currently treats it as the first.

</details>

<details>
<summary><b>What the right answer is</b></summary>

The comment above `bind_hash` states the intent precisely, and it is the right intent:

[...]

Issue #9, opened 2026-07-31
find_struct_base takes the first struct base, dropping fields and defeating frozen inheritance


JPHutchins:
`find_struct_base` returns the **first** `StructMeta` instance in `bases`, not the one that owns the layout. If a fieldless struct is listed first, every field of the later struct base is dropped from the field plan — and the frozen-inheritance constraint goes with them.

```python
>>> class Fieldless(Struct): pass
>>> class WithFields(Struct):
...     a: int
...     b: int

>>> class C(Fieldless, WithFields):
...     c: int
>>> C.__struct_fields__
('c',)
>>> C(1).a
AttributeError: 'C' object has no attribute 'a'
```

The slots for `a` and `b` still exist in the layout — nothing reaches them. `repr`, `==` and `hash` silently ignore two thirds of the object.

Reversing the base order fixes it, which is the tell: `class C(WithFields, Fieldless)` gives `('a', 'b', 'c')`.

**The same root cause defeats `frozen`**, which is the part that matters most:

```python
>>> class M(Fieldless, WithFields, frozen=False):
...     c: int
>>> m = M(1)
>>> m.a = 9              # accepted
>>> isinstance(m, WithFields)
True
```

A mutable subclass of a frozen struct now exists, and code holding a `WithFields`-typed reference can be handed it. [src/options.h:11-14](https://github.com/JPHutchins/salix/blob/74d2647/src/options.h#L11-L14) says that is precisely what `frozen` is there to prevent:

> Only `frozen` refuses to differ from the base, and only mutation earns that: a mutable subclass of a frozen one hands a writable object to everything holding a reference of the base's type.

<details>
<summary><b>Root cause</b></summary>

[src/meta.c:198](https://github.com/JPHutchins/salix/blob/74d2647/src/meta.c#L198):

```c
static StructType * find_struct_base(PyObject * const bases) {
	for (Py_ssize_t i = 0; i < PyTuple_GET_SIZE(bases); ++i) {
		PyObject * const base = PyTuple_GET_ITEM(bases, i);

		if (PyObject_TypeCheck(base, &StructMeta_Type)) {
			return (StructType *) base;
		}
	}

	return NULL;
}
```

First match wins. Its docstring says "Find the (single) struct base among `bases`" — the singleness is assumed, never checked. Two struct bases with fields *are* rejected, but by CPython, and only because their layouts conflict:

```python
>>> class D(A, B): pass
TypeError: multiple bases have instance lay-out conflict
```

[...]

PR #43, opened 2026-07-31
fix: three bugs in class creation

### LINKED ISSUE #8: A subclass that re-defaults an inherited field shadows the slot: obj.x and repr(obj) disagree
A subclass that gives an inherited field a new default leaves the default sitting in the class dict, where it shadows the base's slot descriptor. The instance then reports **two different values for the same field** depending on how you ask.

```python
>>> class A(Struct):
...     x: int = 1
>>> class B(A):
...     x: int = 2

>>> b = B(5)
>>> b.x            # read through the attribute
2
>>> repr(b)        # read through the slot
'B(x=5)'
```

`b.x` is not 5. The value the caller passed is in the slot, and `repr`, `==` and `hash` all read it correctly — attribute access is the one path that gets the wrong answer, and it is the path users actually use.

<details>

`build_class_namespace` deletes field names from the copied namespace so the class variable cannot collide with the `__slots__` descriptor of the same name. It only deletes the names in `new_names` — [src/meta.c:388](https://github.com/JPHutchins/salix/blob/74d2647/src/meta.c#L388):

```c
static enum result drop_class_variables(PyObject * const namespace, PyObject * const new_names) {
	for (Py_ssize_t i = 0; i < PyList_GET_SIZE(new_names); ++i) {
		PyObject * const field_name = PyList_GET_ITEM(new_names, i);
		...
```

An inherited field is deliberately *not* in `new_names` — it gets no new slot, only a new default. So `x = 2` survives into `B.__dict__`, and it is ahead of `A`'s member descriptor in the MRO.

The comment above the function states the invariant that is being missed:

> A field with a default is bound in the class body, where it would collide with the `__slots__` descriptor of the same name.

That is equally true when the descriptor belongs to a base.

`all_names` is already computed and already passed to `build_class_namespace`; passing it to `drop_class_variables` instead of `new_names` is the shape of the fix, but the default has to be captured into the field plan before the deletion — `append_declared` reads it out of the namespace at [src/fields.c:134](https://github.com/JPHutchins/salix/blob/74d2647/src/fields.c#L134), and `field_plan_build` runs before `build_class_namespace`, so the ordering already works.

</details>

<details>
<summary><b>The test that was supposed to catch this certifies it instead</b></summary>

[tests/test_inheritance.py:40](https://github.com/JPHutchins/salix/blob/74d2647/tests/test_inheritance.py#L40):

```python
def test_a_subclass_may_give_an_inherited_field_a_default():
    class Defaulted(Base):
        y: int = 5

    assert Defaulted(1).y == 5
```

`Defaulted(1).y == 5` holds whether `y` is read from the slot (which the constructor filled with the default) or from the leftover class variable. The two behaviours are indistinguishable at that call. One more assertion separates them:

```python
    assert Defaulted(1, 9).y == 9      # currently returns 5
```

`tests/test_free_threading.py:181` has the same blind spot: it annotates `z: int` and binds `z: i`, then asserts `subclass(1).z == i` — which passes off the class variable.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Closes #8, #10, #21. Three defects in `StructMeta.__new__`; two are silent and one is unreachable from Python.

### A subclass that re-defaults an inherited field reported two different values for it (#8)

```python
class A(Struct):
    x: int = 5

class B(A):
    x: int = 2

b = B()
b.x            # 2  -- the class variable, ahead of the base's descriptor in the MRO
repr(b)        # B(x=5)
```

`drop_class_variables` deletes field names from the copied namespace so a class variable cannot shadow the `__slots__` descriptor of the same name — but it was given `new_names`, and an inherited field is deliberately not in there: it gets a new default, not a new slot. So `x = 2` survived into `B.__dict__`. Attribute access was the one path that lied, and it is the path people use.

`all_names` is already computed and already passed to `build_class_namespace`, and `field_plan_build` captures the defaults before the deletion, so passing it through is the whole fix.

### A body-defined `__eq__` did not clear `__hash__` (#10)

Python's rule is that a class defining `__eq__` and not `__hash__` is unhashable. salix kept the tuple hash, so equal values hashed differently and occupied two slots in a set, with no error anywhere.

<details>
<summary><b>The trap, if this is ever revisited</b></summary>

Reading `__eq__` out of the namespace *inside* `bind_hash` breaks 47 tests, all of them `eq=False`. `apply_options` rebinds all six comparison names before calling `bind_hash`, so by then `__eq__` is present whether the body wrote one or salix did, and the check cannot tell them apart.

The flag is read at the top of `apply_options`, before any rebinding, where its presence still means the body:

```c
	/* Read before the rebinds below: afterwards every comparison name is
	 * present whether the body wrote one or salix did. */
	bool const body_defines_eq = PyDict_GetItemString(namespace, "__eq__") != NULL;
```

</details>

### Two calls discarded a `-1` error return and continued with an exception set (#21)

`PyUnicode_Compare` in `inherits_field` reads failure as "not equal" and appends the name as a new field; `PyDict_Contains` in `drop_class_variables` sat behind an `&&` that short-circuits past the error. Neither is reachable from Python — both operate on `str` keys in a plain dict — so neither gets a test. `inherits_field` returns a tri-state now, the shape `find_field` in `construct.c` already uses for the identical call.

---

Verified against the reproductions in the issues: `b.x` is 5 and `B().x` is still 2; `Custom.__hash__` is None and a set of two `Custom`s raises. `camas check` green at 25/25.

> [!WARNING]
> **LLM Disclosure**
>

### DIFF
--- a/src/fields.c
+++ b/src/fields.c
@@ -165,19 +165,22 @@ static enum result append_declared(
 	return RESULT_OK;
 }
 
+/* PyObject_RichCompareBool answers -1 for an error and nothing else, so the
+ * three cases are the return value itself -- no PyErr_Occurred() to tell a
+ * failure apart from a "less than", and so no invariant about the exception
+ * state on entry. The same tri-state compare.c reads into `enum comparison`. */
 static enum inheritance inherits_field(StructType const * const base, PyObject * const field_name) {
 	Py_ssize_t const inherited_count = base != NULL ? base->struct_field_count : 0;
 
 	for (Py_ssize_t i = 0; i < inherited_count; ++i) {
-		int const compared =
-			PyUnicode_Compare(field_name, PyTuple_GET_ITEM(base->struct_field_names, i));
+		enum inheritance const inherited = PyObject_RichCompareBool(
+			field_name,
+			PyTuple_GET_ITEM(base->struct_field_names, i),
+			Py_EQ
+		);
 
-		if (compared == 0) {
-			return INHERITANCE_INHERITED;
-		}
-
-		if (compared == -1 && PyErr_Occurred()) {
-			return INHERITANCE_ERROR;
+		if (inherited != INHERITANCE_NEW) {
+			return inherited;
 		}
 	}
 
--- a/src/fields.c
+++ b/src/fields.c
@@ -7,6 +7,12 @@
 #include "result.h"
 #include "types.h"
 
+enum inheritance : int {
+	INHERITANCE_ERROR = -1,
+	INHERITANCE_NEW = 0,
+	INHERITANCE_INHERITED = 1,
+};
+
 static enum result append_inherited(
 	StructType const * base,
 	PyObject * all_names,
@@ -22,7 +28,7 @@ static enum result append_declared(
 );
 static PyObject * build_defaults(PyObject * all_names, PyObject * default_by_name);
 static PyObject * checked_annotations(PyObject * namespace);
-static bool inherits_field(StructType const * base, PyObject * field_name);
+static enum inheritance inherits_field(StructType const * base, PyObject * field_name);
 
 /* The plan only takes references once every step has succeeded; the working
  * collections belong to this scope either way. */
@@ -142,8 +148,13 @@ static enum result append_declared(
 
 		/* Skip if this name was already inherited (override of annotation, not
 		 * a new slot). */
-		if (inherits_field(base, field_name)) {
-			continue;
+		switch (inherits_field(base, field_name)) {
+			case INHERITANCE_ERROR:
+				return RESULT_ERROR;
+			case INHERITANCE_INHERITED:
+				continue;
+			case INHERITANCE_NEW:
+				break;
 		}
 
 		if (PyList_Append(all_names, field_name) < 0 || PyList_Append(new_names, field_name) < 0) {
@@ -154,16 +165,23 @@ static enum result append_declared(
 	return RESULT_OK;
 }
 
-static bool inherits_field(StructType const * const base, PyObject * const field_name) {
+static enum inheritance inherits_field(StructType const * const base, PyObject * const field_name) {
 	Py_ssize_t const inherited_count = base != NULL ? base->struct_field_count : 0;
 
 	for (Py_ssize_t i = 0; i < inherited_count; ++i) {
-		if (PyUnicode_Compare(field_name, PyTuple_GET_ITEM(base->struct_field_names, i)) == 0) {
-			return true;
+		int const compared =
+			PyUnicode_Compare(field_name, PyTuple_GET_ITEM(base->struct_field_names, i));
+
+		if (compared == 0) {
+			return INHERITANCE_INHERITED;
+		}
+
+		if (compared == -1 && PyErr_Occurred()) {
+			return INHERITANCE_ERROR;
 		}
 	}
 
-	return false;
+	return INHERITANCE_NEW;
 }
 
 /* Build the defaults tuple as the trailing run of defaulted fields, and
--- a/src/meta.c
+++ b/src/meta.c
@@ -28,14 +28,15 @@ static void StructMeta_dealloc(PyObject * self);
 
 static StructType * find_struct_base(PyObject * bases);
 static bool has_weakref_slot(StructType const * base);
-static bool inherits_body_eq(StructType const * base);
+static struct options base_options(StructType const * base);
 static PyObject * build_class_namespace(
 	PyObject * original_namespace,
 	PyObject * all_names,
 	PyObject * new_names,
 	struct options options,
-	struct options inherited,
-	StructType const * base
+	StructType const * base,
+	bool body_defines_eq,
+	bool inherits_body_eq
 );
 static PyObject * build_slots(PyObject * new_names, bool weakref);
 static enum result set_match_args(PyObject * namespace, PyObject * all_names, bool wanted);
@@ -43,15 +44,15 @@ static enum result apply_options(
 	PyObject * namespace,
 	struct options options,
 	struct options inherited,
-	bool base_defines_eq
+	bool body_defines_eq,
+	bool inherits_body_eq
 );
 static enum result rebind(PyObject * namespace, char const * const * names, bool from_mixin);
 static enum result bind_hash(
 	PyObject * namespace,
 	struct options options,
-	struct options inherited,
 	bool body_defines_eq,
-	bool base_defines_eq
+	bool inherits_body_eq
 );
 static enum result drop_class_variables(PyObject * namespace, PyObject * all_names);
 static StructType * create_class(
@@ -64,7 +65,8 @@ static enum result install_fields(
 	StructType * struct_class,
 	StructType const * base,
 	struct field_plan const * plan,
-	struct options options
+	struct options options,
+	bool resolves_body_eq
 );
 static enum result install_post_init(StructType * struct_class);
 static bool defines_own_init(StructType const * struct_class);
@@ -160,7 +162,7 @@ PyObject * StructMeta_new(
 	}
 
 	StructType const * const base = find_struct_base(bases);
-	struct options const inherited = base != NULL ? base->struct_options : options_initial();
+	struct options const inherited = base_options(base);
 	struct options_request const request =
 		options_read(keywords, inherited, base != NULL && base->struct_field_count > 0);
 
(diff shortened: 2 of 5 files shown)


JPHutchins:
Both addressed — one taken structurally, one declined with a measurement. `camas check` green at 25/25.

**1 — `PyUnicode_Compare` and stale exception state.** Taken, but by removing the ambiguity rather than asserting around it. The finding is right that `compared == -1 && PyErr_Occurred()` is correct only while no exception is set on entry, and that nothing enforces it. `PyObject_RichCompareBool` answers `-1` for an error and nothing else, so the three cases *are* the return value:

```c
	for (Py_ssize_t i = 0; i < inherited_count; ++i) {
		enum inheritance const inherited = PyObject_RichCompareBool(
			field_name,
			PyTuple_GET_ITEM(base->struct_field_names, i),
			Py_EQ
		);

		if (inherited != INHERITANCE_NEW) {
			return inherited;
		}
	}
```

No `PyErr_Occurred()`, so no invariant to break. It is also the tri-state `compare.c` already reads into `enum comparison`, so this leaves one idiom in the codebase instead of two.

`find_field` in `construct.c` keeps the old pattern deliberately: it runs per keyword argument per *construction*, not once per field at class creation, and its identity scan already answers first for interned names. A hot path is not where to spend a 0.3-confidence hypothetical. Happy to change it if you disagree.

**2 — `default:` on the switch over `enum inheritance`.** Declined. CLAUDE.md is explicit:

> Don't add default case for switch statements on enums or tagged unions - the compiler will warn if a case is missing and adding a default case suppresses that useful warning.

The finding's own reasoning leans on `-Wswitch` as the safety net, and this build makes that net a hard stop. Adding a fourth enumerator and building:

```
src/fields.c:152:11: error: enumeration value 'INHERITANCE_PROBE' not handled in switch [-Werror,-Wswitch]
error: command '.../zig' failed with exit code 1
```

So the switch being exhaustive is proved at compile time, on every build, on every compiler in the matrix. A `default:` would trade that proof for silence — the enum could then gain a value and this switch would fall through it quietly, which is the outcome the finding is trying to prevent.

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Both taken, and the major one was a genuine hole in #10's fix. `camas check` green at 25/25.

**1 — an inherited body `__eq__` did not clear `__hash__`.** Confirmed exactly as described before touching anything:

```
A.__hash__      : None
B.__hash__      : <slot wrapper '__hash__' of 'salix._StructMixin' objects>
B(1,0) == B(2,0): True
hashes equal    : False
set size        : 2
```

`body_defines_eq` read only the direct body, so `B` fell through both gates and got a structural hash beside an `__eq__` that resolves to `A`'s. Same defect as #10, reached through inheritance.

<details>
<summary><b>My first attempt was the recommendation, and it was wrong — the suite caught it</b></summary>

Testing `base->tp_hash == PyObject_HashNotImplemented` broke an existing test, `test_a_child_of_a_fieldless_mutable_base_may_freeze_itself`. That test is right and the attempt was wrong.

A base is unhashable for **two** reasons, and only one survives into a child:

- its body defined `__eq__` — the child resolves that same `__eq__`, so the debt is inherited;
- it was mutable with structural equality — which does **not** survive, because `frozen` refusing to differ from its base only bites when the base has fields. A *fieldless* mutable base may be frozen by its child, and that child should hash structurally.

I had probed this and concluded it was unreachable, using a base with a field. That was the wrong probe.

So the question is now asked exactly rather than by proxy: *will this class resolve an `__eq__` that salix did not bind?* salix binds the mixin's or `object`'s; anything else came from a body, and it transitively covers grandchildren for free. `tp_hash` is kept as the **gate** rather than the answer, so the three lookups stay off the common path — a hashable base cannot be carrying one, and class creation is the number this project is careful about.

Eight cases checked, including the corner that separates the two reasons:

```
OK  the bug: child of a frozen body-eq base              None
OK    and its grandchild                                 None
OK  eq=False takes equality back                         identity
OK  the test I broke: frozen child of fieldless mutable  structural
OK  the corner tp_hash alone would miss                  None
OK  no regression: plain frozen subclass                 structural
OK  no regression: eq=False subclass                     identity
OK  no regression: mutable structural subclass           None
```

</details>

**2 — `drop_class_variables` over `all_names` broadens what gets deleted.** Taken, and it is worth naming as more than a scope change: it is #8's own symptom reached without an annotation.

```python
class Base(Struct):
    size: int

class Child(Base):
    size = 42          # no re-annotation
```

Before, `size = 42` survived into `Child.__dict__` ahead of the base's slot descriptor, so `Child(7).size` answered **42** while `repr` said `Child(size=7)` — attribute access lying, which is exactly what #8 is. Now the binding is dropped and the slot wins. Tested.

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Both taken. `camas check` green at 25/25.

**1 — `inherits_body_eq` swallowed an error and answered `false`.** Right, and the direction was the part that mattered. The branch is unreachable by the function's own comment, but if it were ever reached, `false` is the wrong safe answer:

```c
	/* Unreachable, and answered in the safe direction anyway: a class wrongly
	 * left unhashable says so the first time anyone hashes it, where one wrongly
	 * given a hash puts two equal instances in two slots of a set and says
	 * nothing. */
	if (inherited_eq == NULL || mixin_eq == NULL || object_eq == NULL) {
		PyErr_Clear();

		return true;
	}
```

That is the third of the three options offered, and I think it is the right one. Not `Py_UNREACHABLE()`: aborting the interpreter over a `getattr` that cannot fail is a worse trade than answering safely, and this is a class-creation path a user reaches by writing an ordinary class body. Not propagating either — a tri-state return and a new error path through `apply_options` is real plumbing in service of a branch that cannot execute.

**2 — the `drop_class_variables` comment was stale.** Correct, and it was the last thing still describing the pre-`all_names` behaviour. It now covers both shapes:

```c
/* Any class-body binding of a field name -- an annotated default, or a bare
 * assignment over a name the base already declared -- would sit in this class's
 * dict ahead of the __slots__ descriptor that reads the value. */
```

The bare-assignment shape is the one the previous round added `test_a_bare_binding_over_an_inherited_field_is_not_shadowed_either` for, so the comment was contradicting a test sitting next to it.

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Findings 1 and 2 turned out to be one change, and it deletes more than it adds. `camas check` green at 25/25.

**2 — store it on `StructType` instead of recomputing.** Taken. `inherits_body_eq` was asking, with three attribute lookups behind a `tp_hash` gate, a question the base already answered at its own creation. `StructType.struct_resolves_body_eq` now carries it: the function, the gate, the lookups and one threaded parameter are gone.

**1 (gap 1) — the `tp_hash` gate was wrong, not just slow.** This is the part worth having. A base that defines *both* `__eq__` and `__hash__` is hashable, so the gate returned early and the subclass got salix's structural hash beside an inherited `__eq__` that answers `True`:

```
A.__hash__         <function A.__hash__>        hash(A(1)) == 42
B.__hash__         <slot wrapper of _StructMixin>
B(1,0) == B(2,0)   True    hashes equal  False    set size  2
```

The replacement rule is simpler than the one it replaces: **an `__eq__` that came from a class body is not salix's to answer for, and neither is the hash beside it.** Bind nothing, and let the MRO carry whatever that body left — `None` from Python's own rule, or the `__hash__` the same body wrote. `B` inherits 42 now and two equal instances occupy one slot.

Nine cases verified, including the four this branch already had — the fieldless-mutable-frozen-child case that caught my first attempt at this still answers `structural`.

**1 (gap 2) — a non-struct base shadowing `__eq__`.** Reproduced, and filed as [#47](https://github.com/JPHutchins/salix/issues/47) rather than folded in:

```
B.__mro__[:3]   ['B', 'Mixin', 'Base']
B.__eq__ is Mixin.__eq__   True
len({B(1,0), B(2,0)})      2
```

`struct_resolves_body_eq` answers correctly for the struct chain, and `Mixin` is not in it. Closing that means scanning every base rather than the struct base — a different change, and one that shares a root with #9 (`find_struct_base` taking the first match), so it probably wants settling in the same pass.

**3 — the `build_class_namespace` comment.** Taken; it still said "default-bearing field names", which stopped being true when `drop_class_variables` moved to `all_names`. That is the second stale comment this change left behind and the second one you have caught.

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
No change this round: the single remaining finding is [#47](https://github.com/JPHutchins/salix/issues/47), which this review's own recommendation names as filed separately. We agree on both the defect and the scope.

To restate why it is out of this PR rather than deferred out of convenience: `struct_resolves_body_eq` answers correctly for the struct chain, and closing the mixin case means asking every base instead of the struct base. That is the same root as #9 — *what salix reads off "the base" is not always what the class gets* — and both should be settled in one pass rather than half-answered here.

This branch is complete from my side: #8, #10 and #21 fixed, five review rounds, and two defects found along the way that were not in the original issues — the inherited-body-`__eq__` hash (fixed here) and the shadowing mixin (#47).

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, closing out the automated review iteration on this PR.

Issue #2, opened 2026-07-29
object.__setattr__ on a frozen struct works only on 3.13+, so __post_init__ cannot portably derive a field


JPHutchins:
`__post_init__` can validate a frozen struct on every supported interpreter. It cannot *derive a field* on one, and whether it can depends on the interpreter — which is worse than a straight no.

## Measured

Same source, same struct, `object.__setattr__(instance, "y", 9)` inside `__post_init__`:

| interpreter | result |
| --- | --- |
| 3.10, 3.11, 3.12 | `TypeError: can't apply this __setattr__ to F object` |
| 3.13, 3.14, 3.15 | allowed |

## Why

CPython's setattr `hackcheck` walks the MRO backwards to whichever type defines the current `tp_setattro`, then walks that type's `tp_base` chain and refuses if it meets a non-Python type that overrides the function being called. For a frozen dataclass the defining type is `object`, so `object.__setattr__` is accepted. For a struct it is `_StructMixin`, whose slot *is* what freezing is. 3.13 dropped the check.

`tests/test_post_init.py::test_object_setattr_on_a_frozen_struct_follows_the_interpreter` asserts both branches, so this is pinned rather than latent.

## Consequence

Validation in `__post_init__` is portable. Deriving a field is not — that needs `frozen=False`. The frozen-dataclass escape hatch does not reach a struct on 3.10–3.12.

## What a fix would cost

Freezing would have to stop being a C slot on a static base and become something whose defining type is a Python class, so that `tp_setattro` is `slot_tp_setattro` and the hackcheck's walk finds nothing to refuse. Note that splitting the mixin into two static types does **not** work — I tried it and measured no change, because the check keys on the type that *defines* the slot, not on `tp_base` of the instance's type. A wrapper descriptor also enforces `__objclass__`, so the implementation has to stay reachable through the MRO.

It is a real design change, and it makes frozen bypassable on 3.10–3.12 the way it already is on 3.13+ and the way a frozen dataclass always is. That is the decision: uniformly bypassable, or non-portable as today.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to open issues for the work left outstanding after implementing the dataclass-style class options. This one is a design decision I deliberately did not make on her behalf.

JPHutchins:
## Design review — four ways out, and what each one costs

[...]

PR #45, opened 2026-07-31
fix: write a field the way CPython writes one

### LINKED ISSUE #7: set_field is not free-threading safe and segfaults on 3.14t

JPHutchins:
Closes #7. `set_field` crashed a free-threaded interpreter. Eight threads writing one shared slot while four read it, three runs each, on `cpython-3.14.6+freethreaded`:

```console
before: exit=139  exit=139  exit=139        # SIGSEGV, core dumped
after:  survived  survived  survived        # gil enabled: False
```

`Py_XSETREF` is a load, a store and a decref. Two threads read the same previous value, both store, and both release it — one reference, two releases — and a reader between the store and the decref can incref a pointer that is about to be freed. `shared.value = v` on a mutable struct survives the identical load, which is what isolated the race to `set_field` alone.

### The fix inherits the guarantee instead of restating it

`PyMember_SetOne` is what makes the descriptor path safe: a critical section on the instance around an atomic store, with the release deferred past the end of it. salix calls it now.

```c
	PyMemberDef slot = {
		.name = "",
		.type = SLOT_MEMBER_TYPE,
		.offset = type->struct_slot_offsets[index],
	};

	return PyMember_SetOne((char *) self, &slot, value) == 0 ? RESULT_OK : RESULT_ERROR;
```

`src/salix.c` has claimed since before `set_field` existed that *"a free-threaded build's guarantees there are inherited rather than reimplemented"*. This is the change that makes the sentence true again.

<details>
<summary><b>Why not open-code the critical section</b></summary>

That was the issue's option 1 and its recommendation. It is a near-copy of `PyMember_SetOne`, and only *near*: `Py_BEGIN_CRITICAL_SECTION` is public API, but the `FT_ATOMIC_STORE_PTR_RELEASE` beside it is not — it lives behind `Py_BUILD_CORE`, so an extension gets a plain store and a version of the guarantee that has to be re-audited whenever CPython's own moves.

The issue's objection to delegating was "costs a member lookup per call". It does not: the `PyMemberDef` is built from the offset `type.__new__` already gave the field, so it describes a slot that exists and nothing is resolved by name.

`Py_T_OBJECT_EX` is spelled `T_OBJECT_EX` before 3.12, out of the `structmember.h` that `types.h` already includes for that era.

</details>

### The test that was missing

`tests/test_free_threading.py` covered concurrent construction, concurrent subclass creation, concurrent descriptor writes and concurrent reads — and never `set_field`, which is why the release went out green. It has that case now.

### Also from #25

Same file, and the reason the two travel together: `salix.c`'s header comment called the project "struct", called itself a seed implementation of a feasibility prototype, and credited freezing to "read-only members for a frozen class" — a mechanism that was option D in #2 and was reverted before it ever landed. Freezing is `_StructMixin`'s `tp_setattro`. The README points readers at this comment, so it is worth being right.

`camas check` green at 25/25.

> [!WARNING]
> **LLM Disclosure**
>

### DIFF
--- a/src/construct.c
+++ b/src/construct.c
@@ -12,15 +12,6 @@ struct field_lookup {
 	Py_ssize_t index;
 };
 
-/* What type.__new__ makes a __slots__ entry. The Py_-prefixed spellings arrived
- * with 3.12; before that the name lives in structmember.h, which types.h pulls
- * in for exactly this reason. */
-#if PY_VERSION_HEX < 0x030C0000
-enum { SLOT_MEMBER_TYPE = T_OBJECT_EX };
-#else
-enum { SLOT_MEMBER_TYPE = Py_T_OBJECT_EX };
-#endif
-
 static void bind_positional(
 	StructType const * type,
 	PyObject * self,
@@ -281,8 +272,7 @@ PyObject * Struct_set_field(PyObject * const module, PyObject * const arguments)
  * the instance and defers the release past the end of it.
  *
  * The offset is the one type.__new__ gave this field, so the descriptor here
- * describes a slot that already exists; nothing is resolved by name, which is
- * the cost the alternative would have carried.
+ * describes a slot that already exists rather than looking one up.
  */
 static enum result write_slot(
 	StructType const * const type,
@@ -290,8 +280,15 @@ static enum result write_slot(
 	Py_ssize_t const index,
 	PyObject * const value
 ) {
+	char const * const name =
+		PyUnicode_AsUTF8(PyTuple_GET_ITEM(type->struct_field_names, index));
+
+	if (name == NULL) {
+		return RESULT_ERROR;
+	}
+
 	PyMemberDef slot = {
-		.name = "",
+		.name = name,
 		.type = SLOT_MEMBER_TYPE,
 		.offset = type->struct_slot_offsets[index],
 	};
--- a/src/construct.c
+++ b/src/construct.c
@@ -12,6 +12,15 @@ struct field_lookup {
 	Py_ssize_t index;
 };
 
+/* What type.__new__ makes a __slots__ entry. The Py_-prefixed spellings arrived
+ * with 3.12; before that the name lives in structmember.h, which types.h pulls
+ * in for exactly this reason. */
+#if PY_VERSION_HEX < 0x030C0000
+enum { SLOT_MEMBER_TYPE = T_OBJECT_EX };
+#else
+enum { SLOT_MEMBER_TYPE = Py_T_OBJECT_EX };
+#endif
+
 static void bind_positional(
 	StructType const * type,
 	PyObject * self,
@@ -31,6 +40,12 @@ static enum result fill_defaults(
 	Py_ssize_t positional_count
 );
 static struct field_lookup find_field(StructType const * type, PyObject * name);
+static enum result write_slot(
+	StructType const * type,
+	PyObject * self,
+	Py_ssize_t index,
+	PyObject * value
+);
 static enum result run_post_init(StructType const * type, PyObject * self);
 
 /*
@@ -199,9 +214,10 @@ static enum result fill_defaults(
  * safety argument: the escape hatch reaches exactly what the class already
  * spells out, and nothing else.
  *
- * Intended for __post_init__, before the instance has escaped. Writing one that
- * another thread already holds is the caller's problem, as it is for any object
- * whose invariants outlive its constructor.
+ * Intended for __post_init__, before the instance has escaped. Which value a
+ * racing write leaves behind is the caller's problem, as it is for any object
+ * whose invariants outlive its constructor -- but only which value: the write
+ * itself is as safe as `self.x = v`, and for the same reason.
  */
 PyObject * Struct_set_field(PyObject * const module, PyObject * const arguments) {
 	PyObject * self = NULL;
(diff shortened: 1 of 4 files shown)


JPHutchins:
All four taken. `camas check` green at 25/25 across 3.10–3.15 and 3.14t.

**1 — the fabricated `PyMemberDef` named the field `""`.** Taken, and the finding's framing is the right one: it worked because `PyMember_SetOne` reads `.name` only for error messages, which is a dependency on CPython not looking — the class of thing this entire change exists to remove. The descriptor names the member it describes now:

```c
	char const * const name =
		PyUnicode_AsUTF8(PyTuple_GET_ITEM(type->struct_field_names, index));
```

`PyUnicode_AsUTF8` caches on the `str`, and `struct_field_names` was already in hand, so the per-call cost is a pointer read after the first. `.name` is `const char *` on every version in the matrix — checked 3.10 and 3.14 headers — so no cast.

**2 — "exactly two paths" is brittle.** Correct. It states the guarantee rather than the count now: after construction, an instance's slots are written *only* through `PyMember_SetOne`. The mechanism was always documented at `write_slot`, which is where it belongs.

**3 — the test could not detect a no-op `set_field`.** Correct and free. Seeded with `[0]` instead of `[0, 0]`, so `assert len(shared.value) == 2` now fails if the write stops happening rather than passing on the value it started with.

**4 — `SLOT_MEMBER_TYPE` belongs in `types.h`.** Taken. Both halves of the 3.12 rename — the `structmember.h` include and the constant's `Py_` prefix — are one `#if` block now, so the whole adaptation goes away together whenever the floor reaches 3.12 rather than leaving half of it behind in `construct.c`.

<details>
<summary><b>Re-measured, since finding 1 changes the write path itself</b></summary>

Three runs on `cpython-3.14.6+freethreaded`, eight writers and four readers on one shared slot:

```
exit=0 gil enabled: False survived
exit=0 gil enabled: False survived
exit=0 gil enabled: False survived
```

Unchanged from the pre-review measurement, and still `exit=139` three times on the `Py_XSETREF` this replaced.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Thank you for this one — it is a **crash**, not a minor observation, and it is now [#46](https://github.com/JPHutchins/salix/issues/46).

Measured rather than reasoned about. Four writers, three `repr()` readers and three `==` readers on one shared struct, `cpython-3.14.6+freethreaded`:

```
exit=139   exit=139   exit=139        # SIGSEGV, 3/3
```

**But it is not this PR's.** The same script with `frozen=False` and `shared.value = [tag, i]` — CPython's own member descriptor doing the writing, the path this PR held up as already-safe — crashes identically:

```
descriptor-write exit=139
descriptor-write exit=134
descriptor-write exit=134
```

So the defect is in salix's readers, not in either writer, and it predates #7. `PyMember_SetOne` defers its release past the critical section, which is what makes two *writers* safe; a reader takes no critical section at all, so the window between its load and its incref is unprotected either way.

Filed separately rather than folded in, because this PR's scope is the write path and closing the read side means touching `repr`, `hash` and both comparison paths — four sites and a regression test of their own. #46 records the measurement and argues for reading through `PyMember_GetOne`, symmetric with the `PyMember_SetOne` this PR introduced, so that the two halves have one owner.

Worth noting the finding also corrects something in the merged #30: taking owned references in `ordering_result` closed a use-after-free *within* a comparison, and does not close this one — `Py_NewRef` on an already-freed pointer is itself the race.

> [!WARNING]
> **LLM Disclosure**
>

Issue #24, opened 2026-07-31
Writing to a frozen struct raises TypeError, not AttributeError


JPHutchins:
Writing to a frozen struct raises `TypeError`. Every comparable thing in Python raises an `AttributeError` subclass, so `except AttributeError:` — the natural way to write this — does not catch it.

```python
>>> class P(Struct):
...     x: int
>>> P(1).x = 2
TypeError: P object does not support attribute assignment
```

| what | exception | catchable as `AttributeError`? |
| --- | --- | --- |
| frozen `salix.Struct` | `TypeError` | **no** |
| frozen `dataclass` | `dataclasses.FrozenInstanceError` → `AttributeError` | yes |
| a read-only `property` | `AttributeError` | yes |
| a `__slots__` class, undeclared name | `AttributeError` | yes |
| a `namedtuple` field | `AttributeError` | yes |

salix already raises `AttributeError` for the adjacent case, which is the inconsistency:

```python
>>> object.__setattr__(P(1), "zz", 1)
AttributeError: 'P' object has no attribute 'zz'
```

So "you may not write this field" is a `TypeError` while "there is no such field" is an `AttributeError`.

<details>
<summary><b>Where it is raised, and the smaller wart alongside it</b></summary>

[src/mixin.c:43](https://github.com/JPHutchins/salix/blob/74d2647/src/mixin.c#L43):

```c
	PyErr_Format(
		PyExc_TypeError,
		"%.200s object does not support attribute %s",
		Py_TYPE(self)->tp_name,
		value == NULL ? "deletion" : "assignment"
	);
```

The message also omits the quotes CPython puts around a type name everywhere else — `P object ...` where CPython writes `'P' object ...`. One character each side, and it makes salix's errors read as not-quite-Python.

Neither the type name nor the *field* name appears in a form a caller can act on; `dataclasses` says `cannot assign to field 'x'`, which is more useful. `Struct_set_attribute` has `name` in hand and does not use it.

</details>

<details>
<summary><b>Why this is on the pre-publish list</b></summary>

The exception type of a documented failure is API. Changing `TypeError` to `AttributeError` after v1 breaks anyone who wrote `except TypeError:`; leaving it means salix is permanently the odd one out, and every user who writes the obvious `except AttributeError:` gets a traceback instead.

Three shapes are available:

1. **`AttributeError`**, matching `__slots__` and `property`. Simplest, and it makes the two adjacent failures consistent.
[...]

Issue #20, opened 2026-07-31
inspect.signature returns (), and the module has no __version__ or docstrings


JPHutchins:
`inspect.signature` on a struct class returns an empty signature, so every tool that reads a callable's signature — IDE completion, `help()`, docs generators, CLI builders, FastAPI-style frameworks — sees a constructor that takes nothing.

```python
>>> class Point(Struct):
...     x: float
...     y: float = 0.0

>>> inspect.signature(Point)
<Signature ()>
```

A dataclass gives `(x: float, y: float = 0.0)` because its `__init__` is a real function. NamedTuple does too. salix constructs through a vectorcall with no `__init__` and no `__signature__`, so there is nothing for `inspect` to read.

The information exists — `Point.__struct_fields__` is `('x', 'y')`, `Point.__struct_defaults__` is `(0.0,)`, and the annotations are in `Point.__annotations__`. Nothing exposes it in the shape `inspect` wants.

<details>
<summary><b>Related gaps in the same area</b></summary>

| probe | result |
| --- | --- |
| `inspect.signature(Point)` | `()` |
| `salix.Struct.__doc__` | `None` |
| `salix.StructMeta.__doc__` | `None` |
| `salix.__version__` | absent |
| `salix.__all__` | absent |

`help(salix.Struct)` therefore shows the inherited mixin wrappers and nothing about what a struct is. For the module that *is* the public API of a published package, that is thin.

`__version__` is the one with a concrete cost: `importlib.metadata.version("salix")` works, but a great deal of tooling and a great many users reach for `salix.__version__` first. A C extension can take it from a `-D` at build time, which keeps `pyproject.toml` as the single source.

`__all__` matters less — the module exports only `Struct`, `StructMeta` and `set_field`, none underscore-prefixed, so `import *` already does the right thing. Worth adding only if #15 decides `StructMeta` should not be public.

Class docstrings are a live tension with CLAUDE.md, which says docstrings must never restate the source. A docstring on `Struct` saying what a struct *is* — frozen by default, fields from annotations, no serialisation — is not a restatement of any line of C, so it looks like it clears that bar. Worth a decision rather than an assumption.

</details>

<details>
<summary><b>Shape of a fix for the signature</b></summary>

[...]

Issue #12, opened 2026-07-31
A body-defined __new__ and a body-defined __slots__ are both silently discarded


JPHutchins:
Two things a class body can write are discarded without a word. Everywhere else this design **rejects rather than ignores** — `kw_only=True` is refused precisely so it cannot be silently accepted, and `tests/test_options.py` has a test saying so. These two paths ignore.

**A body-defined `__new__` never runs:**

```python
>>> class A(Struct):
...     x: int
...     def __new__(cls, *args, **kwargs):
...         print("__new__ RAN")
...         return super().__new__(cls)
>>> A(1)
A(x=1)                    # nothing printed
```

**A body-defined `__slots__` is thrown away:**

```python
>>> class R(Struct):
...     __slots__ = ("extra",)
...     x: int
>>> R.__slots__
('x',)
>>> R(1).extra = 1
AttributeError: 'R' object has no attribute 'extra'
```

This is the exact failure mode `0eb9a2c` fixed for `__init__`. From that commit's test file: *"a body that defined one used to have it discarded without a word"*. `__new__` still has it.

<details>
<summary><b>Root cause: <code>__new__</code></b></summary>

`install_fields` decides between two constructor strategies at [src/meta.c:451](https://github.com/JPHutchins/salix/blob/74d2647/src/meta.c#L451):

```c
	if (defines_own_init(struct_class)) {
		struct_class->heap_type.ht_type.tp_new = PyType_GenericNew;
	} else {
		struct_class->heap_type.ht_type.tp_vectorcall = Struct_vectorcall;
	}
```

Both branches lose a body `__new__`:

- the vectorcall branch allocates directly with `python_class->tp_alloc` ([src/construct.c:64](https://github.com/JPHutchins/salix/blob/74d2647/src/construct.c#L64)) and never consults `tp_new`;
- the `__init__` branch *overwrites* `tp_new` with `PyType_GenericNew`, clobbering the slot `type.__new__` had just installed from the body.

`defines_own_init` is careful about exactly this problem for `__init__`, and its comment explains why:

> A body that writes its own `__init__` means it. The generated constructor is what a struct gets, not what it is stuck with, and leaving the vectorcall installed would discard the definition in silence.

The same sentence applies to `__new__` word for word.

[...]

PR #48, opened 2026-08-01
fix: ask whether the object is actually a struct

### LINKED ISSUE #15: StructMeta is public but produces a broken class when used as a bare metaclass
`StructMeta` is exported from the module and declared in the stub, so it is public API. Used as a bare metaclass — without `Struct` in the bases — it produces a class that is quietly broken in three ways at once.

```python
>>> class C(metaclass=salix.StructMeta):
...     x: int
...     y: int = 3

>>> hash(C(1))
TypeError: descriptor '__hash__' requires a 'salix._StructMixin' object but received a 'C'

>>> repr(C(1))
'<__main__.C object at 0x...>'      # no structural repr

>>> C(1) == C(1)
False                               # identity, not structural

>>> c = C(1); c.x = 99              # accepted, despite frozen defaulting to True
```

The class builds, constructs, and reports `__struct_fields__` correctly. It just is not a struct.

<details>

Two mechanisms interact.

`bind_hash` binds `_StructMixin.__hash__` into the namespace of a class that has no mixin in its MRO — [src/meta.c:383](https://github.com/JPHutchins/salix/blob/74d2647/src/meta.c#L383), via `rebind` at [src/meta.c:345](https://github.com/JPHutchins/salix/blob/74d2647/src/meta.c#L345):

```c
	PyObject * const source = (
		from_mixin ? (PyObject *) &StructMixin_Type :
		(PyObject *) &PyBaseObject_Type
	);
```

The slot wrapper it fetches carries a type check, which is what produces that `TypeError` at call time rather than at class-creation time.

The other three symptoms come from `apply_options` only acting on *transitions* — [src/meta.c:318](https://github.com/JPHutchins/salix/blob/74d2647/src/meta.c#L318):

```c
	if (options.eq != inherited.eq && rebind(namespace, comparison, options.eq) != RESULT_OK) {
```

With no struct base, `inherited` is `options_initial()` and the resolved options equal it, so no transition fires and nothing gets bound. The class inherits `object`'s `__eq__`, `__repr__` and `__setattr__` — hence identity equality, the default repr, and a `frozen=True` class that accepts assignment.

This is the correct behaviour *for a class that inherits the mixin*, where "no transition" means "keep what the base already has". Without the mixin there is nothing to keep.

</details>

<details>

**1. Reject it.** `StructMeta.__new__` refuses bases that do not include `_StructMixin`, with a message pointing at `Struct`. This is the honest position if `StructMeta` exists only to be `type(Struct)` — which, reading the code, it does. It would also close the `StructMeta("X", (), {})` hole, where the same broken class is one call away.

**2. Make it work.** Insert `_StructMixin` into the bases when it is absent. More surface area, and it changes what `class C(metaclass=StructMeta)` means for anyone relying on the current MRO.

**3. Stop exporting it.** The stub needs `StructMeta` for `Struct`'s `metaclass=`, but the *runtime* module does not have to publish it — [src/salix.c:93](https://github.com/JPHutchins/salix/blob/74d2647/src/salix.c#L93) adds it deliberately. `type(MyStruct)` reaches it anyway, so this only removes the invitation.

Option 1 looks right, possibly with 3. Whichever way it goes, it should be decided **before** the first PyPI upload: `StructMeta` is in `__init__.pyi` today, so removing or constraining it later is a breaking change.

</details>

> [!WARNING]
> **LLM Disclosure**
>

### LINKED ISSUE #28: Subclassing _StructMixin with a co-base segfaults: a struct's own __eq__ is the trigger
`_StructMixin` carries `Py_TPFLAGS_BASETYPE`, and every slot it installs reinterprets `Py_TYPE(self)` as a `StructType *`. A subclass of the mixin whose metaclass is plain `type` is not a `StructType`, so those reads run past the end of the heap-type allocation. **The interpreter segfaults**, and the crashing call is an ordinary struct's own `__eq__`.

```python
>>> import salix
>>> Mixin = salix.Struct.__mro__[1]              # salix._StructMixin, handed out by the MRO
>>> F = type("F", (Mixin, list), {"__slots__": ("zzzzz",)})
>>> f = F()

>>> class P(salix.Struct):
...     x: int
>>> P(1) == f
Segmentation fault (core dumped)
```

`f.__struct_fields__` and `f.__struct_defaults__` crash the same way. No ctypes, no private attributes — the mixin is reachable from `Struct.__mro__`.

The co-base matters: `type("F", (Mixin,), {})` alone gives `TypeError: cannot create 'F' instances`, because a heap subclass inherits the mixin's NULL `tp_new`. Adding any base that supplies a working `tp_new` — `list` here — defeats that. **A first look at this stops at the bare subclass and concludes the mixin is closed off; it is not.**

<details>

[src/mixin.c:15-25](https://github.com/JPHutchins/salix/blob/74d2647/src/mixin.c#L15-L25) installs four struct-aware entry points on a subclassable static type:

```c
PyTypeObject StructMixin_Type = {
	PyVarObject_HEAD_INIT(NULL, 0)
	.tp_name = "salix._StructMixin",
	.tp_basicsize = sizeof(PyObject),
	.tp_flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
	.tp_setattro = Struct_set_attribute,
	.tp_repr = Struct_repr,
	.tp_hash = Struct_hash,
	.tp_richcompare = Struct_rich_compare,
	.tp_getset = Struct_getset,
};
```

All of them reach the field table through [src/types.h:37](https://github.com/JPHutchins/salix/blob/74d2647/src/types.h#L37), which is an unchecked downcast:

```c
static inline StructType * struct_type_of(PyObject * const self) {
	return (StructType *) Py_TYPE(self);
}
```

The only guard anywhere is in [src/compare.c:26](https://github.com/JPHutchins/salix/blob/74d2647/src/compare.c#L26), and it tests the wrong thing:

```c
	if (!PyObject_TypeCheck(other, &StructMixin_Type)) {
		Py_RETURN_NOTIMPLEMENTED;
	}
```

That admits any mixin subclass. [src/compare.c:107-111](https://github.com/JPHutchins/salix/blob/74d2647/src/compare.c#L107-L111) then reads `other_type->struct_field_names` off a plain `PyHeapTypeObject`, and [src/types.h:82](https://github.com/JPHutchins/salix/blob/74d2647/src/types.h#L82) does `Py_NewRef` on whatever came back — a refcount **write** to an address that is whatever happened to be in that memory.

`repr`, `hash` and `f == f` happen to survive for this particular layout: the out-of-bounds `struct_field_count` lands on a `PyMemberDef` field `type_new` left zero, and self-comparison short-circuits on pointer identity before any dereference. That is luck, not a guard, and a different layout gives a different answer.

</details>

<details>

`Struct_set_field` already knows the correct test — [src/construct.c:215](https://github.com/JPHutchins/salix/blob/74d2647/src/construct.c#L215):

```c
	if (!PyObject_TypeCheck((PyObject *) Py_TYPE(self), &StructMeta_Type)) {
```

"Is this object's *type* a `StructMeta` instance" is the question every one of the mixin's slots needs to answer, and none of them do. Two ways to close it:

1. **Check in `struct_type_of`'s callers.** `Struct_rich_compare` swaps its `StructMixin_Type` test for the `StructMeta_Type` one on both operands (returning `NotImplemented` for `other`); `Struct_repr`, `Struct_hash` and the two getsets each need the same check on `self`. Five call sites, and the cost is one type check on paths that already do far more.
2. **Drop `Py_TPFLAGS_BASETYPE` from the mixin** and give `Struct` its dunders another way. Removes the vector entirely, but `Struct` is built by `create_struct_base` with the mixin as its base ([src/salix.c:114](https://github.com/JPHutchins/salix/blob/74d2647/src/salix.c#L114)), so the flag is currently load-bearing.

Note that fixing only `__eq__` is not enough: `repr`, `hash` and both getsets read the same out-of-bounds region and merely happen to find a zero there today.

A test belongs next to this. It cannot assert the crash — a segfault takes pytest with it — so it has to assert the guarded outcome: `NotImplemented` from the comparison, and a clean `TypeError` from the getsets.

</details>

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for a whole-repository code review in preparation for publishing salix to PyPI, with the findings filed as issues for future agents to resolve. An earlier pass in the same review wrongly concluded the mixin was closed off, having tried only the bare subclass; #26 has been corrected.

### LINKED ISSUE #29: create_class casts PyType_Type.tp_new's result unchecked, so a StructMeta subclass can segfault
`create_class` casts whatever `PyType_Type.tp_new` returns straight to `StructType *` and never checks it. `type_new` does not always build the object itself — when the winning metatype differs from the one it was handed, it delegates to `winner->tp_new`, and a Python metaclass's `__new__` may return anything. `install_fields` then writes six fields at `StructType` offsets into that object.

```python
import salix

calls = []
class Probe(salix.StructMeta):
    def __new__(mcls, *a, **k):
        calls.append(1)
        if len(calls) == 1:
            return super().__new__(mcls, *a, **k)
        return "not a type"          # handed straight back to create_class

class P2(salix.Struct, metaclass=Probe):
    a: int

salix.StructMeta("Z", (P2,), {"__annotations__": {"b": int}})
```

```console
P2: <class '__main__.P2'> <class '__main__.Probe'>
delegating call -> create_class gets a str back
Segmentation fault (core dumped)
```

`bytearray(8192)` in place of the string crashes identically.

<details>

[src/meta.c:403-416](https://github.com/JPHutchins/salix/blob/74d2647/src/meta.c#L403-L416):

```c
static StructType * create_class(
	PyTypeObject * const metatype,
	PyObject * const name,
	PyObject * const bases,
	PyObject * const namespace
) {
	PY_OWNED(type_args, PyTuple_Pack(3, name, bases, namespace));

	if (type_args == NULL) {
		return NULL;
	}

	return (StructType *) PyType_Type.tp_new(metatype, type_args, NULL);
}
```

The NULL case is handled by the caller. Nothing checks that a non-NULL result is a type, let alone a `StructMeta` instance.

`resolve_slot_offsets` gets there first, reading `Py_SIZE(struct_class)` and `struct_heap_type_members(struct_class)` — the latter being `(char *) type + Py_TYPE(type)->tp_basicsize` ([src/types.h:88-90](https://github.com/JPHutchins/salix/blob/74d2647/src/types.h#L88-L90)) — off the bogus pointer. Then [src/meta.c:439-444](https://github.com/JPHutchins/salix/blob/74d2647/src/meta.c#L439-L444) writes through it:

```c
	struct_class->struct_field_names = py_move(&field_names);
	struct_class->struct_defaults = Py_NewRef(plan->defaults);
	struct_class->struct_slot_offsets = offsets;
	struct_class->struct_field_count = field_count;
```

`StructMeta` is exported from the module ([src/salix.c:93](https://github.com/JPHutchins/salix/blob/74d2647/src/salix.c#L93)), so subclassing it and overriding `__new__` is inside the public surface. A metaclass `__new__` that returns a cached class, a proxy, or `NotImplemented` on some path is an ordinary mistake, not an attack.

</details>

<details>

One check between the call and the cast:

```c
	PY_MOVABLE(created, PyType_Type.tp_new(metatype, type_args, NULL));

	if (created != NULL && !PyObject_TypeCheck(created, &StructMeta_Type)) {
		PyErr_Format(
			PyExc_TypeError,
			"StructMeta.__new__ produced %.200s, not a struct class",
			Py_TYPE(created)->tp_name
		);

		return NULL;
	}
```

`PyObject_TypeCheck(created, &StructMeta_Type)` is the right predicate rather than `PyType_Check`: `install_fields` needs `StructType` storage, which only a `StructMeta` instance has.

This is closely related to #15 and #28 — all three are the same missing question, "is this actually a struct?", asked in three different places. Worth fixing as one pass so the predicate is written once.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Closes #15. Closes #28. Closes #29. Three issues, one missing question.

`struct_type_of` reinterprets `Py_TYPE(self)` as a `StructType *`, and only an instance of `StructMeta` has that storage — every other type stops at `PyHeapTypeObject`, so the read runs off the end of the allocation. `is_struct` and `is_struct_class` in `types.h` are that question written once; `set_field` already asked it inline and now shares the name.

### #28 — a mixin subclass segfaults, and the crashing call is an ordinary struct's `__eq__`

```python
Mixin = salix.Struct.__mro__[1]
F = type("F", (Mixin, list), {"__slots__": ("zzzzz",)})
P(1) == F()                          # segfault
```

`_StructMixin` carries `Py_TPFLAGS_BASETYPE` and `Struct.__mro__` hands it out, so a subclass of it whose metaclass is plain `type` reaches all four slots and both getsets while being no such thing. The co-base is what makes it constructible — a bare subclass inherits the mixin's NULL `tp_new` — which is why an earlier look concluded the mixin was closed off.

Where there is no struct the mixin has nothing to contribute, so the object gets what it would have had without it: `object`'s repr and hash, generic setattr, `NotImplemented` from the comparison. Raising out of `repr` or `hash` would turn an object that is merely useless into one a debugger cannot print. The two getsets are the exception — they report struct metadata and nothing else — so they raise `TypeError`.

### #29 — `create_class` cast `type.__new__`'s result unchecked

`type_new` hands off to the winning metatype's `tp_new` when that differs from the one it was given, so a `StructMeta` subclass overriding `__new__` decides what comes back — and `install_fields` writes six fields at `StructType` offsets into it. Returning a `str` or a `bytearray(8192)` segfaults. Checked now, and the error names both the metaclass and what it produced.

### #15 — a bare `metaclass=StructMeta` built something that was not a struct

It reported its fields correctly and had identity equality, `object`'s repr, and a `frozen` that accepted assignment — because `apply_options` only acts on transitions and with no struct base there are none. `bind_hash` bound the mixin's `__hash__` into a class with no mixin in its MRO, so `hash(C(1))` raised about a `'_StructMixin'` object.

Refused now: everything a struct does comes from the mixin, so a class without it in the bases is not one.

<details>
<summary><b>One decision this takes, and one it leaves open</b></summary>

That is **option 1** of the three in #15. **Option 3** — dropping `StructMeta` from the module's exports, keeping it only in the stub where `Struct`'s `metaclass=` needs it — is a separate decision and still open. It has to be made before the first upload either way, and it also settles what #24 and #20 should do about the namespace.

</details>

Sixteen tests in `tests/test_struct_identity.py`, each pinning a guarded outcome rather than a crash: a segfault takes pytest with it.

Rebased onto `main` after #43 and #45 landed, which rewrote `apply_options`/`bind_hash` and `Struct_set_field` underneath this. All three reproductions re-verified after the rebase, along with the merged behaviour from both: `B.__hash__ is None` for an inherited body `__eq__`, and `set_field` still writing. `camas check` green at 25/25.

> [!WARNING]
> **LLM Disclosure**
>

### DIFF
--- a/salix/__init__.pyi
+++ b/salix/__init__.pyi
@@ -2,12 +2,8 @@ from typing import Any, Final
 
 from typing_extensions import dataclass_transform
 
-class _StructMeta(type):
-    __struct_fields__: Final[tuple[str, ...]]
-    __struct_defaults__: Final[tuple[Any, ...]]
-
 @dataclass_transform(frozen_default=True)
-class Struct(metaclass=_StructMeta):
+class Struct:
     __struct_fields__: Final[tuple[str, ...]]
     __struct_defaults__: Final[tuple[Any, ...]]
     def __init_subclass__(
--- a/salix/__init__.pyi
+++ b/salix/__init__.pyi
@@ -2,12 +2,12 @@ from typing import Any, Final
 
 from typing_extensions import dataclass_transform
 
-class StructMeta(type):
+class _StructMeta(type):
     __struct_fields__: Final[tuple[str, ...]]
     __struct_defaults__: Final[tuple[Any, ...]]
 
 @dataclass_transform(frozen_default=True)
-class Struct(metaclass=StructMeta):
+class Struct(metaclass=_StructMeta):
     __struct_fields__: Final[tuple[str, ...]]
     __struct_defaults__: Final[tuple[Any, ...]]
     def __init_subclass__(
--- a/src/compare.c
+++ b/src/compare.c
@@ -1,7 +1,6 @@
 #include <Python.h>
 
 #include "compare.h"
-#include "mixin.h"
 #include "owned.h"
 #include "types.h"
 
@@ -24,7 +23,7 @@ static enum comparison values_equal(
 );
 
 PyObject * Struct_rich_compare(PyObject * const self, PyObject * const other, int const op) {
-	if (!PyObject_TypeCheck(other, &StructMixin_Type)) {
+	if (!is_struct(self) || !is_struct(other)) {
 		Py_RETURN_NOTIMPLEMENTED;
 	}
 
--- a/src/construct.c
+++ b/src/construct.c
@@ -1,7 +1,6 @@
 #include <Python.h>
 
 #include "construct.h"
-#include "meta.h"
 #include "owned.h"
 #include "result.h"
 #include "types.h"
@@ -219,7 +218,7 @@ PyObject * Struct_set_field(PyObject * const module, PyObject * const arguments)
 		return NULL;
 	}
 
-	if (!PyObject_TypeCheck((PyObject *) Py_TYPE(self), &StructMeta_Type)) {
+	if (!is_struct(self)) {
 		PyErr_Format(
 			PyExc_TypeError,
 			"set_field() expects a struct, not %.200s",
--- a/src/hash.c
+++ b/src/hash.c
@@ -24,7 +24,13 @@ Py_hash_t Struct_hash(PyObject * const self) {
 	 * NotImplemented and the co-base's reflected __eq__ says yes -- so an
 	 * identity hash made it equal to a value it could not be looked up beside.
 	 * Refusing to hash is what puts equality and hashing back in one place, and
-	 * being unhashable is an ordinary thing for an object to be. #66. */
+	 * being unhashable is an ordinary thing for an object to be. #66.
+	 *
+	 * Unhashable to the operator and not to the ABC: collections.abc.Hashable
+	 * asks whether tp_hash is non-NULL, and the mixin's is, so an impostor is an
+	 * instance of it and raises when hashed. Setting tp_hash to NULL is not
+	 * available -- it is the same slot every struct hashes through. CPython's
+	 * own writable memoryview diverges the same way. */
 	if (!is_struct(self)) {
 		PyErr_Format(PyExc_TypeError, "unhashable type: '%.200s'", Py_TYPE(self)->tp_name);
 
--- a/src/hash.c
+++ b/src/hash.c
@@ -18,6 +18,10 @@ enum : Py_hash_t {
  * a Python frame, so the interpreter's own depth limit never sees the descent.
  */
 Py_hash_t Struct_hash(PyObject * const self) {
+	if (!is_struct(self)) {
+		return PyBaseObject_Type.tp_hash(self);
+	}
+
 	StructType const * const type = struct_type_of(self);
 	PY_OWNED(values, PyTuple_New(type->struct_field_count));
 
--- a/src/hash.c
+++ b/src/hash.c
@@ -18,8 +18,17 @@ enum : Py_hash_t {
  * a Python frame, so the interpreter's own depth limit never sees the descent.
  */
 Py_hash_t Struct_hash(PyObject * const self) {
+	/* Unhashable rather than hashed by pointer, which is the one fallback in
+	 * this file that could not be a local decision. A mixin subclass over a
+	 * value type is compared by that type -- rich_compare answers
+	 * NotImplemented and the co-base's reflected __eq__ says yes -- so an
+	 * identity hash made it equal to a value it could not be looked up beside.
+	 * Refusing to hash is what puts equality and hashing back in one place, and
+	 * being unhashable is an ordinary thing for an object to be. #66. */
 	if (!is_struct(self)) {
-		return PyBaseObject_Type.tp_hash(self);
+		PyErr_Format(PyExc_TypeError, "unhashable type: '%.200s'", Py_TYPE(self)->tp_name);
+
+		return HASH_ERR;
 	}
 
 	StructType const * const type = struct_type_of(self);
--- a/src/meta.c
+++ b/src/meta.c
@@ -62,6 +62,7 @@ static StructType * create_class(
 	PyObject * bases,
 	PyObject * namespace
 );
+static PyTypeObject * winning_metatype(PyTypeObject * requested, PyObject * bases);
 static enum result install_fields(
 	StructType * struct_class,
 	StructType const * base,
@@ -214,11 +215,11 @@ PyObject * StructMeta_new(
 		NULL
 	);
 
-	/* Already installed means type.__new__ delegated to a more derived metatype,
-	 * which re-entered here and built the class in full -- or a metaclass __new__
-	 * handed back a struct class it did not just make. Installing over either one
-	 * overwrites a field table without releasing it, and the plan that produced
-	 * it came from this same name, bases and namespace anyway. */
+	/* create_class builds as the winning metatype, so the ordinary handoff no
+	 * longer comes back here already installed. What still can is a metaclass
+	 * __new__ that returned a struct class -- possibly one it did not just make,
+	 * whose slot offsets belong to a layout this plan knows nothing about.
+	 * Installing over that is a field table pointed at the wrong memory. */
 	if (
 		struct_class != NULL &&
 		struct_class->struct_field_names == NULL &&
@@ -507,20 +508,27 @@ static StructType * create_class(
 		return NULL;
 	}
 
-	PY_MOVABLE(created, PyType_Type.tp_new(metatype, type_args, NULL));
+	/* type_new hands off to the winning metatype's tp_new when that is not the
+	 * one it was given. Where the winner would only land back in StructMeta_new,
+	 * building as the winner in the first place reaches the same class without
+	 * the round trip -- which re-planned from the transformed namespace and with
+	 * no keywords, so the requested options and the body's defaults were gone by
+	 * the time it returned. */
+	PyTypeObject * const winner = winning_metatype(metatype, bases);
+	PyTypeObject * const builder = winner->tp_new == StructMeta_new ? winner : metatype;
+	PY_MOVABLE(created, PyType_Type.tp_new(builder, type_args, NULL));
 
 	if (created == NULL) {
 		return NULL;
 	}
 
-	/* type_new hands off to the winning metatype's tp_new when that is not the
-	 * one it was given, so a StructMeta subclass overriding __new__ decides what
-	 * comes back here -- and install_fields writes StructType storage into it. */
+	/* A StructMeta subclass overriding __new__ still decides what comes back
+	 * here -- and install_fields writes StructType storage into it. */
 	if (!is_struct_class(created)) {
 		PyErr_Format(
 			PyExc_TypeError,
 			"%.200s.__new__ returned %.200s, which is not a struct class",
-			metatype->tp_name,
+			winner->tp_name,
 			Py_TYPE(created)->tp_name
 		);
 
@@ -528,6 +536,26 @@ static StructType * create_class(
 	}
 
 	return (StructType *) py_move(&created);
+}
+
+/*
+ * type_new's own rule: the most derived of the requested metatype and the
+ * bases' metatypes builds the class. Metatypes that are unrelated to each
+ * other are a conflict rather than a winner, and this returns the requested one
+ * for that case so type_new raises the error it has always raised.
+ */
+static PyTypeObject * winning_metatype(PyTypeObject * const requested, PyObject * const bases) {
+	PyTypeObject * winner = requested;
+
+	for (Py_ssize_t i = 0; i < PyTuple_GET_SIZE(bases); ++i) {
+		PyTypeObject * const candidate = Py_TYPE(PyTuple_GET_ITEM(bases, i));
+
+		if (PyType_IsSubtype(candidate, winner)) {
+			winner = candidate;
+		}
+	}
+
+	return winner;
 }
 
 /* The type exists but is not yet a struct; this is what makes it one. */
--- a/src/meta.c
+++ b/src/meta.c
@@ -91,7 +91,7 @@ static struct member_lookup find_member(
 
 PyTypeObject StructMeta_Type = {
 	PyVarObject_HEAD_INIT(NULL, 0)
-	.tp_name = "salix.StructMeta",
+	.tp_name = "salix._StructMeta",
 	.tp_basicsize = sizeof(StructType),
 	.tp_itemsize = sizeof(PyMemberDef),
 	.tp_flags = (
@@ -152,7 +152,7 @@ PyObject * StructMeta_new(
 	if (
 		!PyArg_ParseTuple(
 			args,
-			"UO!O!:StructMeta.__new__",
+			"UO!O!:_StructMeta.__new__",
 			&name,
 			&PyTuple_Type,
 			&bases,
@@ -172,8 +172,8 @@ PyObject * StructMeta_new(
 	if (base == NULL && !inherits_mixin(bases)) {
 		PyErr_SetString(
 			PyExc_TypeError,
-			"a struct class inherits salix.Struct; StructMeta is the metaclass of "
-			"one, not a way to make one"
+			"a struct class inherits salix.Struct; the metaclass of one is not a "
+			"way to make one"
 		);
 
 		return NULL;
@@ -250,7 +250,7 @@ PyObject * StructMeta_new(
  * bindings, the hash, the setattr that makes frozen mean something. Without it
  * in the MRO the class still builds, still constructs and still reports its
  * fields, and is a struct in every visible way except behaviour -- so a bare
- * `metaclass=StructMeta` is refused rather than answered with that.
+ * `metaclass=type(Struct)` is refused rather than answered with that.
  */
 static bool inherits_mixin(PyObject * const bases) {
 	for (Py_ssize_t i = 0; i < PyTuple_GET_SIZE(bases); ++i) {
--- a/src/meta.c
+++ b/src/meta.c
@@ -26,7 +26,14 @@ static int StructMeta_traverse(PyObject * self, visitproc visit, void * arg);
 static int StructMeta_clear(PyObject * self);
 static void StructMeta_dealloc(PyObject * self);
 
-static bool inherits_mixin(PyObject * bases);
+static PyObject * build_struct_class(
+	PyTypeObject * metatype,
+	StructType const * base,
+	PyObject * name,
+	PyObject * bases,
+	PyObject * original_namespace,
+	PyObject * keywords
+);
 static StructType * find_struct_base(PyObject * bases);
 static bool has_weakref_slot(StructType const * base);
 static struct options base_options(StructType const * base);
@@ -128,17 +135,23 @@ static PyGetSetDef StructMeta_getset[] = {
 };
 
 static PyObject * StructMeta_get_field_names(PyObject * const self, void * const closure) {
-	return struct_tuple_or_empty(((StructType *) self)->struct_field_names);
+	return struct_metadata((StructType *) self, STRUCT_FIELD_NAMES);
 }
 
 static PyObject * StructMeta_get_defaults(PyObject * const self, void * const closure) {
-	return struct_tuple_or_empty(((StructType *) self)->struct_defaults);
+	return struct_metadata((StructType *) self, STRUCT_DEFAULTS);
 }
 
 /*
- * Creating a struct class is four steps: work out the fields, build the
- * namespace type.__new__ wants, make the type, then hand it the field table
- * that makes it a struct.
+ * Everything a struct does comes from _StructMixin, and every struct class
+ * carries it because Struct does. A class with no struct base has no way to
+ * have got it: it would build, construct, report its fields, and be a struct
+ * in every visible way except behaviour.
+ *
+ * The one class that legitimately has no struct base is Struct, and the module
+ * builds it through struct_create_root rather than through here -- so this
+ * refusal has no exception to carve out, and there is no shape of `bases` that
+ * gets a caller past it.
  */
 PyObject * StructMeta_new(
 	PyTypeObject * const metatype,
@@ -165,11 +178,7 @@ PyObject * StructMeta_new(
 
 	StructType const * const base = find_struct_base(bases);
 
-	/* A struct base is a mixin subtype already -- Struct is built on the mixin
-	 * and every struct descends from it -- so the scan below is only for the
-	 * classes that have no struct base at all, Struct's own creation among
-	 * them. */
-	if (base == NULL && !inherits_mixin(bases)) {
+	if (base == NULL) {
 		PyErr_SetString(
 			PyExc_TypeError,
 			"a struct class inherits salix.Struct; the metaclass of one is not a "
@@ -179,6 +188,32 @@ PyObject * StructMeta_new(
 		return NULL;
 	}
 
+	return build_struct_class(metatype, base, name, bases, original_namespace, keywords);
+}
+
+/* Struct itself, built once from module init. The only class with no struct
+ * base, and the only caller that does not come through a metaclass call. */
+PyObject * struct_create_root(
+	PyObject * const name,
+	PyObject * const bases,
+	PyObject * const namespace
+) {
+	return build_struct_class(&StructMeta_Type, NULL, name, bases, namespace, NULL);
+}
+
+/*
+ * Creating a struct class is four steps: work out the fields, build the
+ * namespace type.__new__ wants, make the type, then hand it the field table
+ * that makes it a struct.
+ */
+static PyObject * build_struct_class(
+	PyTypeObject * const metatype,
+	StructType const * const base,
+	PyObject * const name,
+	PyObject * const bases,
+	PyObject * const original_namespace,
+	PyObject * const keywords
+) {
 	struct options const inherited = base_options(base);
 	struct options_request const request =
 		options_read(keywords, inherited, base != NULL && base->struct_field_count > 0);
@@ -243,25 +278,6 @@ PyObject * StructMeta_new(
 	field_plan_clear(&plan);
 
 	return (PyObject *) struct_class;
-}
-
-/*
- * Everything a struct does comes from _StructMixin: the comparison and repr
- * bindings, the hash, the setattr that makes frozen mean something. Without it
- * in the MRO the class still builds, still constructs and still reports its
- * fields, and is a struct in every visible way except behaviour -- so a bare
- * `metaclass=type(Struct)` is refused rather than answered with that.
- */
-static bool inherits_mixin(PyObject * const bases) {
-	for (Py_ssize_t i = 0; i < PyTuple_GET_SIZE(bases); ++i) {
-		PyObject * const base = PyTuple_GET_ITEM(bases, i);
-
-		if (PyType_Check(base) && PyType_IsSubtype((PyTypeObject *) base, &StructMixin_Type)) {
-			return true;
-		}
-	}
-
-	return false;
 }
 
 /* Find the (single) struct base among ``bases``. A fieldless one still carries
@@ -554,6 +570,12 @@ static StructType * create_class(
  * raises the metaclass-conflict error it has always raised. For unrelated
  * metatypes that is the first one this locked onto rather than the requested
  * one, which is why the loop can stay this simple.
+ *
+ * It is the third walk of `bases` in a class creation, after find_struct_base
+ * and _PyType_CalculateMetaclass. Measured rather than assumed: replacing the
+ * body with `return requested` moves class creation from 9.77-9.87 to
+ * 9.77-9.87 us for a 16-field class, which is to say it does not move it. The
+ * walk is one Py_TYPE and one PyType_IsSubtype per base, and a class has one.
  */
 static PyTypeObject * winning_metatype(PyTypeObject * const requested, PyObject * const bases) {
 	PyTypeObject * winner = requested;
--- a/src/meta.c
+++ b/src/meta.c
@@ -572,10 +572,11 @@ static StructType * create_class(
  * one, which is why the loop can stay this simple.
  *
  * It is the third walk of `bases` in a class creation, after find_struct_base
- * and _PyType_CalculateMetaclass. Measured rather than assumed: replacing the
- * body with `return requested` moves class creation from 9.77-9.87 to
- * 9.77-9.87 us for a 16-field class, which is to say it does not move it. The
- * walk is one Py_TYPE and one PyType_IsSubtype per base, and a class has one.
+ * and _PyType_CalculateMetaclass. Measured, and smaller than the measurement:
+ * replacing the body with `return requested` leaves class creation at 9.77-9.87
+ * us for a 16-field class either way, so the walk is bounded by the width of
+ * that band rather than shown to be free. It is one Py_TYPE and one
+ * PyType_IsSubtype per base, and a class has one.
  */
 static PyTypeObject * winning_metatype(PyTypeObject * const requested, PyObject * const bases) {
 	PyTypeObject * winner = requested;
--- a/src/meta.c
+++ b/src/meta.c
@@ -265,7 +265,7 @@ static StructType * find_struct_base(PyObject * const bases) {
 	for (Py_ssize_t i = 0; i < PyTuple_GET_SIZE(bases); ++i) {
 		PyObject * const base = PyTuple_GET_ITEM(bases, i);
 
-		if (PyObject_TypeCheck(base, &StructMeta_Type)) {
+		if (is_struct_class(base)) {
 			return (StructType *) base;
 		}
 	}
@@ -540,9 +540,15 @@ static StructType * create_class(
 
 /*
  * type_new's own rule: the most derived of the requested metatype and the
- * bases' metatypes builds the class. Metatypes that are unrelated to each
- * other are a conflict rather than a winner, and this returns the requested one
- * for that case so type_new raises the error it has always raised.
+ * bases' metatypes builds the class. CPython keeps that rule in
+ * _PyType_CalculateMetaclass, which is not in the public API, so it is written
+ * out here rather than called.
+ *
+ * Only the winner is wanted, and only to decide who builds. Metatypes unrelated
+ * to each other are a conflict rather than a winner, and returning the
+ * requested one for that case hands type_new an argument it will reject with
+ * the metaclass-conflict error it has always raised -- the one message a caller
+ * already knows.
  */
 static PyTypeObject * winning_metatype(PyTypeObject * const requested, PyObject * const bases) {
 	PyTypeObject * winner = requested;
--- a/src/meta.c
+++ b/src/meta.c
@@ -77,6 +77,11 @@ static enum result install_fields(
 	struct options options,
 	bool resolves_body_eq
 );
+static enum result reject_unless_planned(
+	StructType const * struct_class,
+	struct field_plan const * plan,
+	struct options options
+);
 static enum result install_post_init(StructType * struct_class);
 static bool defines_own_init(StructType const * struct_class);
 static PyObject * StructMeta_call(PyObject * self, PyObject * args, PyObject * keywords);
@@ -259,20 +264,23 @@ static PyObject * build_struct_class(
 	 * longer comes back here already installed. What still can is a metaclass
 	 * __new__ that returned a struct class -- possibly one it did not just make,
 	 * whose slot offsets belong to a layout this plan knows nothing about.
-	 * Installing over that is a field table pointed at the wrong memory. */
-	if (
-		struct_class != NULL &&
-		struct_class->struct_field_names == NULL &&
-		install_fields(
+	 * Installing over that is a field table pointed at the wrong memory, so the
+	 * class it did build is held to what this call planned instead. */
+	if (struct_class != NULL) {
+		enum result const settled = (
+			struct_class->struct_field_names == NULL ? install_fields(
 				struct_class,
 				base,
 				&plan,
 				request.options,
 				body_defines_eq || inherits_body_eq
-			) !=
-			RESULT_OK
-	) {
-		Py_CLEAR(struct_class);
+			) :
+			reject_unless_planned(struct_class, &plan, request.options)
+		);
+
+		if (settled != RESULT_OK) {
+			Py_CLEAR(struct_class);
+		}
 	}
 
 	field_plan_clear(&plan);
@@ -590,6 +598,57 @@ static PyTypeObject * winning_metatype(PyTypeObject * const requested, PyObject 
 	}
 
 	return winner;
+}
+
+/*
+ * The class came back already a struct, which means something other than this
+ * call built it. A metaclass __new__ written in Python is one such thing: it is
+ * re-entered through type_new with no keywords and a namespace the transform
+ * has already taken the body defaults out of, so the class it installs is
+ * planned from less than this call was handed. The options and the defaults are
+ * where that shows.
+ *
+ * Refused rather than returned, and rather than installed over: the field table
+ * this call planned describes a layout that class may not have. #55 is where
+ * the hand-off itself is argued -- until it forwards what it was given, a
+ * caller gets an error instead of a class that quietly is not the one asked
+ * for.
+ */
+static enum result reject_unless_planned(
+	StructType const * const struct_class,
+	struct field_plan const * const plan,
+	struct options const options
+) {
+	PY_OWNED(planned, PyList_AsTuple(plan->all_names));
+
+	if (planned == NULL) {
+		return RESULT_ERROR;
+	}
+
+	int const same_fields =
+		PyObject_RichCompareBool(struct_class->struct_field_names, planned, Py_EQ);
+
+	if (same_fields < 0) {
+		return RESULT_ERROR;
+	}
+
+	if (
+		same_fields == 1 &&
+		struct_class->struct_default_count == PyTuple_GET_SIZE(plan->defaults) &&
+		options_agree(struct_class->struct_options, options)
+	) {
+		return RESULT_OK;
+	}
+
+	PyErr_Format(
+		PyExc_TypeError,
+		"%.200s.__new__ returned a struct class this call did not plan: its "
+		"metaclass is re-entered without the keywords or the class body's "
+		"defaults, so what it built is not what was asked for",
+		Py_TYPE(struct_class)->tp_name
+	);
+
+	return RESULT_ERROR;
 }
 
 /* The type exists but is not yet a struct; this is what makes it one. */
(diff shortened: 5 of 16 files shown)


JPHutchins:
Two taken — they turned out to be one cause and a real leak — two do not reproduce, and one is documented. `camas check` green at 25/25.

**4, and 2 with it — `install_fields` running twice.** Confirmed, and measured on the `__post_init__` reference:

```
class Derived(StructMeta): pass
class Base(Struct, metaclass=Derived):
    x: int
    __post_init__ = post_init

StructMeta("Built", (Base,), {...})     post_init refcount +3
```

One of those three is the namespace dict. The other two are `install_post_init` running twice, and only one is ever given back — along with the defaults tuple, the field-names tuple and the malloc'd slot-offset array, one leak per class. **+2 now**, with a regression test in `tests/test_refcounts.py`.

The same guard answers finding 2: a metaclass `__new__` returning a struct class it did not just make is no longer installed over either.

<details>
<summary><b>1 and 3 — do not reproduce, and CPython is why</b></summary>

Both need a `StructMeta`-metaclass class that never ran `install_fields`, and both routes to one are refused by the interpreter:

```
type.__new__(salix.StructMeta, "X", (object,), {})
TypeError: type.__new__(salix.StructMeta) is not safe, use salix.StructMeta.__new__()

class Bypass(StructMeta):
    def __new__(mcls, *a): return type.__new__(mcls, *a)
class X(Struct, metaclass=Bypass): ...
TypeError: type.__new__(Bypass) is not safe, use salix.StructMeta.__new__()
```

`tp_new_wrapper` refuses when the static base's `tp_new` is not the one being called, and `StructMeta` is that static base. So `is_struct` being a metaclass-identity check cannot admit a fieldless class, and the class-side getsets cannot be reached on one.

Worth noting the same guard is what blocks honouring a body `__new__` in #12 — it cuts both ways.

If you have a route past it I have not found, I will take the marker-bit fix; it is a good design and I would rather not add state for a state that cannot occur.

</details>

<details>
<summary><b>5 — fair, and documented rather than fixed</b></summary>

Right that "what it would have had without the mixin" holds only when the co-base is `object`: an impostor over `list` becomes hashable by pointer where a plain `list` subclass is unhashable, and over `tuple` it hashes by identity instead of by value.

Deferring to the co-base means walking the MRO for a `__repr__`/`__hash__` to borrow, on a path only reachable by subclassing a private class with a leading underscore. The rationale in `mixin.c` says `object` semantics explicitly now rather than implying the co-base is consulted.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
**Finding 3 is right twice, and the second time is on me.**

The comment does claim a non-struct mixin instance "gets what it would have had without it", and that is false for a co-base that is not `object`. But my previous round's reply said *"the rationale in `mixin.c` says `object` semantics explicitly now rather than implying the co-base is consulted"* — and it did not. I described an edit I never applied. Corrected in `6ceee5b`, and I have checked the file rather than my memory of it this time.

**2 — the raw mixin as a co-base.** Reproduced, and it does not do what the finding expects. `salix.StructMeta("X", (Struct.__mro__[1],), {...})` passes the gate and builds a **working** struct, not a broken one:

```
fields                      ('a',)
X(1) == X(1)                True        (structural)
repr(X(1))                  X(a=1)
frozen                      yes
hash(X(1)) == hash((1,))    True
X(1) == P(1)                True        (compares to a real struct)
isinstance(X(1), Struct)    False
issubclass(X, Struct)       False
X.__mro__                   ['X', '_StructMixin', 'object']
```

So #15's refusal is not defeated: its purpose was that a bare `metaclass=StructMeta` produced a class that *reported* its fields while behaving like `object` — identity equality, `object`'s repr, a frozen that accepted assignment, and a `hash` that raised. None of that is true here, because the mixin supplies exactly the dunders whose absence was the bug.

What is true is the last three lines: it is a functional struct that is not `Struct`-derived, so `isinstance` says no while everything else says yes. That is a curiosity of naming a private class deliberately, not the defect #15 describes. Happy to tighten the gate to `PyType_IsSubtype(base, &StructMeta_Type)`-style if you would rather the only way in is through `Struct`.

<details>
<summary><b>1 — a metaclass <code>__new__</code> returning a different existing struct class</b></summary>

Correct that `is_struct_class(created)` does not check the class matches the request, and that the new `struct_field_names == NULL` guard then skips installation over it.

I think skipping is right and refusing would be wrong. A `__new__` returning something other than what it was asked to build is an ordinary Python facility — interning, caching, registry lookups — and the object it returns is already a complete struct with a field table that its own creation installed. Overwriting that table was the leak this round fixed; refusing it would forbid a pattern the language allows, for a case where the returned object is well-formed.

What the caller gets is the class the metaclass chose, which is what a metaclass `__new__` returning something else means everywhere else in Python.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Finding 1 was right and the fix is upstream of where you pointed. Two taken, one declined onto #15. `camas check` green at 25/25 on `2172f1c`.

**1 — the guard dropped keyword options and field defaults.** Reproduced both halves before touching anything:

```
order over derived metatype  : TypeError: '<' not supported between instances of 'B' and 'B'
order over plain StructMeta  : True
default over derived metatype: TypeError: D() missing required argument 'y'
default over plain StructMeta: 42
```

I did not forward the keywords, because that fixes the options and leaves the defaults: the re-entered call reads the namespace *after* `drop_class_variables` has taken the field names out of it, so `y: int = 42` is gone before `field_plan_build` looks for it, keywords or not.

The round trip is the thing to remove. `type_new` hands off to the winning metatype when that is not the one it was given — and where the winner's `tp_new` *is* `StructMeta_new`, that handoff only re-derives the same class from worse inputs. `create_class` now resolves the winner the way `type_new` does and builds as it:

```c
PyTypeObject * const winner = winning_metatype(metatype, bases);
PyTypeObject * const builder = winner->tp_new == StructMeta_new ? winner : metatype;
```

One build, one install, with the caller's keywords and the body's namespace:

```
order over derived metatype  : True
default over derived metatype: 42
metatype of B                : Derived
post_init refcount delta     : 2 (want 2)
alias returned Existing      : True / Existing fields still: ('a',)
```

Last line matters: **the guard is still load-bearing.** I checked whether the aliasing case you called superseded is actually reachable, and it is —

```python
class Wrong(salix.StructMeta):
    def __new__(mcls, *args, **kw):
        return super().__new__(mcls, *args, **kw) if first_call else Existing
```

`Existing` reaches `create_class` and passes `is_struct_class`. Installing this plan's `struct_slot_offsets` over it points a field table at a layout its instances do not have, so the guard stays. My round-2 rebuttal was narrower than I made it sound; thank you for pushing on it.

What the guard still costs is a Python `__new__` on the metaclass losing options and defaults — filed as **#55** rather than fixed here, because telling "the class my delegate just built" from "a class my delegate found lying around" needs more than a NULL check and the ordinary routes are unaffected:

```python
class Sub(Base):                  # class statement, Forwarding metaclass
    y: int = 42
Sub(1).y                          # 42
Ordered(1, 2) < Ordered(1, 3)     # True
```

It needs an explicit `salix.StructMeta(...)` call over a base with such a metatype — which #15's option 3 removes outright.

**4 — the refusal named the wrong metatype.** Taken, `winner->tp_name`. Test now matches `Wrong\.__new__ returned .*is not a struct class` instead of just the tail.

**2 — inline `type(...)` recursion.** Not salix's to fix, and confirmed by removing salix from it:

```python
class PlainMeta(type):
    def __new__(mcls, *a, **kw):
        return super().__new__(mcls, *a, **kw) if first_call else type(*a)
# -> RecursionError
```

`type(name, bases, ns)` re-derives the same metaclass from the bases and re-enters, every time, with or without a struct in sight. Your recommendation's second branch is the right one: I narrowed the promise. The test is now `test_one_that_returns_a_non_type_is_refused` and its docstring says the refusal covers what a `__new__` returns, not every way one can fail to return.

**3 — the raw `_StructMixin` co-base.** Declined here, recorded on [#15](https://github.com/JPHutchins/salix/issues/15#issuecomment-5150326712) with the measurement. Short version: every stateless tightening refuses `Struct`'s own creation, because `create_struct_base` builds it with exactly that shape —

| tightening | effect on `Struct`'s own creation |
| --- | --- |
| require `find_struct_base(bases) != NULL` | refused — no struct base exists yet |
| require the base's metatype be a `StructMeta` subtype | refused — `type(_StructMixin)` is `type` |
| require a proper subclass of the mixin | refused — the mixin is the only base |

A "have we built `Struct` yet" flag works once and then refuses the rebuild after `del sys.modules["salix"]`. You are right that the error message overclaims; the honest fix is option 3 on #15, which is with the owner.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues and answer each finding on its PR. This is round 4 on #48.

JPHutchins:
The major finding does not reproduce. Three taken, four declined onto issues that already hold them. `camas check` green at 25/25 on `9447015`.

## 1 — `type.__new__(StructMeta, ...)` on <3.14

Built the case on every supported version:

```
3.10  refused: type.__new__(salix.StructMeta) is not safe, use salix.StructMeta.__new__()
3.11  refused: type.__new__(salix.StructMeta) is not safe, use salix.StructMeta.__new__()
3.12  refused: type.__new__(salix.StructMeta) is not safe, use salix.StructMeta.__new__()
3.13  refused: type.__new__(salix.StructMeta) is not safe, use salix.StructMeta.__new__()
3.14  refused: type.__new__(salix.StructMeta) is not safe, use salix.StructMeta.__new__()
```

The grep is where it went wrong: the binary carries the **template**, not the rendered message.

```
$ strings libpython3.12.so | grep "is not safe"
%s.__new__(%s) is not safe, use %s.__new__()
```

Present in 3.10, 3.12 and 3.13, in both the executable and `libpython`. `tp_new_wrapper` walks to the most derived non-heap base of the requested type — `StructMeta`, whose `tp_new` is `StructMeta_new`, not `type`'s — and refuses. That is not a 3.14 addition.

I also checked the routes the finding did **not** name, since a guard is only as good as its narrowest hole:

| route | 3.10 | 3.13 |
| --- | --- | --- |
| `type.__new__(StructMeta, ...)` | refused | refused |
| `type.__new__(SubNoNew, ...)` | refused | refused |
| `type.__new__(SubWithNew, ...)` | refused | refused |
| `SubWithNew("S", (P,), {})` — the substitution route | refused | refused |
| `types.new_class(metaclass=StructMeta)` | **BUILT, fields=('a',)** | **BUILT, fields=('a',)** |

The last row is the control: the supported route still installs, so this is not a guard that refuses everything.

Nothing in the repository said the invariant rested on a CPython guard, which is why it read as unproven — so `TestTheUninstalledClassCannotBeBuilt` now pins all five rows. A CPython that ever drops `tp_new_wrapper`'s check fails the suite instead of segfaulting.

**2 — the metaclass-side getsets.** Declined, and it follows from the above. The mixin-side getters were hardened because a *non-struct* can reach them: `type("X", (MIXIN, list), {})` is a plain-`type` instance with mixin slots. `StructMeta_get_field_names` can only ever be reached on a `StructMeta` instance, and there is no such instance without an installed table — that is exactly what the table above establishes. A check for a state that cannot be constructed is dead code, and I would rather the invariant be pinned by tests than restated as a runtime branch.

**3 — the Python-`__new__` residue.** Agreed, unchanged, and it is **#55**, filed after last round with the reproduction and the reason a NULL check cannot tell "the class my delegate built" from "a class my delegate found".

**4 — the raw `_StructMixin` co-base.** Still declined, recorded on [#15](https://github.com/JPHutchins/salix/issues/15#issuecomment-5150326712) with the three tightenings that each refuse `Struct`'s own creation.

**5 — the duplicated check.** Taken; `find_struct_base` uses `is_struct_class`.

**6 — seven guards, five fallbacks.** Declined on scope, but the premise deserves an answer: the fallbacks differ because the right answer differs per slot. `rich_compare` must return `NotImplemented` so Python tries the reflected operation; `repr`/`hash` falling back to `object`'s is what a non-struct should get; `setattr` falling back to `PyObject_GenericSetAttr` is what a *mutable* non-struct should get; the getsets and `set_field` have no sensible non-struct answer, so they raise. A single raising `tp_new` would be a different design and might well be a better one — it is not a smaller change, and I would not make it in the PR that fixes the segfault.

**7 — `winning_metatype` re-implements CPython's rule.** True, and unavoidable: `_PyType_CalculateMetaclass` is not public API. The comment says so now. The conflict deferral is deliberate rather than a bet, and `test_two_unrelated_metatypes_still_raise_the_conflict` proves it reaches `metaclass conflict` rather than being swallowed.

**8 — the refcount comment.** Taken. `before` already counts the namespace dict; the two are the class's own binding and `struct_post_init`.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues and answer each finding on its PR. This is round 5 on #48.

JPHutchins:
Finding 4 is right that my docstring was wrong, and wrong about why. Two taken, four unchanged. `camas check` green at 25/25 on `b3e87f2`.

**4 — the named crash vector.** I built both claims:

```
x == x    -> __eq__ invoked 2 times    (== compiles to PyObject_RichCompare)
p in [p]  -> __eq__ invoked 0 times    (that is where the short-circuit lives)
```

So `x == x` **does** reach `Struct_rich_compare` — the identity short-circuit is in `PyObject_RichCompareBool`, which `==` does not go through. And `names_equal` does pass the two field tables straight to `PyObject_RichCompareBool`, so for an uninstalled class they are literally the `NULL, NULL` arguments.

But following it one step further kills my claim anyway: `PyObject_RichCompareBool(NULL, NULL, Py_EQ)` hits its own `v == w` short-circuit and answers *equal* before dereferencing either. So the specific crash I named is not one.

Neither of us should have been asserting a crash we cannot run — the gate cannot be turned off to find out. The docstring now says the tests pin the invariant every guarded slot reads and stops there, and the module docstring no longer sweeps this class into "everything asserted here used to be a segfault", since it pins a CPython guard salix relies on and never crashed without. Thank you for pushing on it; the correction is real even though the reasoning behind the finding was not.

**6 — `winning_metatype`'s comment.** Correct. For unrelated metatypes it returns the first one it locked onto, not the requested one. `type_new` recomputes and raises the conflict either way — which is *why* the loop can stay this simple — so the comment now describes the loop rather than an intention it does not implement.

**5 — the hashability grant.** Agreed on the substance, and it is already written down: `mixin.c` says "an impostor over `list` is given object's repr and hashes by pointer, where a plain list subclass is unhashable", and `test_it_hashes_by_identity` pins it. Nothing to change.

**1 — #55**, **2 — [#15](https://github.com/JPHutchins/salix/issues/15#issuecomment-5150326712)**, **3 — declined on scope last round** with the per-slot reasoning. All unchanged.

**Note on this PR's reviews:** five review jobs on this head were cancelled mid-flight before one landed. Timing data is on the code-review repo's #107, including that a successful retry hides the failed attempt in the jobs API, so the visible failure rate understates the real one.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues and answer each finding on its PR. This is round 6 on #48.

JPHutchins:
One taken, one declined on style grounds, three unchanged. `camas check` green at 25/25 on the new head.

**5 — `bases` scanned twice.** Taken. A struct base settles the mixin question, since `Struct` is built on the mixin and every struct descends from `Struct` — so the mixin scan is only needed for classes with no struct base at all, `Struct`'s own creation among them. Reordered, and both refusals still refuse:

```
StructMeta('B', (), {})        refused
StructMeta('B', (list,), {})   refused
subclass of a struct           ('x', 'y')
```

**4 — the two getsets.** I tried this and backed it out, because every consolidation available is worse than the duplication under this codebase's rules. The blocker is that the member access has to happen *after* `is_struct`, so it cannot be passed as a value:

- an `offsetof` parameter needs `(char *)` pointer arithmetic and a cast back — and "pretty much never cast" is a house rule
- a macro would work and the house rule prefers `static inline`
- an accessor function per member is more code than the two lines it removes

The duplication is one ternary each, and the thing you are protecting against — a third getter copying it — has not happened. I would rather leave two honest copies than introduce pointer arithmetic into a guard whose whole job is preventing an out-of-bounds read. Happy to be overruled if you see a spelling I have missed.

**1 — #55**, **2 — [#15](https://github.com/JPHutchins/salix/issues/15#issuecomment-5150326712)**, **3 — declined on scope** with the per-slot reasoning in round 5. All unchanged.

**On this PR's reviews:** this verdict landed on the fifth attempt; the four before it were cancelled mid-flight. Timing and the corrected failure rate are on code-review#107.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues and answer each finding on its PR. This is round 7 on #48.

JPHutchins:
Two corrections accepted, no code change this round. Head is unchanged at `1005fa7`, green at 25/25.

**5 — `winning_metatype` still walks the bases.** Fair, and it lands on my commit message rather than the code. `1005fa7` is titled "stop scanning twice" and what it actually did was drop one of *three* walks — `inherits_mixin` in the common case, leaving `find_struct_base` in `StructMeta_new` and `winning_metatype` in `create_class`. Calling that "stop scanning twice" overstated it.

Folding the last two means threading the winner from `create_class` back up, or the base down; both are a parameter for one walk of a one- or two-element tuple. I would rather leave it and have the record say two walks than add the plumbing — but the commit message was wrong and I would not want that to be what a future reader takes from it.

**4 — the duplication is four, not two.** Also correct. I answered last round with "two honest copies", counting only the mixin pair; the metaclass has its own `StructMeta_get_field_names`/`StructMeta_get_defaults`. Those two are simpler — always on a `StructType`, so no `is_struct` guard — which is why they did not come to mind, and is not a reason to have miscounted.

The decline stands on the same ground: the consolidation needs `offsetof` plus `(char *)` arithmetic and a cast back, inside a guard whose job is preventing an out-of-bounds read. Four copies of a two-line accessor is a real cost; pointer arithmetic there is a worse one. If #15's option 3 drops `StructMeta` from the exports, two of the four go with it.

**1 — #55**, **2 — [#15](https://github.com/JPHutchins/salix/issues/15#issuecomment-5150326712)**, **3 — declined on scope in round 5.** Unchanged.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues and answer each finding on its PR. This is round 8 on #48.

JPHutchins:
**#15 is ruled — option 3 — and this branch now implements it.** `dc04441`, green at 25/25.

`salix.StructMeta` is gone from the module. The stub calls it `_StructMeta`, and `tp_name` says `salix._StructMeta` so a repr cannot advertise a name the module does not have.

```python
>>> [n for n in dir(salix) if not n.startswith("__")]
['Struct', 'set_field']
>>> type(salix.Struct)
<class 'salix._StructMeta'>
```

**One assumption, and it is worth one word from you if it is wrong.** The issue said "option 1 looks right, possibly with 3", and the ruling named 3. I read that as *3 on top of what is already here*, so the refusal this branch added in round 1 stays: `type(Struct)("X", (), {})` is still a `TypeError` rather than a broken class. Option 3 removes the invitation; it does not remove the route, because `type(Struct)` is one lookup away. If you meant 3 *instead of* 1, say so and the refusal comes back out.

<details>
<summary><b>What the ruling settles, and what it only narrows</b></summary>

| finding | before | now |
| --- | --- | --- |
| 1 — a metaclass `__new__` drops options and defaults (#55) | `salix.StructMeta(...)` | needs `type(Struct)(...)` and a subclass of it — narrowed, not closed |
| 2 — the raw `_StructMixin` as a base | `salix.StructMeta` + `Struct.__mro__[1]` | both now have to be dug out of `Struct` |
| 4 — the metadata getsets | four | still four, and I was wrong to expect otherwise |

**Finding 4, corrected again.** I noted when the ruling landed that dropping the export would halve the four getsets. It does not, and the reason is one I could have checked before writing it: `Point.__struct_fields__` resolves through `type(Point)`, so the metaclass pair answers for the class and the mixin pair for the instance. Neither pair is reachable from the other and neither is redundant. Four is the honest count and the export was never what made it four.

**Finding 2, measured rather than argued.** With the mixin as a base and no struct base, the class that comes out is not broken — it is a working struct that simply is not `Struct`-derived:

```python
>>> X = type(Struct)("X", (Struct.__mro__[1],), {"__annotations__": {"a": int}})
>>> X(1) == X(1), repr(X(1)), isinstance(X(1), Struct)
(True, 'X(a=1)', False)
>>> X(1).a = 2
TypeError: X object does not support attribute assignment
```

That is the whole of the hole: fields, repr, equality, hash and frozen all work, `isinstance(x, Struct)` is `False`, and reaching it takes `Struct.__mro__[1]` by hand. #15's symptom was a class that reported its fields and behaved like `object`; this is not that, so I am leaving it. It stays a nit rather than a fix.

</details>

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She ruled option 3 on #15 and asked that the open PRs keep iterating against the automated reviewer; this round applies that ruling to this branch.

JPHutchins:
**Five taken, two declined with measurements, two filed.** `51cfd80`, green at 25/25.

**7 — the sharpest one, and mine.** The stub kept `class _StructMeta(type)` at top level after the runtime dropped it, so `from salix import _StructMeta` type-checked and raised. The stub does not need it at all — `__struct_fields__` and `__struct_defaults__` are already declared on `Struct`, which is where a checker resolves them for the class as much as the instance. Removing it makes the metaclass nameable from nowhere.

It also *gained* a check I did not expect. A class keyword goes to the metaclass when the stub names one, and mypy stops comparing it against `__init_subclass__`:

```python
class Typo(Struct, frozn=False): ...
```

That was caught by pyright and ty only, and the test said so in its docstring. All three catch it now, and the docstring and the third suppression say why.

**3 — the gate now means what its message says.** You are right that `PyType_IsSubtype(base, &StructMixin_Type)` is true for the mixin itself. The exception existed for exactly one class — `Struct`, which has no struct base — so `Struct` is built through `struct_create_root` and the gate is `find_struct_base(bases) == NULL` and nothing else. `inherits_mixin` is deleted.

```console
raw mixin base   refused: a struct class inherits salix.Struct; the metaclass of one is not a way to make one
impostor base    refused: ...
no bases         refused: ...
list base        refused: ...
struct base      built  ('x', 'y')
```

**4, 5, 9 — taken as recommended.** The getsets raise `AttributeError`, so `hasattr` and `getattr(..., default)` behave (all three asserted). `struct_metadata` in types.h answers which tuple and `metadata_of` in mixin.c decides what a non-struct gets, so the read is one place and the fallback policy is another — four entry points remain because `PyGetSetDef` needs four. And the test now asserts `not hasattr(salix, META.__name__)`, which is the relationship rather than the string.

<details>
<summary><b>8 — refuted: <code>PyObject_TypeCheck</code> already short-circuits, and the guard is unmeasurable</b></summary>

The finding is that `is_struct` puts a metaclass-MRO walk on every dunder. It does not, for any ordinary struct. From `object.h` in the interpreter this was built against:

```c
static inline int PyObject_TypeCheck(PyObject *ob, PyTypeObject *type) {
    return Py_IS_TYPE(ob, type) || PyType_IsSubtype(Py_TYPE(ob), type);
}
```

`is_struct(self)` is `PyObject_TypeCheck(Py_TYPE(self), &StructMeta_Type)`, and for a class whose metaclass *is* `StructMeta` — which is every struct not built through a metaclass subclass — `Py_IS_TYPE` answers on one pointer compare and `PyType_IsSubtype` is never called. The recommended identity short-circuit is already there, one layer down.

Measured too, by mutating `is_struct` to `return true` and swapping the two builds back and forth:

| | `is_struct` → `true` | as written |
| --- | --- | --- |
| hash | 45.9, 46.1, 46.2 ns | 45.3, 45.4, 45.4 ns |
| eq | 22.1, 22.2, 22.3 ns | 22.6, 22.6, 22.7 ns |
| repr | 352.6, 354.6, 366.4 ns | 349.3, 351.8, 366.3 ns |

Removing the guard entirely makes `hash` slower and `eq` faster by a few tenths of a nanosecond. That is layout noise in both directions, which is the honest reading of a change that costs one pointer compare.

</details>

<details>
<summary><b>6 — declined again, and the measurement is now in the comment</b></summary>

The recommendation's first half is what the loop already does — it cannot know no base's metaclass is a proper subtype without asking each one. So this takes the second half. Replacing `winning_metatype`'s body with `return requested`, 16-field class, 2000 classes, min of 20, builds swapped:

```
no-walk     9.766 us/class      as written  9.765 us/class
no-walk     9.863 us/class      as written  9.866 us/class
no-walk     9.777 us/class      as written  9.810 us/class
```

The walk is one `Py_TYPE` and one `PyType_IsSubtype` per base, and a class has one base. The comment says this now, so the next round does not have to re-derive it.

</details>

**1 — #55**, unchanged and still out of scope; your recommendation ("verify the installed field names against the outer plan before skipping the install") is now recorded there as the concrete shape of a fix.

**2 — filed as #66**, and reproduced first:

```
>>> imp == plain, plain == imp
(True, True)
>>> {plain: "from the tuple"}.get(imp, "MISSED")
'MISSED'
```

Equality comes from the co-base's reflected `__eq__` and the hash from `object`, so the two disagree. It is #48's residue rather than its regression — before this branch that slot read past the end of a non-`StructType` allocation — but fixing it means changing a fallback this PR deliberately chose, so it belongs in its own issue with the three options.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked that the open PRs keep iterating against the automated reviewer now that its runner problem is fixed; this answers the review that landed on `dc04441`.

JPHutchins:
**Nine findings down to two, and one of the two is now a fix.** `cfd661a`, green at 25/25.

**2 — taken.** This is #66, and JPH ruled option 1 on it today, so it stops being an out-of-scope decline. `Struct_hash` no longer falls back to `object`'s identity hash for a non-struct; it refuses.

You had the diagnosis exactly right: equality came from the co-base and hashing from `object`, and nothing made them agree.

```console
>>> imp == (1, 2)
True
>>> {(1, 2): "from the tuple"}.get(imp, "MISSED")
'MISSED'
```

Being unhashable is an ordinary thing for a Python object to be, and one that cannot be a dict key cannot break the invariant. It is also the only fallback in `mixin.c` that could not be a local decision — repr keeps its, because raising out of repr turns an object that is merely useless into one a debugger cannot print.

`test_it_hashes_by_identity` becomes `test_it_is_unhashable`, and the contract is now pinned over a tuple co-base rather than only the hash half of it.

**1 — #55, unchanged.** Still reachable through `type(Struct)`, still out of this PR's scope, and your recommendation is recorded there as the concrete shape of a fix — verify the installed table against the outer plan, because `struct_field_names == NULL` cannot tell "built by my delegate" from "found by my delegate".

<details>
<summary><b>One change from CLAUDE.md rather than from the review</b></summary>

`struct_metadata` matched its two-valued enum with a ternary, so a third kind of metadata would have silently taken the false arm. It is a switch with no `default:` now, which makes that a compiler error instead:

```c
	switch (which) {
		case STRUCT_FIELD_NAMES:
			return struct_tuple_or_empty(type->struct_field_names);
		case STRUCT_DEFAULTS:
			return struct_tuple_or_empty(type->struct_defaults);
	}

	Py_UNREACHABLE();
```

</details>

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked that the open PRs keep iterating against the automated reviewer, and ruled on #66 in the same session; this answers the review that landed on `51cfd80`.

JPHutchins:
**Three taken, one is #55.** `d4bf0c5`, green at 25/25.

Findings 2 and 3 are the same observation from two angles, and both are right: `mixin.c`'s comment explains the fallback for repr and hash and says nothing about the two places that same policy has consequences.

**3 — setattr does not fall back, it overrides.** Built the case:

```
the co-base alone      records ['z']
an impostor over it    records []          and the value lands anyway
```

A co-base with its own `__setattr__` never sees the write. Same trade as repr — honouring it means walking the MRO for a setter to borrow — but "falls back to `PyObject_GenericSetAttr`" reads as *deferring* when it is *overriding*, and that difference is now stated and pinned.

**2 — equality is intransitive, and #66's fix does not change it.**

```
left == (1, 2)   True
right == (1, 2)  True
left == right    False
```

`rich_compare` answers NotImplemented for a non-struct, so two content-equal impostors fall back to identity while each compares equal to the plain value they wrap. That is a consequence of the fallback policy rather than a decision beside it, so it goes in the same comment and is asserted rather than left to be discovered.

**4 — taken.** The three `is not safe` refusals were one assertion written three times; parametrized over the metatype. The fourth route stays separate because the parametrization cannot express it: `Substituting("S", ...)` is the metaclass *call* reaching `type.__new__`, not a direct call to it. The builder moved to module level — a `parametrize` in a class body cannot name its own class, which the first attempt discovered.

**1 — #55**, unchanged. Reproducing at HEAD is expected; the route narrowed when the metaclass stopped being exported but did not close, and your root-cause note — that `struct_field_names == NULL` cannot distinguish "built by my delegate" from "found by my delegate" — is recorded there as the shape a fix has to take.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked that the open PRs keep iterating against the automated reviewer; this answers the review that landed on `cfd661a`.

JPHutchins:
**Two taken, one refuted with the case it describes, one deferred with a test instead of a sentence.** `f02077c`, green at 25/25.

## 2 — the out-of-bounds read does not reproduce

I built every route in the finding. The getters can only ever be handed an instance of the metaclass, and it is CPython's own descriptor check that guarantees it:

```console
descriptor.__get__(META)      TypeError: descriptor '__struct_fields__' for 'salix._StructMeta' objects doesn't apply to a 'type' object
descriptor.__get__(type)      TypeError
descriptor.__get__(int)       TypeError
descriptor.__get__(object())  TypeError

META.__struct_fields__     -> <attribute '__struct_fields__' of 'salix._StructMeta' objects>
Derived.__struct_fields__  -> ('x',)
```

The premise is that `self` can be `&StructMeta_Type`, and it cannot: `type(type(Struct)) is type`, so META is **not** an instance of itself. An attribute lookup on META finds the descriptor in its own `__dict__` with `obj = NULL` and gets the descriptor back unread — which is the second-to-last line above.

And every object that *can* reach the getter is allocated at `META.__basicsize__`, because that is what `type_new` sizes an instance of META at:

```console
META.__basicsize__   992      # sizeof(StructType)
type.__basicsize__   936
Meta.__basicsize__   992      # class Meta(META): pass
```

<details>
<summary>Also checked: a metaclass subclass, and a class built through it</summary>

```console
class Meta(META): pass
Meta.__struct_fields__       -> the descriptor        # Meta is an instance of type, not of META
descriptor.__get__(Meta)     -> TypeError

class Derived(Struct, metaclass=Meta): x: int
Derived.__struct_fields__    -> ('x',)                # 992-byte instance of Meta
```
</details>

`test_its_own_metadata_getters_are_never_handed_the_metaclass` now asserts all three legs of that, so the reasoning is not what the safety rests on.

## 3 — the Hashable divergence is real, and accepted

Confirmed. `tp_hash` is the slot every struct hashes through, so it cannot be NULL for the impostor alone, and the ABC asks for the slot rather than the call. CPython does the same thing:

```console
isinstance(impostor, Hashable)                       True
hash(impostor)                                       TypeError: unhashable type: 'Impostor'
isinstance(memoryview(bytearray(b"ab")), Hashable)   True
hash(memoryview(bytearray(b"ab")))                   ValueError: cannot hash writable memoryview object
```

Documented in `hash.c` and asserted — including the memoryview half, so the precedent is a test rather than a claim.

## 4 — the benchmark said more than it showed

Taken. Two identical ranges bound the walk by the width of the band; they do not show it is free. The comment says that now.

## 1 — still #55's, but no longer only prose

It reproduces at this head, exactly as you describe:

| built through | default `y = 42` | `order=True` |
| --- | --- | --- |
| a plain `META` subclass | kept | works |
| a `__new__` that only forwards | **dropped** | **dropped** |
| the class statement | kept | works |

The fix is forwarding the caller's keywords and the untransformed namespace through the `type_new` hand-off — `create_class` surgery, and a different subject from this PR's #15/#28/#29. What I have done instead is put it beside the two tests that *do* pass, as `xfail(strict=True)`:

```python
@pytest.mark.xfail(strict=True, reason="#55")
def test_both_survive_a_delegate_that_writes_its_own_new(self):
```

Same convention as [`0c59347`](https://github.com/JPHutchins/salix/commit/0c59347) for #58: it is red for the right reason today and errors as an unexpected pass the moment #55 lands, which a fix cannot walk past.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, answering this round's findings on her PR.

JPHutchins:
**Both taken, one of them halfway on purpose.** `fd089e5`, green at 25/25 and `nix flake check` green in 246s.

## 1 — refused rather than mis-planned

You are right that it still reproduces, and the whole fix is #55's: forwarding the keywords and the untransformed namespace through the `type_new` hand-off is `create_class` surgery. Your second recommendation is the half that belongs here, and it is what this PR is about — a class that is not what it claims should be refused, not returned.

The outer call now compares the returned class against its own plan (field names, default count, options) and raises where they differ:

```console
META("Built", (Base,), {... "y": 42}, order=True)
  via a plain META subclass          y == 42, order works        unchanged
  via a forwarding __new__           TypeError: Forwarding.__new__ returned a
                                     struct class this call did not plan ...
  class Statement(Base, order=True)  y == 42, order works        unchanged
```

The third row has a test of its own now — the class statement never reaches the hand-off, because there the winner *is* the requested metatype, and a refusal that reached the ordinary spelling would be worse than the bug.

The `xfail(strict=True)` stays, because refused is not fixed: what #55 asks for is the first row's answer from the second row's spelling.

## 2 — it cannot usefully re-check

`struct_create_root` has one caller, and it builds both arguments four lines earlier:

```c
PY_OWNED(bases, PyTuple_Pack(1, (PyObject *) &StructMixin_Type));
PY_OWNED(namespace, PyDict_New());
...
return struct_create_root(name, bases, namespace);
```

`PyArg_ParseTuple`'s `O!` checks existed because `StructMeta_new` is reached from Python, where the arguments are whatever the caller passed. This entry point is reached from module init, in C, with objects it constructed. A `PyTuple_Check` there is a branch that cannot be taken; keeping the helper out of `meta.h` is not available either, since `salix.c` is a different translation unit.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, answering the latest round on her PR.

Issue #46, opened 2026-07-31
repr, hash and == read slots unsynchronised, so a concurrent write segfaults on a free-threaded build


JPHutchins:
`set_field` was made thread-safe in #7 by delegating to `PyMember_SetOne`, which takes a critical section around the store and defers the release past it. The **read** side was not, and it crashes.

```console
$ python3.14t reader_race.py    # 4 writers, 3 repr() readers, 3 == readers, one shared struct
gil enabled: False
Segmentation fault (core dumped)      # exit 139, 3/3 runs
```

This is **not** specific to `set_field`. The same script with `frozen=False` and `shared.value = [tag, i]` — CPython's own member descriptor doing the writing — crashes the same way:

```
descriptor-write exit=139
descriptor-write exit=134
descriptor-write exit=134
```

So the defect is in salix's readers, not in either writer, and it predates #7's fix.

<details>
<summary><b>Why the readers race</b></summary>

Every C-level reader loads a slot with a plain dereference and then increfs what it found. [`src/types.h`](https://github.com/JPHutchins/salix/blob/main/src/types.h):

```c
static inline PyObject * struct_slot_or_none(
	StructType const * const type,
	PyObject * const self,
	Py_ssize_t const index
) {
	PyObject * const value = *struct_slot(type, self, index);

	return value != NULL ? value : Py_None;
}
```

A writer replaces the slot and then releases the previous value. `PyMember_SetOne` defers that release until *after* its critical section, which is what makes two concurrent writers safe — but a reader holds no critical section at all, so the window between its load and its incref is unprotected. The pointer it loaded can be the one the writer is about to free.

The readers are `Struct_repr` via `field_repr`, `Struct_hash`, and both comparison paths in `src/compare.c`. Note that #30's fix — taking owned references in `ordering_result` — closed a use-after-free *within* a comparison; it does not close this one, because `Py_NewRef` on an already-freed pointer is itself the race.

</details>

<details>
<summary><b>What it costs, and the options</b></summary>

salix ships 15 free-threaded wheels, and `Py_mod_gil` declares `Py_MOD_GIL_NOT_USED`. Reading a struct from one thread while another writes it is not exotic — it is what a shared cache or a worker pool does — and the documented-safe write path (`self.x = v` on a mutable struct) is enough to trigger it.

[...]

Issue #18, opened 2026-07-31
README becomes the PyPI landing page and points at four things a reader cannot reach


JPHutchins:
`readme = "README.md"` means this file becomes the PyPI project page. Judged as a repository README it is fine and deliberately terse. Judged as the landing page — which is what publishing makes it — it points a reader at four things they cannot reach, and never says how to install the package.

The whole body, after the doctest:

> `salix/__init__.pyi` is the API. `src/salix.c` says what this is and is not. `tasks.py` is what runs. `uv run camas benchmark` says what it costs.

For someone arriving from a PyPI search:

| reference | reachable from the PyPI page? |
| --- | --- |
| `salix/__init__.pyi` | in the wheel, but they have not installed anything yet |
| `src/salix.c` | not in the wheel; in the sdist only |
| `tasks.py` | in neither the wheel nor the sdist |
| `uv run camas benchmark` | needs the repo, nix, and camas |

And there is no link to the repository, because the metadata has no `Project-URL` — #17. Every one of those four is a dead end.

<details>
<summary><b>What is missing for a reader who has never seen this project</b></summary>

- **How to install it.** `pip install salix` appears nowhere.
- **Which interpreters it runs on.** 3.10–3.15 plus free-threaded, 50 wheels — not stated.
- **What it is for, and when not to use it.** The one-line summary sells "programs that pay for their imports", which is the real pitch, but nothing says how salix compares to `dataclasses`, `NamedTuple`, `attrs` or `msgspec`, or that it does no serialisation or validation. `src/salix.c`'s header comment answers this well and no reader will ever see it.
- **The headline numbers.** They were removed deliberately at `07e205c` because `camas benchmark` prints them, but a PyPI reader cannot run that command. Two numbers in prose — import cost and type-creation time, with the comparison and the date — would carry the claim the summary line already makes.
- **The feature set.** A reader sees a two-field `Point`. Not shown: `frozen`, `eq`, `order`, `repr`, `match_args`, `weakref`, `__post_init__`, `set_field`, defaults, inheritance, pattern matching.
- **The semantics that will surprise people.** Equality is *structural* — `A(1) == B(1)` is `True` for two different classes with the same field names, which is a deliberate design call and diverges from `dataclasses`. It is documented in [src/compare.c:90](https://github.com/JPHutchins/salix/blob/74d2647/src/compare.c#L90) and nowhere a user will look.

</details>

<details>
<summary><b>Constraints worth preserving</b></summary>

[...]

Issue #13, opened 2026-07-31
Structs cannot be pickled, copied, or deep-copied


JPHutchins:
A struct cannot be pickled, copied, or deep-copied:

```python
>>> class P(Struct):
...     x: int
...     y: int = 2

>>> pickle.dumps(P(1))
TypeError: cannot pickle 'P' object
>>> copy.copy(P(1))
TypeError: cannot pickle 'P' object
>>> copy.deepcopy(P(1))
TypeError: cannot pickle 'P' object
```

For a data-container library this is a first-class feature gap. `dataclasses`, `NamedTuple`, `attrs` and `msgspec.Struct` are all picklable and copyable. It also rules salix out of multiprocessing, `concurrent.futures`, most caching layers, and any `copy.replace`-shaped workflow.

Nothing in the suite covers it — `grep -rn "pickle\|copy" tests/` finds nothing.

<details>
<summary><b>Root cause</b></summary>

CPython's default reducer refuses a type with no `tp_new`. From CPython 3.14, [`Objects/typeobject.c:7660`](https://github.com/python/cpython/blob/3.14/Objects/typeobject.c#L7660), in `reduce_newobj` (declarations elided):

```c
static PyObject *
reduce_newobj(PyObject *obj)
{
    /* ... */
    if (Py_TYPE(obj)->tp_new == NULL) {
        PyErr_Format(PyExc_TypeError,
                     "cannot pickle '%.200s' object",
                     Py_TYPE(obj)->tp_name);
        return NULL;
    }
```

That message is verbatim what salix produces, so this is the branch.

`tp_new` is NULL because `StructMixin_Type` is a static type that declares none ([src/mixin.c:15](https://github.com/JPHutchins/salix/blob/74d2647/src/mixin.c#L15)) and struct classes only get one on the path that declines the vectorcall — [src/meta.c:451](https://github.com/JPHutchins/salix/blob/74d2647/src/meta.c#L451):

```c
	if (defines_own_init(struct_class)) {
		struct_class->heap_type.ht_type.tp_new = PyType_GenericNew;
	} else {
		struct_class->heap_type.ht_type.tp_vectorcall = Struct_vectorcall;
	}
```

This is deliberate and the comment says why:

> The mixin has no `tp_new`, because nothing ever needed one: the vectorcall allocates. [...] so the mixin itself stays uninstantiable and nothing can hold a struct's dunders over an object that has no field table.

[...]

PR #49, opened 2026-08-01
fix: give each instance its own mutable default

### LINKED ISSUE #19: A mutable default is shared by every instance; dataclasses rejects this
A mutable default is shared by every instance of the class, silently:

```python
>>> class A(Struct, frozen=False):
...     xs: list = []

>>> a, b = A(), A()
>>> a.xs.append(1)
>>> b.xs
[1]
```

`dataclasses` refuses this outright:

```
ValueError: mutable default <class 'list'> for field xs is not allowed: use default_factory
```

salix accepts it and hands every instance the same object.

<details>
<summary><b>Where the shared reference comes from</b></summary>

The defaults tuple is built once at class creation ([src/fields.c:172](https://github.com/JPHutchins/salix/blob/74d2647/src/fields.c#L172)) and the constructor hands out a new reference to the same object every time — [src/construct.c:186](https://github.com/JPHutchins/salix/blob/74d2647/src/construct.c#L186):

```c
		*slot = Py_NewRef(PyTuple_GET_ITEM(type->struct_defaults, i - required_count));
```

Which is correct and is the whole point: no per-construction allocation is what makes the constructor fast. The problem is not the mechanism, it is that nothing warns.

</details>

<details>
<summary><b>Why the frozen default softens this, and why it does not remove it</b></summary>

`frozen=True` is the default, and a frozen struct cannot rebind `xs`. But it can still *mutate* what `xs` points at — `a.xs.append(1)` works on a frozen struct too, exactly as it does on a frozen dataclass — so the sharing is observable at the default setting as well. The example above uses `frozen=False` only to make the ordinary case obvious.

</details>

<details>

**1. Reject a mutable default,** the way `dataclasses` does. Its test is a small blocklist (`list`, `dict`, `set`) plus an unhashability check; the modern implementation uses `__hash__ is None`. That last form is one lookup per defaulted field at class creation, which is cheap and is exactly the "provably correct at compile time" instinct. The cost: no `default_factory` exists here, so rejecting leaves the user with `__post_init__` and `set_field` as the workaround — which does work, and is arguably the salix-shaped answer.

**2. Add `default_factory`.** Full compatibility, but it means a field specifier object, `field_specifiers` in the `dataclass_transform` declaration, and a call per construction on a path that currently makes none. That is a real cost against the thing salix sells.

**3. Document it.** `NamedTuple` shares mutable defaults too and nobody considers that a bug, because tuples advertise immutability. salix's default is frozen, so the same argument partly applies.

Option 1 looks right for a library whose defaults are frozen: reject the footgun, do not grow a factory mechanism, and point at `__post_init__` in the message. But this is a design call — and it is one that gets much more expensive after v1, because rejecting something previously accepted is a breaking change.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Closes #19. `xs: list = []` handed every instance the same list, silently:

```python
class A(Struct, frozen=False):
    xs: list = []

a, b = A(), A()
a.xs.append(1)
b.xs                        # [1]
```

The four builtins that mean *container I will mutate* — `list`, `dict`, `set`, `bytearray` — are copied at construction now. Everything else is shared, which is what you want for an `int`, a `str`, a `tuple`, or a frozen struct: none of them can be changed out from under another instance.

salix's own nested structs land on the right side of that line for free. `frozen` is the default, a frozen struct cannot be rebound, and it is hashable — so sharing it is as correct as sharing an int. A `frozen=False` struct is unhashable, which is exactly the signal that it would have been a real footgun.

<details>
<summary><b>Why copy rather than reject, which is what dataclasses does</b></summary>

The first proposal was to refuse the shape the way `dataclasses` does and point at a `default_factory` salix does not have. @JPHutchins asked what it actually costs. Measured against msgspec, the rival this project benchmarks itself against:

```
msgspec, bare `xs: list = []`     copied per instance, not shared
  Plain(a, b, c)                    55.3 ns
  WithMutable(a, b)                 61.8 ns      ← the copy
  explicit default_factory          65.0 ns
```

msgspec already makes the common spelling mean what it looks like it means, for about 6 ns on the field that has one. Rejecting buys nothing and costs the API. Its rule, also measured: `list`/`dict`/`set`/`bytearray` copied, arbitrary mutable objects shared — which is what this implements.

</details>

<details>
<summary><b>No measurable cost to the path that does not copy</b></summary>

Four type compares now run for every default, so the question is whether the common case regressed. Measured against a control with no defaults at all, which the change cannot touch:

```
before   control 48.44 / 50.54    immutable default 46.64 / 46.99
after    control 46.94 / 46.92    immutable default 46.76 / 46.24
```

Run-to-run spread on the control is wider than any difference in the treatment. An earlier single-number comparison appeared to show the *new* build being faster, which is how I knew it was noise rather than signal.

</details>

### One behaviour change worth naming

`test_a_value_may_be_a_default` asserted `Local().field is value` for every value in the set — it encoded the old contract precisely. It now asserts equality always, identity for the shared kinds, and non-identity for the copied ones.

A **supplied** value is still stored and not copied; only the default is the class's to hand out repeatedly. `test_a_supplied_value_is_still_stored_rather_than_copied` pins that.

Seven tests in `tests/test_construction.py`. `camas check` green at 25/25.

> [!WARNING]
> **LLM Disclosure**
>
> This PR was authored by claude-opus-5 on behalf of @JPHutchins, implementing the ruling she gave on #19 after pushing back on the first recommendation and asking what the cost actually was.

### DIFF
--- a/src/construct.c
+++ b/src/construct.c
@@ -236,6 +236,10 @@ static enum result fill_defaults(
 PyObject * struct_default_copy(PyObject * const declared) {
 	PyTypeObject * const kind = Py_TYPE(declared);
 
+	if (!struct_copies_default(kind)) {
+		return Py_NewRef(declared);
+	}
+
 	if (kind == &PyList_Type) {
 		return PyList_GetSlice(declared, 0, PyList_GET_SIZE(declared));
 	}
@@ -248,11 +252,7 @@ PyObject * struct_default_copy(PyObject * const declared) {
 		return PySet_New(declared);
 	}
 
-	if (kind == &PyByteArray_Type) {
-		return PyByteArray_FromObject(declared);
-	}
-
-	return Py_NewRef(declared);
+	return PyByteArray_FromObject(declared);
 }
 
 /*
--- a/src/construct.c
+++ b/src/construct.c
@@ -208,14 +208,19 @@ static enum result fill_defaults(
  * A mutable default belongs to the instance, not to the class: `xs: list = []`
  * reads as an empty list per struct, and handing every instance the same one is
  * a bug people write by accident. The four builtins that spell "container I
- * will mutate" are copied; everything else is shared, which is what you want
- * for an int, a string, a tuple, or a frozen struct -- none of them can be
- * changed out from under another instance.
+ * will mutate" are copied, and only when empty -- a non-empty one is refused at
+ * class creation, since copying it could only be shallow.
  *
- * dataclasses refuses this shape instead, and needs default_factory to express
- * it at all. msgspec copies, and so does this: it costs the copy only on the
- * fields that have one, and it makes the common spelling mean what it looks
- * like it means.
+ * Everything else is shared. That is right for an int, a string or a tuple of
+ * them, and for a frozen struct of them: none can be rebound or mutated. It is
+ * *not* right for a shallowly-immutable container of something mutable --
+ * `([],)`, or a frozen struct holding a list -- which is shared here and whose
+ * contents therefore are too. Hashability is the test that would catch those,
+ * and #51 is where that is argued; msgspec shares them as well.
+ *
+ * dataclasses refuses the shape outright and needs default_factory to express
+ * it at all. This copies, so the common spelling means what it looks like it
+ * means, and pays for it only on the fields that have one.
  */
 static PyObject * default_for(PyObject * const declared) {
 	PyTypeObject * const kind = Py_TYPE(declared);
--- a/src/construct.c
+++ b/src/construct.c
@@ -216,12 +216,18 @@ static enum result fill_defaults(
  * and every instance would get a shallow copy of it. msgspec severs the same
  * alias by turning the default into a Factory.
  *
- * Everything else is shared. That is right for an int, a string or a tuple of
- * them, and for a frozen struct of them: none can be rebound or mutated. It is
- * *not* right for a shallowly-immutable container of something mutable --
- * `([],)`, or a frozen struct holding a list -- which is shared here and whose
- * contents therefore are too. Hashability is the test that would catch those,
- * and #51 is where that is argued; msgspec shares them as well.
+ * Everything else is shared, and "everything else" is wider than it sounds.
+ * Sharing is right for an int, a string or a tuple of them, and for a frozen
+ * struct of them: none can be rebound or mutated. It is wrong for two kinds of
+ * default this does not reach -- a shallowly-immutable container of something
+ * mutable (`([],)`, a frozen struct holding a list), and a mutable container
+ * that simply is not one of the four (`array.array`, `deque`, `defaultdict`, a
+ * writable `memoryview`, or a subclass of any of the four). Those are shared
+ * outright, and a non-empty one is not refused either, because the refusal
+ * whitelists the same four types.
+ *
+ * Every one of them is unhashable, which is the single test that would replace
+ * this list; #51 is where that is argued. msgspec shares them as well.
  *
  * dataclasses refuses the shape outright and needs default_factory to express
  * it at all. This copies, so the common spelling means what it looks like it
--- a/src/construct.c
+++ b/src/construct.c
@@ -232,29 +232,49 @@ static enum result fill_defaults(
  * dataclasses refuses the shape outright and needs default_factory to express
  * it at all. This copies, so the common spelling means what it looks like it
  * means, and pays for it only on the fields that have one.
+ *
+ * A row is a type and the constructor that copies it, because no predicate can
+ * supply the second. One table, so the refusal and the copy cannot come to
+ * different answers about which types those are.
  */
+struct default_copy {
+	PyTypeObject const * kind;
+	PyObject * (*copy)(PyObject * declared);
+};
+
+/* PyList_GetSlice takes bounds; the other three constructors take the object. */
+static PyObject * copy_list(PyObject * const declared) {
+	return PyList_GetSlice(declared, 0, PyList_GET_SIZE(declared));
+}
+
+static struct default_copy const COPIED_WHEN_EMPTY[] = {
+	{.kind = &PyList_Type, .copy = copy_list},
+	{.kind = &PyDict_Type, .copy = PyDict_Copy},
+	{.kind = &PySet_Type, .copy = PySet_New},
+	{.kind = &PyByteArray_Type, .copy = PyByteArray_FromObject},
+};
+
+static struct default_copy const * copies_default(PyTypeObject const * const kind) {
+	for (size_t at = 0; at < sizeof COPIED_WHEN_EMPTY / sizeof *COPIED_WHEN_EMPTY; ++at) {
+		if (COPIED_WHEN_EMPTY[at].kind == kind) {
+			return &COPIED_WHEN_EMPTY[at];
+		}
+	}
+
+	return NULL;
+}
+
+bool struct_copies_default(PyTypeObject const * const kind) {
+	return copies_default(kind) != NULL;
+}
+
 PyObject * struct_default_copy(PyObject * const declared) {
-	PyTypeObject * const kind = Py_TYPE(declared);
+	struct default_copy const * const copier = copies_default(Py_TYPE(declared));
 
-	if (kind == &PyList_Type) {
-		return PyList_GetSlice(declared, 0, PyList_GET_SIZE(declared));
-	}
-
-	if (kind == &PyDict_Type) {
-		return PyDict_Copy(declared);
-	}
-
-	if (kind == &PySet_Type) {
-		return PySet_New(declared);
-	}
-
-	if (kind == &PyByteArray_Type) {
-		return PyByteArray_FromObject(declared);
-	}
-
-	/* Everything else, which is the answer that degrades safely: copying with a
-	 * constructor for the wrong type would change the value. */
-	return Py_NewRef(declared);
+	/* Everything else is the object itself, a subclass of one of the four
+	 * included: copying with a constructor for the wrong type would change the
+	 * value. */
+	return copier != NULL ? copier->copy(declared) : Py_NewRef(declared);
 }
 
 /*
--- a/src/construct.c
+++ b/src/construct.c
@@ -236,10 +236,6 @@ static enum result fill_defaults(
 PyObject * struct_default_copy(PyObject * const declared) {
 	PyTypeObject * const kind = Py_TYPE(declared);
 
-	if (!struct_copies_default(kind)) {
-		return Py_NewRef(declared);
-	}
-
 	if (kind == &PyList_Type) {
 		return PyList_GetSlice(declared, 0, PyList_GET_SIZE(declared));
 	}
@@ -256,9 +252,8 @@ PyObject * struct_default_copy(PyObject * const declared) {
 		return PyByteArray_FromObject(declared);
 	}
 
-	/* Unreachable while struct_copies_default names these four and no more. If
-	 * that stops being true, sharing is the answer that degrades: copying with
-	 * the wrong constructor would change the value's type. */
+	/* Everything else, which is the answer that degrades safely: copying with a
+	 * constructor for the wrong type would change the value. */
 	return Py_NewRef(declared);
 }
 
--- a/src/construct.c
+++ b/src/construct.c
@@ -38,6 +38,7 @@ static enum result write_slot(
 	PyObject * value
 );
 static enum result run_post_init(StructType const * type, PyObject * self);
+static PyObject * default_for(PyObject * declared);
 
 /*
  * Instances are built straight into slot memory: allocate, then let each of
@@ -189,10 +190,53 @@ static enum result fill_defaults(
 			return RESULT_ERROR;
 		}
 
-		*slot = Py_NewRef(PyTuple_GET_ITEM(type->struct_defaults, i - required_count));
+		PyObject * const value = default_for(
+			PyTuple_GET_ITEM(type->struct_defaults, i - required_count)
+		);
+
+		if (value == NULL) {
+			return RESULT_ERROR;
+		}
+
+		*slot = value;
 	}
 
 	return RESULT_OK;
+}
+
+/*
+ * A mutable default belongs to the instance, not to the class: `xs: list = []`
+ * reads as an empty list per struct, and handing every instance the same one is
+ * a bug people write by accident. The four builtins that spell "container I
+ * will mutate" are copied; everything else is shared, which is what you want
+ * for an int, a string, a tuple, or a frozen struct -- none of them can be
+ * changed out from under another instance.
+ *
+ * dataclasses refuses this shape instead, and needs default_factory to express
+ * it at all. msgspec copies, and so does this: it costs the copy only on the
+ * fields that have one, and it makes the common spelling mean what it looks
+ * like it means.
+ */
+static PyObject * default_for(PyObject * const declared) {
+	PyTypeObject * const kind = Py_TYPE(declared);
+
+	if (kind == &PyList_Type) {
+		return PyList_GetSlice(declared, 0, PyList_GET_SIZE(declared));
+	}
+
+	if (kind == &PyDict_Type) {
+		return PyDict_Copy(declared);
+	}
+
+	if (kind == &PySet_Type) {
+		return PySet_New(declared);
+	}
+
+	if (kind == &PyByteArray_Type) {
+		return PyByteArray_FromObject(declared);
+	}
+
+	return Py_NewRef(declared);
 }
 
 /*
--- a/src/construct.c
+++ b/src/construct.c
@@ -252,7 +252,14 @@ PyObject * struct_default_copy(PyObject * const declared) {
 		return PySet_New(declared);
 	}
 
-	return PyByteArray_FromObject(declared);
+	if (kind == &PyByteArray_Type) {
+		return PyByteArray_FromObject(declared);
+	}
+
+	/* Unreachable while struct_copies_default names these four and no more. If
+	 * that stops being true, sharing is the answer that degrades: copying with
+	 * the wrong constructor would change the value's type. */
+	return Py_NewRef(declared);
 }
 
 /*
--- a/src/construct.c
+++ b/src/construct.c
@@ -38,7 +38,6 @@ static enum result write_slot(
 	PyObject * value
 );
 static enum result run_post_init(StructType const * type, PyObject * self);
-static PyObject * default_for(PyObject * declared);
 
 /*
  * Instances are built straight into slot memory: allocate, then let each of
@@ -190,7 +189,7 @@ static enum result fill_defaults(
 			return RESULT_ERROR;
 		}
 
-		PyObject * const value = default_for(
+		PyObject * const value = struct_default_copy(
 			PyTuple_GET_ITEM(type->struct_defaults, i - required_count)
 		);
 
@@ -211,6 +210,12 @@ static enum result fill_defaults(
  * will mutate" are copied, and only when empty -- a non-empty one is refused at
  * class creation, since copying it could only be shallow.
  *
+ * Class creation takes a copy too, so what the class stores is not the object
+ * the body named. `shared = []` kept at module level and appended to afterwards
+ * would otherwise make the stored default non-empty behind the refusal's back,
+ * and every instance would get a shallow copy of it. msgspec severs the same
+ * alias by turning the default into a Factory.
+ *
  * Everything else is shared. That is right for an int, a string or a tuple of
  * them, and for a frozen struct of them: none can be rebound or mutated. It is
  * *not* right for a shallowly-immutable container of something mutable --
@@ -222,7 +227,7 @@ static enum result fill_defaults(
  * it at all. This copies, so the common spelling means what it looks like it
  * means, and pays for it only on the fields that have one.
  */
-static PyObject * default_for(PyObject * const declared) {
+PyObject * struct_default_copy(PyObject * const declared) {
 	PyTypeObject * const kind = Py_TYPE(declared);
 
 	if (kind == &PyList_Type) {
--- a/src/construct.c
+++ b/src/construct.c
@@ -233,32 +233,37 @@ static enum result fill_defaults(
  * it at all. This copies, so the common spelling means what it looks like it
  * means, and pays for it only on the fields that have one.
  *
- * A row is a type and the constructor that copies it, because no predicate can
- * supply the second. One table, so the refusal and the copy cannot come to
- * different answers about which types those are.
+ * The type and the constructor that copies it are named together, because no
+ * predicate can supply the second. One list, so the refusal and the copy cannot
+ * come to different answers about which types those are.
+ *
+ * A list of statements rather than a static array of {type, constructor}: on
+ * Windows a `PyTypeObject` is imported from python3.dll, and the address of a
+ * dllimport symbol is not a compile-time constant, so the array version
+ * compiles everywhere except the platform half the wheels are cross-built for.
  */
-struct default_copy {
-	PyTypeObject const * kind;
-	PyObject * (*copy)(PyObject * declared);
-};
+typedef PyObject * (*default_copier)(PyObject * declared);
 
 /* PyList_GetSlice takes bounds; the other three constructors take the object. */
 static PyObject * copy_list(PyObject * const declared) {
 	return PyList_GetSlice(declared, 0, PyList_GET_SIZE(declared));
 }
 
-static struct default_copy const COPIED_WHEN_EMPTY[] = {
-	{.kind = &PyList_Type, .copy = copy_list},
-	{.kind = &PyDict_Type, .copy = PyDict_Copy},
-	{.kind = &PySet_Type, .copy = PySet_New},
-	{.kind = &PyByteArray_Type, .copy = PyByteArray_FromObject},
-};
+static default_copier copies_default(PyTypeObject const * const kind) {
+	if (kind == &PyList_Type) {
+		return copy_list;
+	}
 
-static struct default_copy const * copies_default(PyTypeObject const * const kind) {
-	for (size_t at = 0; at < sizeof COPIED_WHEN_EMPTY / sizeof *COPIED_WHEN_EMPTY; ++at) {
-		if (COPIED_WHEN_EMPTY[at].kind == kind) {
-			return &COPIED_WHEN_EMPTY[at];
-		}
+	if (kind == &PyDict_Type) {
+		return PyDict_Copy;
+	}
+
+	if (kind == &PySet_Type) {
+		return PySet_New;
+	}
+
+	if (kind == &PyByteArray_Type) {
+		return PyByteArray_FromObject;
 	}
 
 	return NULL;
@@ -269,12 +274,12 @@ bool struct_copies_default(PyTypeObject const * const kind) {
 }
 
 PyObject * struct_default_copy(PyObject * const declared) {
-	struct default_copy const * const copier = copies_default(Py_TYPE(declared));
+	default_copier const copy = copies_default(Py_TYPE(declared));
 
 	/* Everything else is the object itself, a subclass of one of the four
 	 * included: copying with a constructor for the wrong type would change the
 	 * value. */
-	return copier != NULL ? copier->copy(declared) : Py_NewRef(declared);
+	return copy != NULL ? copy(declared) : Py_NewRef(declared);
 }
 
 /*
--- a/src/construct.h
+++ b/src/construct.h
@@ -14,24 +14,14 @@ PyObject * Struct_vectorcall(
 PyObject * Struct_set_field(PyObject * module, PyObject * arguments);
 
 /*
- * The four exact builtins that spell "container I will mutate", as the refusal
- * sees them. Copying names them a second time, in struct_default_copy, because
- * each one needs its own constructor and no predicate can supply that.
- *
- * The two lists have to agree and nothing makes them: a type copied but not
- * refused gets shallow-copied while non-empty, and a type refused but not
- * copied has its emptiness checked on the caller's object rather than on a
- * private one. A type added here and not there is shared, which is the
- * direction that degrades safely; the other way round changes a value's type.
+ * Whether the type is one of the exact builtins that spell "container I will
+ * mutate" -- the question the refusal asks. It and the copy read one table, so
+ * the two cannot disagree: a type copied but not refused would be
+ * shallow-copied while non-empty, and a type refused but not copied would have
+ * its emptiness checked on the caller's object rather than on a private one.
+ * Neither drift is the safe direction; the table is why neither is reachable.
  */
-static inline bool struct_copies_default(PyTypeObject const * const kind) {
-	return (
-		kind == &PyList_Type ||
-		kind == &PyDict_Type ||
-		kind == &PySet_Type ||
-		kind == &PyByteArray_Type
-	);
-}
+bool struct_copies_default(PyTypeObject const * kind);
 
 /*
  * A default nothing else holds a reference to: a copy for the four, the object
--- a/src/construct.h
+++ b/src/construct.h
@@ -11,3 +11,11 @@ PyObject * Struct_vectorcall(
 
 /* salix.set_field(instance, name, value) -- see construct.c. */
 PyObject * Struct_set_field(PyObject * module, PyObject * arguments);
+
+/*
+ * A default nothing else holds a reference to: a copy for the four mutable
+ * builtins, the object itself for everything else. Class creation takes one so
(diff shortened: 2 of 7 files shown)


JPHutchins:
Findings 1, 2 and 3 turned out to be one thing, and measuring msgspec properly showed I had implemented the wrong rule. `camas check` green at 25/25.

**My earlier measurement was incomplete.** I tested `xs: list = []`, saw a copy, and concluded "msgspec copies mutable builtins". It copies *empty* ones. A non-empty one it refuses:

```
TypeError: Using a non-empty mutable collection ([1]) as a default value is unsafe.
           Instead configure a `default_factory` for this field.
```

**That is the answer to finding 2, and a better one than documenting the limitation.** Copying a non-empty container can only be shallow, so `xs: list = [[1]]` hands every instance its own outer list around the *same* inner one — this PR's own bug, one level down. An empty container has no contents to alias, so the case cannot arise. salix refuses it now, pointing at `__post_init__` and `set_field` since there is no `default_factory` here.

Measured, salix against msgspec, row for row:

```
[] {} set() bytearray()       copied per instance
[1] {'k': 1} [[1]]            refused
MyList() defaultdict deque    shared
(1, 2) and other immutables   shared
```

**Finding 1 — only exact types are copied.** Deliberate, and now said out loud in the code rather than left implicit: the copy has to preserve the type, and `PyDict_Copy` of a `defaultdict` is a `dict`. msgspec shares subclasses for exactly that reason. Tested.

**Finding 3 — the test asserted a contract the implementation does not provide.** Correct, and it was my bug: `isinstance(value, COPIED_PER_INSTANCE)` against C that compares exact types, passing only because the value set happens to contain no subclasses. It tests the real rule now, refusal included.

**Findings 5 and 6 — both fair.** A test named `..._still_shares_its_mutable_field` asserted the opposite of sharing; it is `test_frozen_does_not_make_the_copy_unnecessary` now. And the dynamic class construction is a plain class body.

<details>
<summary><b>4 — the free-threading read, not changed here</b></summary>

Real, and worth recording. Reaching it means mutating a class-level default concurrently with construction, which is already a bug on its own. More to the point it is the same family as [#46](https://github.com/JPHutchins/salix/issues/46): salix's C readers are unsynchronised generally — `repr`, `hash` and `==` all crash under a concurrent write today, measured at exit 139. Fixing one read site here while those stand would be piecemeal. #46 is where it belongs.

</details>

<details>
<summary><b>7 — PyList_NewCopy, declined</b></summary>

I could not find `PyList_NewCopy` or `Py_SHALLOW_COPY` in the public headers of any interpreter in the matrix — grepping 3.14's `include/` and `include/cpython/` finds neither. If it exists it is private (`_PyList_Copy`), which this project cannot use. `PyList_GetSlice` stays.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Three taken, one declined with reasoning. `camas check` green at 25/25.

**1 — the error message told the user to do something the error prevented.** Correct and embarrassing: "build it in `__post_init__` with `set_field`" cannot be followed when the class definition is what was refused. The default has to be an empty container first. Reworded, and the advice now works:

```python
class Works(Struct):
    xs: list = []
    def __post_init__(self):
        set_field(self, "xs", [1, 2])

a, b = Works(), Works(); a.xs.append(3)
#  a.xs = [1, 2, 3]   b.xs = [1, 2]
```

On the wider point — no `default_factory` escape hatch — @JPHutchins has deferred that deliberately, and confirmed the refusal itself is what she wants.

**2 — the comment's claim is false for a container of something mutable.** Right, and this is the sharpest finding of the round. `([],)` and a frozen struct holding a list are both shared here, and their contents with them; "cannot be changed out from under another instance" is true of the *reference* and not of what it reaches. The PR body's "a frozen struct is hashable, so sharing it is as correct as sharing an int" is false for exactly the same reason — `Inner([])` is unhashable.

The comment says so now and points at **[#51](https://github.com/JPHutchins/salix/issues/51)**, filed for it: hashability is precisely the test that separates the two, because `hash` on a container hashes its elements and salix's structural hash does the same for a struct. Worth noting this is where salix would go *past* msgspec rather than a gap against it — msgspec shares those as well.

**4 — the immutable-sharing test was partially vacuous.** Correct: `5` and `"x"` are shared by CPython whatever `default_for` does, so the test would have stayed green if it started copying immutables. It builds `10**20` and a concatenated string now, so only the real contract is under test.

<details>
<summary><b>3 — declined: re-checking emptiness at construction</b></summary>

Real inconsistency, and I would rather keep it than pay for it. Reaching it means mutating a class's *declared default* through `__struct_defaults__`, then constructing — and at that point the copy cannot rescue the situation anyway, since the class has already told every reader what its default is. Re-checking would put the emptiness test on the construction path, which is the one number this project protects, to defend against a shape that is out of contract.

The introspection attribute exposing live mutable objects is arguably the real issue there, and it is not this PR's.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Finding 2 is right and my earlier decline did not cover it. Finding 1 taken in part. `camas check` green at 25/25 on `ab94f5e`.

**2 — the retained alias.** Confirmed, and you are right that this is not the `__struct_defaults__` route I declined. `shared = []` at module level, appended to after the class statement, is ordinary Python, and the refusal had already waved the class through.

The fix is upstream of the copy: **class creation now takes its own copy**, so the emptiness the refusal checked is permanent.

```
shared now               : [[2]]
a.xs                     : []
class default is alias   : False
outer copied per instance: True
subclassing after mutation: []      (was a TypeError before)
immutable default shared : True
```

That closes the construction/inheritance disagreement too — both now read a stored default that nothing outside the class can fill. What is still reachable is `Holder.__struct_defaults__[0].append(...)`, which is the introspection route, and the reason I gave for declining that one does hold for it.

This is what msgspec does, measured rather than assumed:

<details>
<summary>msgspec severs the same alias, by replacing the default with a Factory</summary>

```python
shared = []
class G(msgspec.Struct):
    xs: list = shared
shared.append([2])

shared now      : [[2]]
a.xs            : []
class default   : (<factory>,)
default is alias: False
default type    : <class 'msgspec._core.Factory'>
```

salix has no Factory, but copying at class creation reaches the same place: the class's default is its own object, and it is empty forever.

</details>

`default_for` is now `struct_default_copy` in `construct.h`, so the copy rule is stated once and both callers use it rather than drifting.

**1 — the body-`__init__` over-fire.** Half taken. The message was wrong and is fixed: `__post_init__` genuinely cannot run for such a class — verified, a body `__init__` displaces the vectorcall via `defines_own_init` and `run_post_init` goes with it — so it now names both hooks, and `test_a_body_init_does_not_exempt_the_declared_default` matches on `non-empty list.*your own __init__`.

I am not exempting the class from the refusal, because dataclasses does not either:

```
dataclasses(init=False), mutable default   -> ValueError: mutable default <class 'list'> ... use default_factory
dataclasses(init=False) + body __init__    -> ValueError: mutable default <class 'list'> ... use default_factory
```

The declared default is not only the constructor's input — it is still published on `__struct_defaults__`, and a subclass that *does* take the generated constructor inherits it. "This constructor never reads it" is a property of one class in the chain, not of the declaration.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues and answer each finding on its PR. This is round 4 on #49.

JPHutchins:
Finding 3 fixed, finding 1 was right about the boundary, and finding 2 uncovered something bigger than the refusal it was about. `camas check` green at 25/25 on `a7db92e`.

**3 — the class-creation TOCTOU.** Taken, and it turned out to need no lock. The check and the copy were two reads of an object the module still holds; **copying first and checking the copy** closes the window, because by then nothing else has a reference to what is being checked. For the four copied types the check runs on private memory; for everything else both calls are no-ops, since the refusal whitelists exactly the types the copy handles.

**1 — mutable containers outside the four.** Confirmed, all of them:

| default | refused? | `H().v is H().v` | class stores the alias |
| --- | --- | --- | --- |
| `array.array('i', [1, 2])` | no | **True** | True |
| `memoryview(bytearray(b'abc'))` | no | **True** | True |
| `deque([1, 2])` | no | **True** | True |
| `defaultdict(list, {'k': [1]})` | no | **True** | True |

Not extending the whitelist, because you named the reason yourself: every one of them is unhashable, and #51 is the issue arguing that hashability is the test which replaces the list entirely. Adding `array.array` and `deque` now would be three more special cases in the thing #51 exists to delete.

What I have fixed is the prose, which was the actual defect. The class comment said everything not copied is shared and framed that as right for values that *cannot* be mutated; the `TestMutableDefaults` docstring said the shared group is "anything that cannot be changed out from under another instance". Both false, in the direction that hides the hazard. They now say where the boundary is, and `test_a_mutable_container_outside_the_four_is_shared_and_not_refused` pins it so it moves visibly if #51 lands.

**2 — the body-`__init__` over-fire.** My round-4 rationale had a factual error and you caught it. I said "a subclass that *does* take the generated constructor inherits it". No subclass can:

```
Plain(R)().x        -> unset
__init__ = None     -> TypeError: 'NoneType' object is not callable
```

But chasing that turned up the part worth acting on — **defaults are dead across the whole subtree, including ones the subclass declares itself**:

```python
class R(Struct, frozen=False):
    x: int = 7
    def __init__(self): pass

class T(R):
    y: int = 5

T().x   # AttributeError
T().y   # AttributeError   <- T never wrote an __init__
```

And the dataclasses comparison cuts the other way from how I used it. `@dataclass(init=False)` leaves `D.x == 7` readable, because a dataclass default stays on the class; a salix field is a `member_descriptor` and the default lives only in `__struct_defaults__`. So the class advertises a default no instance will ever have.

Filed as **#56** with three options, because it is not a refusal question — under the first option (write defaults in `tp_new`) the default becomes live and the refusal is plainly right, which dissolves finding 2 rather than answering it. I would rather settle that than exempt a case whose inertness is itself the bug.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues and answer each finding on its PR, splitting side-discussions into follow-up issues. This is round 5 on #49.

JPHutchins:
Five taken, two declined. Findings 3 and 7 are costs I introduced last round and they go away together. `camas check` green at 25/25 on `968040b`.

**3 and 7 — copy-first put work on the error path.** Both right, and 7 is the sharper one: `PySet_New` hashes every element, so a non-empty set default ran user `__hash__` code at class definition *on the way to a refusal*, and an element whose `__hash__` raises would replace the promised `TypeError` with its own exception.

The fix keeps the race closed and gives the error path back. **Check twice.** The first read is O(1), decides nothing, and stops a doomed default from being copied at all; the second is the authoritative one, on a copy nothing else can reach. A thread filling the declaration between them still cannot touch what the class keeps.

`test_a_set_default_is_refused_before_its_elements_are_hashed` pins it — and building the set literal has to happen *before* the counter is cleared, since constructing it hashes the element once before salix is involved. My first version of that test failed for exactly that reason.

**4 — the whitelist in two files.** Taken, and the reasoning is the part worth keeping: a type copied but not refused gets shallow-copied while non-empty, and a type refused but not copied has its emptiness checked on the caller's object. `struct_copies_default` in `construct.h` states the set once; both callers use it.

**1 — a `frozen=False` struct as a default.** Same disposition as `array.array` and `deque` — shared, not refused, and unhashable, which is #51's whole argument. Added to the parametrized test so salix's own type is on the list rather than implied by it.

One correction that matters for #51 itself: the test asks `hash(value)`, not `isinstance(value, Hashable)`. My first version used the ABC and **failed on memoryview** — a writable memoryview has a `__hash__` that raises when called, so the ABC says hashable and the call says otherwise. #51 proposes hashability as the test; it has to be the call.

**6 — no refcount coverage for the copy path.** Taken. `test_a_copied_default_hands_out_no_reference_to_itself` asserts the stored default's count does not move across ten constructions and that all ten copies are distinct. Its sibling is renamed `test_an_uncopied_default_is_shared_by_every_instance`, because "a default is shared" stopped being true of all of them.

**2 — the body-`__init__` refusal.** Unchanged, and **#56** is where it goes: under that issue's first option the default becomes live and the refusal is plainly right, so fixing it here would be answering a question that may not exist.

**5 — inherited defaults re-copied per subclass.** Declined. The redundancy is real, but the cost is one empty-container allocation per inherited defaulted field at class creation, and removing it means making "the class's default is its own" conditional on provenance. An unconditional invariant is worth more than that allocation, and #51 would rewrite this code anyway.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues and answer each finding on its PR. This is round 6 on #49.

JPHutchins:
Finding 3 taken; 1 and 2 unchanged. `camas check` green at 25/25 on the new head.

**3 — the comment overclaims.** Correct, and obviously so once written down: the copy is made *from* the declaration, so a fill landing between the two checks is of course captured by it. "Cannot reach the copy" was wrong.

What is guaranteed is narrower, and is the part worth stating: the second check reads the copy, so a captured fill is refused and the class never stores a non-empty default. What the race costs is the work the first check exists to skip — a set's copy hashing its elements on the way to a refusal — not the invariant. The comment says that now.

You are also right that it is unreachable under the GIL, since there is no bytecode boundary between the two checks. It is a comment about a free-threaded build, and it should describe one accurately.

**1 — the body-`__init__` refusal.** Unchanged, and **#56** is where it goes: under that issue's first option the default becomes live and the refusal is plainly right, so fixing it here would answer a question that may not survive the ruling.

**2 — inherited defaults re-copied.** Declined again, same reasoning as last round: one empty-container allocation per inherited defaulted field at class creation, against making "the class's default is its own" conditional on provenance. If #51 lands, this code is rewritten anyway.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues and answer each finding on its PR. This is round 7 on #49.

JPHutchins:
Five taken, two unchanged. Finding 2 is the sharpest kind: a test whose own docstring names the trap it fell into. `camas check` green at 25/25 on `8d705e2`.

**2 — the "not interned" string was folded to a constant.**

```
'not interned ' * 2  is  'not interned ' * 2   ->  True
BINARY_OP survives compilation                 ->  False
```

The peephole optimiser folds it, so the two evaluations are one object and the identity assertion held for CPython's reasons rather than salix's — which is exactly what the docstring warns about two lines above. `"".join([...])` runs at import; ruff wants it folded back and the noqa matches the one already on the keyword test three functions up, for the same reason.

**7 — the class's own copy was never pinned.** Right: the test checked that each *instance* gets its own and never that the class does, so dropping the class-creation copy — the thing two rounds of this PR were about — would not have failed it. It names both severances now: `stored is not value`, and `Local().field is not stored`.

**4 — the dispatch re-listed the four.** Taken, and the unguarded tail was the real hazard: a fifth type added to `struct_copies_default` would have been handed to `PyByteArray_FromObject`. The bytearray branch is explicit and the fallthrough shares, so the failure mode is a lost per-instance copy rather than a changed type.

**3 — "nothing else can fill" is wrong again.** `__struct_defaults__` hands the stored object out, so something can. That is the route already declined as out of contract, and the comment now points at #51 instead of claiming a wall. Second round running that a comment of mine on these two checks overstated the guarantee.

**6 — the whitelist re-implemented in tests.** Taken in the form that matters: `COPIED_WHEN_EMPTY` stays as the single test-side list, and `test_a_value_may_be_a_default` now derives both branches from it *and* asserts the two-copy contract, so a C-side change that dropped either copy fails there rather than drifting silently.

**1 — the body-`__init__` refusal.** #56, unchanged.

**5 — inherited defaults re-copied.** Declined for the third time, same reasoning: one empty-container allocation per inherited defaulted field at class creation, against making "the class's default is its own" conditional on provenance.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues and answer each finding on its PR. This is round 8 on #49.

JPHutchins:
Three taken, two unchanged. Finding 4 is the same trap one variable along, and worse than the first. `camas check` green at 25/25 on `fd44a23`.

**4 — `10**20` is folded too.** I fixed the string last round and left the int, in a test whose docstring warns about exactly this:

```
10**20 folded  ->  ['RESUME', 'LOAD_CONST', 'STORE_NAME', ...]
int(n) is n    ->  True
(n + 0) is n   ->  False
```

And your second point is the sharper one: the vacuity is not fixable by runtime construction, because `int(n)` returns the argument. No int can ever demonstrate copy-versus-share unless the expression itself survives folding. `10**20 + 0` does, and is not the constant, so the assertion now means something.

**3 — `COPIED_WHEN_EMPTY` was not the single list.** Correct, and my round-8 reply said it was. `test_the_other_mutable_builtins_are_copied_too` still spelled `[dict, set, bytearray]` out; it now says where the list lives. Deriving from the C source is not something pytest can do, so two files naming four types is the floor — what I can stop doing is claiming there is one.

**5 — the claim outlived its own correction.** Right, and this is the third round in a row on this PR. The same commit that removed "nothing outside the class can fill" from the C comment left it standing in a docstring twenty lines away. Both now point at #51 for the `__struct_defaults__` route.

**1 — the body-`__init__` refusal.** #56.

**2 — inherited defaults re-copied.** Declined, fourth time, same reasoning: one empty-container allocation per inherited defaulted field at class creation, against making "the class's default is its own" conditional on provenance.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues and answer each finding on its PR. This is round 9 on #49.

JPHutchins:
**All three taken**, including the one that says my last fix was a no-op. `8ac2e64`, green at 25/25.

**1 — right, and the third time this file has hit the same trap.** `10**20 + 0` folds to the same constant `10**20` does, so the previous commit changed nothing and its docstring's "`+ 0` survives compilation" was false. Measured on 3.14.6:

```console
>>> [i.opname for i in dis.get_instructions(compile("z = 10**20 + 0", "<t>", "exec"))]
['RESUME', 'LOAD_CONST', 'STORE_NAME', 'LOAD_CONST', 'RETURN_VALUE']
>>> (10**20 + 0) is 10**20
True
```

The int is gone rather than respelled a fourth time, because your sharper point is that **no int could ever have worked**:

```console
>>> n = 10**20
>>> int(n) is n, copy.copy(n) is n
(True, True)
```

Both slots hold one object whether salix shares the default or copies everything, so the assertion passes under exactly the regression it existed to catch. `text`, `pair` and `nested` discriminate; the docstring now says why an int cannot instead of claiming a spelling that does.

**3 — taken by making the claim true rather than removing it.** `COPIED_WHEN_EMPTY` moves to `values.py`, which both files already import from, and the parametrization derives from it:

```python
    @pytest.mark.parametrize(
        "factory",
        [kind for kind in COPIED_WHEN_EMPTY if kind is not list],
        ids=lambda factory: factory.__name__,
    )
```

Collection confirms the ids still come out `[dict] [set] [bytearray]`, and adding a type to the tuple now brings a case here with it.

**2 — declined, and your recommendation is what this takes.** No exemption pending #56, and the over-fire is now stated where someone reading the refusal will find it:

> It over-fires deliberately on a class whose body writes its own `__init__`. That class displaces the generated constructor, so its declared default is never handed to an instance and the message's "every instance would share" is about nothing — and the same is true of every subclass below it. #56 is where the whole question of a dead default belongs, and refusing the same shape everywhere is the answer until it is settled.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked that the open PRs keep iterating against the automated reviewer now that its runner problem is fixed; this answers the review that landed on `fd44a23`.

JPHutchins:
**Taken, and it goes one step further than the finding.** `fd03dbf`, green at 25/25.

I dropped the int for being unable to discriminate copy-from-share and then wrote "`text`, `pair` and `nested` already discriminate" in the same commit message. Measured, none of the three do:

| | `str(s)` / `tuple(t)` | full slice | `copy.copy` |
| --- | --- | --- | --- |
| `str` | same object | same object | same object |
| `tuple` | same object | same object | same object |

```console
>>> copy.copy(Inner(1))
TypeError: cannot pickle 'Inner' object          # #13
```

So `nested` cannot fail either, which the finding credits it with. The vacuity was never the int's — it is every immutable's, and the docstring now says that instead of implying a trap that is not there.

Kept rather than dropped, because the test's name is its claim: sharing holds. What it cannot do is catch a copy, and saying so is the fix. The runtime `str.join` stays with its real value stated — not discrimination, but that a salix which *rebuilt* the string rather than storing the declared one would be caught here, where an interned literal would hide it. What pins the copying is the test above, where a mutable default is a distinct object per instance and the assertion can fail.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked that the open PRs keep iterating against the automated reviewer; this answers the review that landed on `8ac2e64`.

JPHutchins:
**Four taken, two answered with measurements.** `68e1922`, green at 25/25.

**2 — you are right and I can show it.** "a set, whose copy hashes every element" was simply false:

```console
set(s) called __hash__: 0 times
copy.copy(s) called __hash__: 0 times
```

`PySet_New` copies the table. The real reason for the first check is the cheap one — an O(1) pre-filter that keeps a default which is about to be refused from being built into a copy and thrown away — and that is what the comment says now, with the correction left in place so the next reader does not re-derive the wrong rationale.

**1 — taken, and the header comment was the worse half.** The guard is gone and the final `return Py_NewRef` is the live else-branch. But "Stated once so the two cannot drift" was never true and could not be: a predicate cannot supply a per-type constructor, so the four are written twice by construction. The header now says that, says nothing makes them agree, and says which direction degrades safely — a type in the predicate but not the copier is shared; the other way round changes a value's type.

**6 and 5 — taken.** Type and value are asserted alongside the distinctness, so a copier answering with a distinct object of the wrong type no longer passes. `test_every_non_empty_builtin_is_refused` derives from `COPIED_WHEN_EMPTY` like its sibling, with the seeds in `values.py` because no one iterable constructs all four.

Two assertions guard that mapping, and the second earned its keep on the first run: my seed for `bytearray` was `b"x"`, which is a **bytes** and so is not refused at all. The test failed; the assertion now says why in one line.

<details>
<summary><b>4 — measured rather than added to that table</b></summary>

Fair that the copy path was unmeasured. Here it is, 200k iterations, min of 9:

| | ns/construction |
| --- | --- |
| 3 positional, no defaults filled | 52.7 |
| 1 positional, 2 immutable defaults filled | 51.9 |
| **1 positional, 2 mutable defaults copied** | **76.0** |
| 3 positional on that same class | 72.7 |

So a copied default costs about **12 ns each**, and the class carrying them costs ~20 ns more per construction whether or not the defaults are used — that is the wider layout, not the copy.

I did not put it in `bench/bench.py` because that table's value is that all five rivals do the identical thing in every row, and a defaulted-field row would either change the shared 3-argument signature or be salix-only in a comparison table. Happy to add it as a separate section if you want it tracked rather than reported.

</details>

<details>
<summary><b>3 — unchanged, and the cost is on the wrong path</b></summary>

Re-running `reject_unsafe_default` in `fill_defaults` is an O(1) size check per defaulted field **per construction** — the numbers above are 52–76 ns, so it is real there in a way it is not at class creation. It would also turn a program that mutated `__struct_defaults__` into one that raises at a distance, on a route the comment already calls out of contract.

The read-only-proxy half of your recommendation is the better shape, but it is a question about what `__struct_defaults__` *is*, which is #51's and #20's rather than this PR's. Recorded there rather than decided here.

</details>

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked that the open PRs keep iterating against the automated reviewer; this answers the review that landed on `fd03dbf`.

JPHutchins:
**All three taken.** `d984841`, green at 25/25.

## 3 — the value set had no empty set and no empty bytearray

Right, and it is the one worth showing rather than agreeing with. I built the regression the finding describes — class creation stores the declared set and bytearray directly, construction still copies — and ran it against both value sets:

```
mutation + value set at 68e1922    645 passed, 8 skipped
mutation + value set as of d984841 FAILED test_a_value_may_be_a_default[set-3911]
                                   FAILED test_a_value_may_be_a_default[bytearray-2290]
```

So the hole was real and the two new values close it. `list` and `dict` each had an empty instance *and* a non-empty one; `set` and `bytearray` had only the non-empty one, which is the whole of the asymmetry.

Rather than parametrize a second test, the fix is upstream of every test that reads the value set — `set()` and `bytearray()` join `[]` and `{}` in `UNHASHABLE`, so `stored is not value` runs for all four. Next to them is what stops it reopening:

```python
assert all(
    {not value for value in UNHASHABLE if type(value) is kind} == {True, False}
    for kind in COPIED_WHEN_EMPTY
)
```

<details>
<summary>Why an assertion and not a comment</summary>

Both halves of the rule — copy when empty, refuse when not — are reached by parametrizing over these values, so the value set has to hold one of each for all four types. That is a property of the data, and this file already keeps `NON_EMPTY` honest the same way. A fourth type added to `COPIED_WHEN_EMPTY` now fails collection until the value set carries both of its instances.
</details>

## 1 — deleted, not reworded

You are right that the assertion is vacuous, and I could not find a rewrite worth keeping. The ordering the test is named for is unobservable through a set: neither `PySet_New` nor the refusal's `PyObject_Size` calls an element's `__hash__`, so no user code runs on either side of the decision. Nothing the test can reach distinguishes the two orders.

What it *also* asserted — that the message names the type, `"non-empty set"` — was worth keeping, so it moved to where it generalises:

```python
with pytest.raises(TypeError, match=f"non-empty {factory.__name__}"):
```

That derives the expected name from the same parameter the value comes from, so it now pins the message for `dict` and `bytearray` too, which the deleted test never did.

## 2 — reworded

Taken. For `xs: list = [[1]]` the first check is what raises, so "it decides nothing" was false. It now says its verdict is not final, and why: the module can still write to the object between the two reads.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, answering round 12's findings on this PR.

JPHutchins:
**Five taken, three answered.** `0114179`, green at 25/25.

## 3 and 4 — one table, so the drift is unreachable

The header admitted the defect it had: *"The two lists have to agree and nothing makes them."* Now one list answers both questions:

```c
static struct default_copy const COPIED_WHEN_EMPTY[] = {
    {.kind = &PyList_Type, .copy = copy_list},
    {.kind = &PyDict_Type, .copy = PyDict_Copy},
    {.kind = &PySet_Type, .copy = PySet_New},
    {.kind = &PyByteArray_Type, .copy = PyByteArray_FromObject},
};
```

`struct_copies_default` is "is there a row" and `struct_default_copy` is "call the row's constructor", so a type cannot be refused without being copied or copied without being refused. That takes finding 4 with it: "a type added here and not there is shared, which is the direction that degrades safely" was exactly backwards — sharing a mutable default is the bug this PR fixes — and the sentence is gone rather than corrected, because the case it described can no longer arise.

`PyList_GetSlice` takes bounds where the other three take the object, so it has a two-line wrapper; that is the whole cost of the uniform signature.

## 2 — the message asserted a sharing that cannot happen

Right, and measured on this head:

```
body __init__ class:    __post_init__ ran? False
generated constructor:  __post_init__ ran? True
```

For a class whose body writes `__init__`, the declared default is never handed to an instance, so "which every instance would share the contents of" was about nothing — and `__post_init__` never runs there, because the constructor that calls it is the one the body displaced.

The message states the rule rather than a consequence now, and names which remedy that class is left with:

```
field 'xs' defaults to a non-empty list, whose copy could only be shallow and
would leave the contents shared; default it to an empty one and fill it with
set_field -- from __post_init__, or from your own __init__ if the body writes
one, which displaces the constructor __post_init__ runs from
```

## 1 — the subclass boundary is pinned now

Taken as recommended. A non-empty subclass of one of the four is shared and not refused, which is the price of copying by exact type (`PyDict_Copy` of a `defaultdict` is a `dict`). It was documented and untested; it is now four more parameters on `test_a_mutable_container_outside_the_four_is_shared_and_not_refused`, beside `array`, `deque`, `memoryview` and a mutable struct — derived from `COPIED_WHEN_EMPTY`, so a fifth type brings its subclass case with it.

Refusing them instead is a real option and it is #51's, where the whole type list is argued.

## 7 and 8 — both taken

`values.py` said how the assertion was arrived at rather than what it enforces; it states the invariant now. And `test_an_immutable_default_is_shared`'s docstring claimed all three assertions were non-discriminating — two are, the nested struct is not, and for the reason the docstring itself gave: `copy.copy(Inner(1))` raises (#13), so a salix that copied every default could not get past that field.

## 5 — declined, and the table is why

`filled < 0` is unreachable today. It is also the fail-closed branch, and the table above makes the reachable set editable in one place: a fifth row whose `sq_length` can fail turns the dead branch live, and the alternative (`if (filled > 0)`) would silently read a failed probe as "nothing to refuse". One comparison for a guard that survives the table growing.

## 6 — half declined, half routed

The CHANGELOG half does not apply: nothing has been released, and the file says so — *"Nothing released yet; everything in `main` is the first release."* There is no version for a reader to be surprised by a change from.

The documentation half is real and is #18's, which is open for exactly this: the README becomes the PyPI landing page. Three facts belong in it, and I have added them there rather than to a CHANGELOG entry nobody will read:

- the four exact builtins are copied per instance when empty;
- a non-empty one is refused at class creation;
- `__struct_defaults__` holds the class's copy, not the object the body named.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, answering round 13's findings on her PR.

JPHutchins:
**The Windows break is real and is fixed; the other two were already answered in the commit you reviewed.** `e6d4505`, `nix flake check` green in 256s.

## 1 — right, and it is mine

`0114179`'s table was a file-scope static initializer holding `&PyList_Type` and its three siblings. On Windows those are `__declspec(dllimport)` and their addresses are not address constants, so clang refused it:

```
src/construct.c:251:11: error: initializer element is not a compile-time constant
error: Cannot build '...salix-wheel-3.10-windows-x86_64-0.0.0.drv'
```

**Why it got past me is worth writing down:** the Windows cross-build lives in `flake_check`, and `flake_check` is in the `ci` task, not in `check`. `camas run check` — the gate I run, 25 leaves, 6 seconds — cannot compile for Windows at all. I have run `flake_check` before this push and it is green in 256s, and I will run it on every C change from here.

The fix keeps the single-list property, which was the point of the refactor, without the static initializer:

```c
static default_copier copies_default(PyTypeObject const * const kind) {
    if (kind == &PyList_Type) { return copy_list; }
    if (kind == &PyDict_Type) { return PyDict_Copy; }
    ...
```

I did not take the suggested runtime-populated static table: a mutable static needs a story for two threads reaching it at once on a free-threaded build, and this needs none — the four comparisons happen on the caller's stack. One function still names the four types once, so the refusal and the copy cannot disagree.

## 2 and 3 — both landed in the commit under review

These read as carried forward rather than re-checked; both were fixed in `0114179` itself.

**3, the message.** Its current text has no "every instance would share":

```c
"field '%U' defaults to a non-empty %.100s, whose copy could only be "
"shallow and would leave the contents shared; default it to an empty "
"one and fill it with set_field -- from __post_init__, or from your own "
"__init__ if the body writes one, which displaces the constructor "
"__post_init__ runs from"
```

It states the rule rather than a consequence, and names which remedy a body-`__init__` class is left with — measured on that head: `__post_init__ ran? False` there, `True` for a generated constructor.

**2, the subclass boundary.** `0114179` added four parameters to `test_a_mutable_container_outside_the_four_is_shared_and_not_refused`, one per type, each a **non-empty** subclass:

```python
*(
    pytest.param(subclass_of(kind)(NON_EMPTY[kind]), id=f"a_{kind.__name__}_subclass")
    for kind in COPIED_WHEN_EMPTY
),
```

**And a correction to my own last reply while I am here.** I wrote that the boundary "was documented and untested". Not so: `test_a_subclass_of_a_mutable_builtin_is_shared_not_copied` has pinned the *empty* subclass case (`defaultdict(list)`) since before this round. What was missing, and what those four parameters add, is the **non-empty** one — which is the half your finding is actually about.

Refusing non-empty subclasses instead of sharing them remains open, and it is #51's, where the whole type list is argued.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, fixing the Windows build break the review caught and correcting one of her own earlier claims.

PR #50, opened 2026-08-01
fix: refuse an annotation that names something other than a field

### LINKED ISSUE #14: ClassVar annotations become fields, so the runtime and the stub disagree

JPHutchins:
Closes #14. Every annotation became a field, `ClassVar` included, so the runtime and the shipped stub disagreed about the constructor:

```python
class Q(Struct):
    registry: ClassVar[int] = 0
    x: int = 1

Q(5).registry        # 5 — the first positional argument landed in the class variable
```

A checker reading `salix/__init__.pyi` synthesises `Q(x: int = 1)`, because `dataclass_transform` excludes `ClassVar`. Checked code and running code disagreed about what the first argument meant.

The commoner shape was worse. A `ClassVar` is usually declared first and usually has a value, so `instances: ClassVar[list] = []` followed by a required field was rejected outright — for following a default, with a message that never mentioned `ClassVar`. That is the first thing a dataclass port hits.

Both forms are refused now, each saying what to do instead:

```
'instances' is annotated ClassVar, which salix does not support; write it below the fields, without an annotation
'seed' is annotated InitVar, which salix does not support; take the value in a custom __init__ and write the fields with set_field
```

<details>
<summary><b>Nothing is imported to find out</b></summary>

A module nobody imported cannot have supplied the annotation, so `PyImport_GetModule` answers for free when `typing` or `dataclasses` is absent — and a bare interpreter has neither loaded:

```
typing imported in a bare interpreter: False
```

That is the trick `dataclasses` uses, for the same reason. When the module *is* loaded, the test is identity against `ClassVar`, against its `__origin__` for the subscripted form, and against `InitVar` as a type. No string matching on the fast path, and no import added to class creation.

**Only the object form is recognised.** Under `from __future__ import annotations` the annotation is source text, and matching that means guessing at import aliases. `dataclasses` guesses; this does not, and the comment says so.

</details>

### Deferred, not closed

@JPHutchins asked for an opt-in per definition, so a class can pay for annotation inspection when it wants the feature. Rejecting is what makes the runtime agree with the checker until that exists — and it is the reversible direction, since supporting these later is not a breaking change.

Six tests in `tests/test_special_forms.py`, including that a field *named* `ClassVar` is still a field: the annotation is what is inspected, never the name. `camas check` green at 25/25.

> [!WARNING]
> **LLM Disclosure**
>
> This PR was authored by claude-opus-5 on behalf of @JPHutchins, implementing the ruling she gave on #14 and #6 — reject now, with an opt-in keyword to follow.

### DIFF
--- a/src/fields.c
+++ b/src/fields.c
@@ -42,7 +42,7 @@ static struct special_form special_form_of(
 	PyObject * init_var
 );
 static struct special_form named_special_form(PyObject * text);
-static bool names_form(char const * source, char const * form);
+static bool names_form(char const * source, Py_ssize_t length, char const * form);
 static PyObject * module_attribute(char const * module_name, char const * attribute);
 
 /* The plan only takes references once every step has succeeded; the working
@@ -143,6 +143,12 @@ static enum result append_declared(
 	PyObject * field_name;
 	PyObject * annotation;
 	Py_ssize_t position = 0;
+
+	/* A class with no annotations of its own declares no fields and so can name
+	 * no forms; the two probes below are the whole cost of asking. */
+	if (PyDict_GET_SIZE(annotations) == 0) {
+		return RESULT_OK;
+	}
 
 	/* Once per class, not once per field. Absent means the module was never
 	 * imported, so nothing in this body can be naming what it holds -- but
@@ -251,6 +257,17 @@ enum : int {
 	SPECIAL_FORM_HOPS = 4,
 };
 
+/* Both paths answer the same question and owe the user the same sentence, so
+ * the answers live here rather than once per path. */
+static struct special_form const CLASS_VAR_FORM = {
+	.name = "ClassVar",
+	.instead = "write it below the fields, without an annotation",
+};
+static struct special_form const INIT_VAR_FORM = {
+	.name = "InitVar",
+	.instead = "take the value in a custom __init__ and write the fields with set_field",
+};
+
 static struct special_form special_form_of(
 	PyObject * const annotation,
 	PyObject * const class_var,
@@ -266,20 +283,14 @@ static struct special_form special_form_of(
 
 	for (int hop = 0; hop < SPECIAL_FORM_HOPS; ++hop) {
 		if (class_var != NULL && current == class_var) {
-			return (struct special_form){
-				.name = "ClassVar",
-				.instead = "write it below the fields, without an annotation",
-			};
+			return CLASS_VAR_FORM;
 		}
 
 		if (
 			init_var != NULL &&
 			(current == init_var || (PyObject *) Py_TYPE(current) == init_var)
 		) {
-			return (struct special_form){
-				.name = "InitVar",
-				.instead = "take the value in a custom __init__ and write the fields with set_field",
-			};
+			return INIT_VAR_FORM;
 		}
 
 		/* A plain class is the common annotation and has no __origin__; asking
@@ -313,7 +324,8 @@ static struct special_form special_form_of(
  * does not.
  */
 static struct special_form named_special_form(PyObject * const text) {
-	char const * const source = PyUnicode_AsUTF8(text);
+	Py_ssize_t length = 0;
+	char const * const source = PyUnicode_AsUTF8AndSize(text, &length);
 
 	if (source == NULL) {
 		PyErr_Clear();
@@ -321,18 +333,12 @@ static struct special_form named_special_form(PyObject * const text) {
 		return (struct special_form){0};
 	}
 
-	if (names_form(source, "ClassVar")) {
-		return (struct special_form){
-			.name = "ClassVar",
-			.instead = "write it below the fields, without an annotation",
-		};
+	if (names_form(source, length, "ClassVar")) {
+		return CLASS_VAR_FORM;
 	}
 
-	if (names_form(source, "InitVar")) {
-		return (struct special_form){
-			.name = "InitVar",
-			.instead = "take the value in a custom __init__ and write the fields with set_field",
-		};
+	if (names_form(source, length, "InitVar")) {
+		return INIT_VAR_FORM;
 	}
 
 	return (struct special_form){0};
@@ -348,33 +354,53 @@ static struct special_form named_special_form(PyObject * const text) {
  * `ClassVar ` and `ClassVar [int]`, both legal and both stored verbatim under
  * future annotations, walked past an earlier version of this.
  *
+ * A byte at or above 0x80 counts as an identifier character. PEP 3131 lets an
+ * identifier hold any Unicode letter, and this reads UTF-8, so the lead byte of
+ * `é` is the middle of a name rather than the end of one -- otherwise
+ * `théClassVar` reads as the form standing on its own.
+ *
  * A heuristic in both directions, and the only thing available once the
  * annotation is source text. It misses a renamed import -- `ClassVar as CV`
  * gives `CV[int]` -- and it refuses a user's own type that happens to be called
  * ClassVar, and `Annotated[int, ClassVar]` where the form is metadata rather
- * than the type. #57 keeps the list. The object path is exact, and it is what
+ * than the type -- including when it is quoted, since a quote is a boundary.
+ * That last one is deliberate: `x: 'ClassVar[int]'` is a nested forward
+ * reference and has to be refused, and nothing short of parsing tells the two
+ * apart. #57 keeps the list. The object path is exact, and it is what
  * runs unless the module asked for `from __future__ import annotations`.
  */
-static bool identifier_character(char const character) {
+static bool identifier_character(unsigned char const byte) {
 	return (
-		(character >= 'a' && character <= 'z') ||
-		(character >= 'A' && character <= 'Z') ||
-		(character >= '0' && character <= '9') ||
-		character == '_'
+		(byte >= 'a' && byte <= 'z') ||
+		(byte >= 'A' && byte <= 'Z') ||
+		(byte >= '0' && byte <= '9') ||
+		byte == '_' ||
+		byte >= 0x80
 	);
 }
 
-static bool names_form(char const * const source, char const * const form) {
-	size_t const length = strlen(form);
+static bool names_form(
+	char const * const source,
+	Py_ssize_t const source_length,
+	char const * const form
+) {
+	Py_ssize_t const form_length = (Py_ssize_t) strlen(form);
 
-	for (
-		char const * found = strstr(source, form);
-		found != NULL;
-		found = strstr(found + 1, form)
-	) {
-		bool const opens = found == source || !identifier_character(found[-1]);
+	/* Bounded by the string's own length rather than its first NUL. A str is
+	 * allowed to contain one, and strstr would stop there -- leaving whatever
+	 * follows unscanned and the guard silently off. */
+	for (Py_ssize_t at = 0; at + form_length <= source_length; ++at) {
+		if (memcmp(source + at, form, (size_t) form_length) != 0) {
+			continue;
+		}
 
-		if (opens && !identifier_character(found[length])) {
+		bool const opens = at == 0 || !identifier_character(source[at - 1]);
+		bool const closes = (
+			at + form_length == source_length ||
+			!identifier_character(source[at + form_length])
+		);
+
+		if (opens && closes) {
 			return true;
 		}
 	}
--- a/src/fields.c
+++ b/src/fields.c
@@ -13,6 +13,15 @@ struct special_form {
 	char const * instead;
 };
 
+/* What the two paths match an annotation against. Built once per class, because
+ * the text path's needles are the same two words for every field in it. */
+struct form_probes {
+	PyObject * class_var;
+	PyObject * init_var;
+	PyObject * class_var_name;
+	PyObject * init_var_name;
+};
+
 enum inheritance : int {
 	INHERITANCE_ERROR = -1,
 	INHERITANCE_NEW = 0,
@@ -32,16 +41,28 @@ static enum result append_declared(
 	PyObject * new_names,
 	PyObject * default_by_name
 );
+
+/* Both paths answer the same question and owe the user the same sentence, so
+ * the answers live here rather than once per path -- and the text path's
+ * needles are built from these names, so each form is spelled once. */
+static struct special_form const CLASS_VAR_FORM = {
+	.name = "ClassVar",
+	.instead = "write it below the fields, without an annotation",
+};
+static struct special_form const INIT_VAR_FORM = {
+	.name = "InitVar",
+	.instead = "take the value in a custom __init__ and write the fields with set_field",
+};
 static PyObject * build_defaults(PyObject * all_names, PyObject * default_by_name);
 static PyObject * checked_annotations(PyObject * namespace);
 static enum inheritance inherits_field(StructType const * base, PyObject * field_name);
 static struct special_form special_form_of(
 	PyObject * annotation,
-	PyObject * class_var,
-	PyObject * init_var
+	struct form_probes const * probes
 );
-static struct special_form named_special_form(PyObject * text);
-static bool names_form(PyObject * text, char const * form);
+static struct special_form named_special_form(PyObject * text, struct form_probes const * probes);
+static bool annotates_with_text(PyObject * annotations);
+static bool names_form(PyObject * text, PyObject * needle);
 static bool continues_identifier(Py_UCS4 character);
 static PyObject * module_attribute(char const * module_name, char const * attribute);
 
@@ -145,7 +166,7 @@ static enum result append_declared(
 	Py_ssize_t position = 0;
 
 	/* A class with no annotations of its own declares no fields and so can name
-	 * no forms; the two probes below are the whole cost of asking. */
+	 * no forms; the four probes below are the whole cost of asking. */
 	if (PyDict_GET_SIZE(annotations) == 0) {
 		return RESULT_OK;
 	}
@@ -166,6 +187,29 @@ static enum result append_declared(
 	if (init_var == NULL && PyErr_Occurred()) {
 		return RESULT_ERROR;
 	}
+
+	/* From the same two strings the refusal quotes, so there is one spelling of
+	 * each form in the file. Built here and not in the matcher, which runs per
+	 * field and per form -- and only for a class that has text to match, since
+	 * building them for every class taxes the ones that never look. */
+	PY_MOVABLE(class_var_name, NULL);
+	PY_MOVABLE(init_var_name, NULL);
+
+	if (annotates_with_text(annotations)) {
+		class_var_name = PyUnicode_FromString(CLASS_VAR_FORM.name);
+		init_var_name = PyUnicode_FromString(INIT_VAR_FORM.name);
+
+		if (class_var_name == NULL || init_var_name == NULL) {
+			return RESULT_ERROR;
+		}
+	}
+
+	struct form_probes const probes = {
+		.class_var = class_var,
+		.init_var = init_var,
+		.class_var_name = class_var_name,
+		.init_var_name = init_var_name,
+	};
 
 	while (PyDict_Next(annotations, &position, &field_name, &annotation)) {
 		if (!PyUnicode_CheckExact(field_name)) {
@@ -197,7 +241,7 @@ static enum result append_declared(
 
 		/* After the inheritance check, so re-annotating an inherited field is
 		 * the no-op it has always been rather than a new refusal. */
-		struct special_form const special = special_form_of(annotation, class_var, init_var);
+		struct special_form const special = special_form_of(annotation, &probes);
 
 		/* Before the answer is used at all, not only when it is "no form": the
 		 * text path allocates on the way to either verdict, and a failure there
@@ -264,25 +308,16 @@ enum : int {
 	SPECIAL_FORM_HOPS = 4,
 };
 
-/* Both paths answer the same question and owe the user the same sentence, so
- * the answers live here rather than once per path. */
-static struct special_form const CLASS_VAR_FORM = {
-	.name = "ClassVar",
-	.instead = "write it below the fields, without an annotation",
-};
-static struct special_form const INIT_VAR_FORM = {
-	.name = "InitVar",
-	.instead = "take the value in a custom __init__ and write the fields with set_field",
-};
-
 static struct special_form special_form_of(
 	PyObject * const annotation,
-	PyObject * const class_var,
-	PyObject * const init_var
+	struct form_probes const * const probes
 ) {
 	if (PyUnicode_Check(annotation)) {
-		return named_special_form(annotation);
+		return named_special_form(annotation, probes);
 	}
+
+	PyObject * const class_var = probes->class_var;
+	PyObject * const init_var = probes->init_var;
 
 	/* Neither module is loaded, so nothing reachable from here can be either
 	 * form and the walk below can only ask __origin__ of things for nothing. */
@@ -344,12 +379,32 @@ static struct special_form special_form_of(
  * before any of this. dataclasses guesses at those against sys.modules; this
  * does not.
  */
-static struct special_form named_special_form(PyObject * const text) {
-	if (names_form(text, "ClassVar")) {
+/* The same question the matcher asks per field, asked once for the class: is
+ * there any text here at all? Under `from __future__ import annotations` every
+ * annotation is, and otherwise almost none are. */
+static bool annotates_with_text(PyObject * const annotations) {
+	PyObject * field_name;
+	PyObject * annotation;
+	Py_ssize_t position = 0;
+
+	while (PyDict_Next(annotations, &position, &field_name, &annotation)) {
+		if (PyUnicode_Check(annotation)) {
+			return true;
+		}
+	}
+
+	return false;
+}
+
+static struct special_form named_special_form(
+	PyObject * const text,
+	struct form_probes const * const probes
+) {
+	if (names_form(text, probes->class_var_name)) {
 		return CLASS_VAR_FORM;
 	}
 
-	if (names_form(text, "InitVar")) {
+	if (names_form(text, probes->init_var_name)) {
 		return INIT_VAR_FORM;
 	}
 
@@ -405,13 +460,7 @@ static bool continues_identifier(Py_UCS4 const character) {
 	return PyUnicode_IsIdentifier(probe) == 1;
 }
 
-static bool names_form(PyObject * const text, char const * const form) {
-	PY_OWNED(needle, PyUnicode_FromString(form));
-
-	if (needle == NULL) {
-		return false;
-	}
-
+static bool names_form(PyObject * const text, PyObject * const needle) {
 	Py_ssize_t const length = PyUnicode_GET_LENGTH(text);
 	Py_ssize_t const form_length = PyUnicode_GET_LENGTH(needle);
 
--- a/src/fields.c
+++ b/src/fields.c
@@ -60,6 +60,8 @@ static struct special_form special_form_of(
 	PyObject * annotation,
 	struct form_probes const * probes
 );
+static struct special_form form_within(PyObject * annotation, struct form_probes const * probes);
+static PyObject * optional_attribute(PyObject * object, char const * name);
 static struct special_form named_special_form(PyObject * text, struct form_probes const * probes);
 static bool annotates_with_text(PyObject * annotations);
 static bool names_form(PyObject * text, PyObject * needle);
@@ -305,8 +307,58 @@ static enum inheritance inherits_field(StructType const * const base, PyObject *
  * cannot spin here.
  */
 enum : int {
-	SPECIAL_FORM_HOPS = 4,
+	/* A budget on work rather than on depth. The walk is a tree since #57's
+	 * ruling, and bounding the shape no longer bounds the effort: four hops was
+	 * enough for a chain -- Optional[Annotated[ClassVar[int], 'm']] is four --
+	 * and this is enough for the arguments beside them, while a
+	 * self-referential __origin__ still costs a fixed amount. */
+	SPECIAL_FORM_NODES = 32,
 };
+
+/*
+ * Owned, every one of them. __origin__ hands back a new reference, and an
+ * argument outlives the tuple it was read from only because this holds it.
+ *
+ * A fixed array rather than a list, because the frontier is bounded anyway and
+ * an allocation here would be one per subscripted annotation per class -- the
+ * cost the needles were just moved out of the matcher to avoid.
+ */
+struct form_frontier {
+	PyObject * nodes[SPECIAL_FORM_NODES];
+	int count;
+};
+
+static void form_frontier_clear(struct form_frontier * const frontier) {
+	while (frontier->count > 0) {
+		Py_DECREF(frontier->nodes[--frontier->count]);
+	}
+}
+
+/* Full is not a failure: the budget is the point, and a walk that runs out
+ * answers "no form", which is what an unrecognised annotation answers. */
+static void form_frontier_push(struct form_frontier * const frontier, PyObject * const node) {
+	if (frontier->count < SPECIAL_FORM_NODES) {
+		frontier->nodes[frontier->count++] = Py_NewRef(node);
+	}
+}
+
+static struct special_form form_named_by(
+	PyObject * const annotation,
+	struct form_probes const * const probes
+) {
+	if (probes->class_var != NULL && annotation == probes->class_var) {
+		return CLASS_VAR_FORM;
+	}
+
+	if (
+		probes->init_var != NULL &&
+		(annotation == probes->init_var || (PyObject *) Py_TYPE(annotation) == probes->init_var)
+	) {
+		return INIT_VAR_FORM;
+	}
+
+	return (struct special_form){0};
+}
 
 static struct special_form special_form_of(
 	PyObject * const annotation,
@@ -316,58 +368,93 @@ static struct special_form special_form_of(
 		return named_special_form(annotation, probes);
 	}
 
-	PyObject * const class_var = probes->class_var;
-	PyObject * const init_var = probes->init_var;
-
 	/* Neither module is loaded, so nothing reachable from here can be either
-	 * form and the walk below can only ask __origin__ of things for nothing. */
-	if (class_var == NULL && init_var == NULL) {
+	 * form and the walk below can only ask attributes of things for nothing. */
+	if (probes->class_var == NULL && probes->init_var == NULL) {
 		return (struct special_form){0};
 	}
 
-	PyObject * current = annotation;
+	return form_within(annotation, probes);
+}
 
-	PY_MOVABLE(held, NULL);
+/*
+ * The form can be anywhere in a subscripted annotation, not only at the end of
+ * the __origin__ chain. `Optional[Annotated[ClassVar[int], 'm']]` keeps it in
+ * the arguments, where a chain walk never looks, so it became a field while
+ * the text path refused the same source -- #14 again, on the path almost every
+ * class takes. Both are walked now, per #57's ruling.
+ *
+ * It also makes the object path refuse `Annotated[int, ClassVar]`, where the
+ * form is metadata rather than the type. That is not a new refusal: the text
+ * path already refuses it and ships that way, and one answer is worth more
+ * than two that disagree.
+ *
+ * A str inside __args__ is walked, not matched -- `Annotated[int, "ClassVar"]`
+ * is a string in a metadata slot, and the text matcher is for an annotation
+ * that reached salix as source, not for anything that looks like one.
+ *
+ * A queue rather than a recursion, because clang-tidy's misc-no-recursion is an
+ * error here and the shape does not need one: the frontier is what bounds the
+ * walk, so neither a self-referential __origin__ nor an __args__ holding its
+ * own owner can spin.
+ */
+static struct special_form form_within(
+	PyObject * const annotation,
+	struct form_probes const * const probes
+) {
+	__attribute__((cleanup(form_frontier_clear))) struct form_frontier frontier = {0};
 
-	for (int hop = 0; hop < SPECIAL_FORM_HOPS; ++hop) {
-		if (class_var != NULL && current == class_var) {
-			return CLASS_VAR_FORM;
+	form_frontier_push(&frontier, annotation);
+
+	for (int at = 0; at < frontier.count; ++at) {
+		PyObject * const current = frontier.nodes[at];
+		struct special_form const named = form_named_by(current, probes);
+
+		if (named.name != NULL) {
+			return named;
 		}
 
-		if (
-			init_var != NULL &&
-			(current == init_var || (PyObject *) Py_TYPE(current) == init_var)
-		) {
-			return INIT_VAR_FORM;
-		}
-
-		/* A plain class is the common annotation and has no __origin__; asking
-		 * anyway costs an AttributeError raised and cleared per field. */
+		/* A plain class is the common annotation and has neither attribute;
+		 * asking anyway costs two AttributeErrors raised and cleared. */
 		if (PyType_Check(current)) {
-			break;
+			continue;
 		}
 
-		PyObject * const next = PyObject_GetAttrString(current, "__origin__");
+		PY_OWNED(origin, optional_attribute(current, "__origin__"));
 
-		if (next == NULL) {
-			/* Not having one is the ordinary answer and ends the walk. Anything
-			 * else is a failure to look, and the caller checks for it -- the
-			 * last place in this file that used to clear what it did not
-			 * expect. */
-			if (!PyErr_ExceptionMatches(PyExc_AttributeError)) {
-				break;
-			}
-
-			PyErr_Clear();
-
-			break;
+		if (origin != NULL) {
+			form_frontier_push(&frontier, origin);
 		}
 
-		Py_XSETREF(held, next);
-		current = held;
+		PY_OWNED(arguments, PyErr_Occurred() ? NULL : optional_attribute(current, "__args__"));
+
+		if (PyErr_Occurred()) {
+			return (struct special_form){0};
+		}
+
+		for (
+			Py_ssize_t i = 0;
+			arguments != NULL && PyTuple_Check(arguments) && i < PyTuple_GET_SIZE(arguments);
+			++i
+		) {
+			form_frontier_push(&frontier, PyTuple_GET_ITEM(arguments, i));
+		}
 	}
 
 	return (struct special_form){0};
+}
+
+/* The attribute if it is there, and NULL if it is not. NULL with an exception
+ * set is a failure to look, which the caller propagates rather than clears --
+ * a guard that cannot see is not a guard that says "ordinary field". */
+static PyObject * optional_attribute(PyObject * const object, char const * const name) {
+	PyObject * const value = PyObject_GetAttrString(object, name);
+
+	if (value == NULL && PyErr_ExceptionMatches(PyExc_AttributeError)) {
+		PyErr_Clear();
+	}
+
+	return value;
 }
 
 /*
--- a/src/fields.c
+++ b/src/fields.c
@@ -14,13 +14,22 @@ struct special_form {
 };
 
 /* What the two paths match an annotation against. Built once per class, because
- * the text path's needles are the same two words for every field in it. */
+ * the text path's needles are the same two words for every field in it. Owning
+ * what it holds, so that a needle is written in one place rather than into a
+ * local and then into here. */
 struct form_probes {
 	PyObject * class_var;
 	PyObject * init_var;
 	PyObject * class_var_name;
 	PyObject * init_var_name;
 };
+
+static void form_probes_clear(struct form_probes * const probes) {
+	Py_CLEAR(probes->class_var);
+	Py_CLEAR(probes->init_var);
+	Py_CLEAR(probes->class_var_name);
+	Py_CLEAR(probes->init_var_name);
+}
 
 enum inheritance : int {
 	INHERITANCE_ERROR = -1,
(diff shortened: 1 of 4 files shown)


JPHutchins:
All taken. Four were real escapes or a real cost, and the round found more than the original issue did. `camas check` green at 25/25.

**2 — bare `InitVar` became a field.** Right: `Py_TYPE(annotation) == init_var` only matches `InitVar[T]`, and unsubscripted it is the class itself. Both spellings now, matching the pair `ClassVar` already had.

**3 — `Annotated[...]` hid the form.** Right, and it is the same position-swallowing bug wearing a wrapper. The chain reaches it two hops down:

```
Annotated[ClassVar[int], 'x'] -> ClassVar[int] -> ClassVar
```

So `__origin__` is walked rather than probed once, bounded so a self-referential chain cannot spin.

**1 — the string form.** Taken. `from __future__ import annotations` left the whole style unguarded, which is #14 reproducing. A module alias comes along for free — `t.ClassVar[int]` still ends in the form after a dot — but a *renamed* import does not:

```
ClassVar[int]              REFUSED      MyClassVarThing   field
typing.ClassVar[int]       REFUSED      ClassVarish       field
t.ClassVar[int]            REFUSED      CV[int]           field  (from typing import ClassVar as CV)
```

My previous comment had this backwards — it claimed `t.ClassVar[int]` was missed. Corrected in the source.

**5 — the cost.** Right, and worse than reported: 2.4×, not double. Both causes were mine — module lookups per field, and an `__origin__` probe raising and clearing an `AttributeError` for every plain annotation. Hoisted to once per class and skipped for a plain type, which is what nearly every annotation is:

```
main, no guard      12.18 µs
per-field lookups   29.60 µs      ← what you measured
this commit         12.39 µs      ← 1.7% over main
```

**4 — the guard ran before the inheritance check.** Correct, and a behaviour change the issue never asked for: re-annotating an inherited field became a new refusal where it had always been a silent no-op. Moved after it. (Whether *that* silence is right is a separate question, and not one #14 raises.)

**6 and 7 — both fair.** `borrow_attribute` returned a borrowed reference out of a `PY_OWNED` scope, which `owned.h` names as the one mistake the cleanup attribute cannot catch; it hands back a new reference now. And `PyImport_GetModule`'s NULL path left a real error set for the caller to trip over.

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Findings 1 and 2 are the same dial turned opposite ways, and 1 wins. `camas check` green at 25/25.

**1 — `Annotated[ClassVar[int], ...]` escapes the text path.** Confirmed, and the comment I wrote last round made it worse by claiming the wrapper was handled — true of the object path, false of the text one, where `ClassVar` sits after a bracket rather than a start or a dot.

The matcher now accepts the form standing on its own anywhere in the text. Both directions:

```
refused                          still fields
Annotated[ClassVar[int], 'x']    int
Annotated[InitVar[int], 'x']     MyClassVarThing
ClassVar[int]   ClassVar         ClassVarish
typing.ClassVar[int]             list[int]   dict[str, int]
t.ClassVar[int]                  Annotated[int, "x"]
```

**2 — it falsely refuses a user type named `ClassVar`.** Also correct, and irreducible: once the annotation is not evaluated, the text is all there is. Widening to catch the wrapper and narrowing to spare that type are the same dial, so they cannot both be had.

I turned it toward refusing. The wrapper is the shape #14 is actually about; the collision needs someone to have named a type `ClassVar` in a module using future annotations. The comment now names both misses — that one, and the renamed import `ClassVar as CV` — instead of implying the text path is exact. The **object** path, which runs whenever the module has not asked for future annotations, has neither problem, and `test_a_renamed_import_is_not_resolved_in_the_source_text_form` pins the hole so it is recorded rather than discovered.

Happy to reverse the dial if you would rather a user type never be refused; it is a one-character change to the boundary set.

<details>
<summary><b>3 and 4 — fail-open, and the <code>__origin__</code> probe. Not changed.</b></summary>

**3.** `module_attribute` does conflate "not imported" with "lookup failed", and the failures it can conflate are `MemoryError` from a two-word `PyUnicode_FromString` and an import-system error on a `sys.modules` read. Distinguishing them means an out-parameter or a tri-state through a function whose whole job is "give me this if it is already loaded". Under a real `MemoryError` the class creation that follows fails anyway.

**4.** Right that a user `__getattr__` on an annotation object runs once per field, and that its errors are cleared. That is inherent to asking any object for `__origin__`, which is how `Annotated` and every subscripted generic are recognised — the alternative is a type allowlist, which fails on the next generic form. The `PyType_Check` short-circuit added last round already keeps it away from plain classes, which is nearly every annotation.

</details>

<details>
<summary><b>What clang-tidy caught in my first attempt</b></summary>

```
error: narrowing conversion from 'int' to signed type 'char' is implementation-defined
  348 |    char const before = found == source ? '\0' : found[-1];
```

The ternary promotes to `int`. Fixed without a cast — the boundary test does not need the intermediate.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Two fixed, four filed as #57. `camas check` green at 25/25 on `3c244bd`.

**1 — the spaced forms.** Confirmed, and it is a hole I opened last round by widening `opens` without widening `after`:

```
'ClassVar'               refused
'ClassVar '              ACCEPTED -> fields=('v', 'name')
'ClassVar [int]'         ACCEPTED -> fields=('v', 'name')
'Annotated[ ClassVar ]'  ACCEPTED -> fields=('v', 'name')
```

Rather than add `' '` to the closer list, I changed the question. Enumerating openers and closers is the wrong shape — what is being asked is whether the match is a *whole identifier*, so neither neighbour may be an identifier character. One rule, both ends, nothing left to keep adding punctuation to:

```c
bool const opens = found == source || !identifier_character(found[-1]);

if (opens && !identifier_character(found[length])) {
```

```
'ClassVar'               REFUSED      'MyClassVar'   accepted
'ClassVar '              REFUSED      'ClassVarish'  accepted
'ClassVar [int]'         REFUSED      'ClassVar_'    accepted
'Annotated[ ClassVar ]'  REFUSED      '_ClassVar'    accepted
'typing.ClassVar[int]'   REFUSED      'list[int]'    accepted
```

Both parametrized lists grew to cover the new spacings and the new negatives, so widening the separator set can never again widen the form set unnoticed.

**3 — `module_attribute` fails open.** Taken, and **my rebuttal last round was wrong in exactly the way you said**: I claimed class creation fails anyway, and it does not, because the error had already been cleared. Absent and failed are now told apart by whether an exception is set, and only `AttributeError` on a module that exists is swallowed:

```c
PY_OWNED(class_var, module_attribute("typing", "ClassVar"));

if (class_var == NULL && PyErr_Occurred()) {
    return RESULT_ERROR;
}
```

**2, 4, 5 and 6's clearing — filed as #57.** Not patches, and I would rather they were written down than argued in a PR thread:

- `__args__` is never walked, so `Optional[Annotated[ClassVar[int], 'm']]` escapes the object path while the text path refuses it.
- `Annotated[int, ClassVar]` is the mirror image — the text path refuses legal metadata, the object path accepts it.
- The 3.14 factory-local alias: FORWARDREF resolves against the module's `__globals__`, so the same source is refused on 3.13 and accepted on 3.14.

Those three are one problem — an exact check on objects and a spelling check on text will disagree at the edges — and #57 sets out the three fixes that actually address it, including walking `__args__`, which buys agreement by adopting the false refusal rather than by removing it. Choosing among them is worth more thought than this PR should carry.

The remaining half of 6 (a hostile `__getattr__` mutating the live `__annotations__` mid-`PyDict_Next`) is on #57 too. It needs a dict copy per class creation, and class-creation speed is the thing salix is for, so it wants a measurement before a patch.

**Note on this PR's review history:** three review jobs were cancelled mid-flight before this one landed, so `702440b` went unreviewed for about 90 minutes. Reported to the code-review repo's #107 with the timing data.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues, answer each finding on its PR, and split side-discussions into follow-up issues. This is round 4 on #50.

JPHutchins:
Findings 1 and 2 are regressions I introduced last round; 1 and 3 are fixed, 2 cannot be. Six taken or filed, four already on #57. `camas check` green at 25/25 on `6da83b0`.

**1 — the boundary was ASCII-only.** Confirmed, and it is a regression: the enumerated-punctuation matcher I replaced did not have it.

```
'théClassVar'   REFUSED        ->  accepted
'ClassVaré'     REFUSED        ->  accepted
'ÄClassVar'     REFUSED        ->  accepted
```

PEP 3131 identifiers hold Unicode letters and this reads UTF-8, so the lead byte of `é` was being read as a separator. A byte at or above `0x80` is now an identifier character — the middle of a name rather than the end of one. All three are in the negative parametrize list now.

**3 — the NUL escape.** Also confirmed, and this one is the unsafe direction:

```
'x\x00ClassVar[int]'   ACCEPTED, fields=('v', 'name')
```

`strstr` stops at the first NUL and a `str` is allowed to contain one, so the rest went unscanned and the guard was silently off — #14 with an escape hatch. The scan now takes its length from `PyUnicode_AsUTF8AndSize` and reads the whole buffer. Pinned in the refusal list.

**2 — the string-literal false refusal. Cannot be fixed by loosening the boundary**, and the measurement says why:

```
'ClassVar[int]'               REFUSED   <- correct: a nested forward reference
Annotated[int, "ClassVar"]    REFUSED   <- wrong: the form is metadata
```

Both are the form adjacent to a quote. Making quotes non-boundaries fixes the second and breaks the first — trading a loud false refusal for a **silent false accept**, which is the direction this whole PR exists to close. Only parsing separates them, because the difference is position in the expression rather than spelling. Recorded on #57 with that reasoning, and the code comment now says it rather than leaving it to be rediscovered.

**9 — probes on annotation-less classes.** Taken. A class with no annotations of its own declares no fields and can name no forms, so `append_declared` returns before either lookup.

**10 — duplicated descriptors.** Taken. `CLASS_VAR_FORM` and `INIT_VAR_FORM` are the answers; the two paths differ only in how they ask, which was the point you were making.

**5 — the `sys.modules` snapshot.** New, verified, and on #57. `del sys.modules["typing"]` after building the annotation disables the identity checks. The fail-closed change told *absent* from *failed*; it cannot tell *never imported* from *was here, now gone*, and the annotation object is the only remaining evidence — which is what option 1 on #57 would use.

**4, 6, 7, 8 — unchanged, all on #57.** The `__args__` gap, the `Annotated[int, ClassVar]` metadata refusal, the 3.14 FORWARDREF routing, and the `__origin__` probe's clearing. Each needs a design decision; #57 sets out the three that would actually settle them.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues, answer each finding on its PR, and keep side-discussions on follow-up issues. This is round 5 on #50.

JPHutchins:
Finding 5 is a regression from last round's fix, and it and 6 had one cause. Seven taken, five already on #57. `camas check` green at 25/25 on `984ab09`.

**5 and 6 — I was matching bytes when the question is about characters.**

Two rounds of hand-written boundary tests, each fixing one direction and breaking the other: ASCII-only refused `théClassVar`, and every-byte-≥-0x80 let `ClassVar€` through. The second is the silent direction, so it was the worse trade.

`names_form` now takes the `str`, finds the form with `PyUnicode_Find`, and asks **`PyUnicode_IsIdentifier`** about the neighbours — via `"a" + character`, which is the public spelling of "does this continue an identifier". Python owns the rule; nothing here classifies characters any more.

```
ClassVar€  €ClassVar  ClassVar  'ClassVar[int]'  x\0ClassVar[int]   REFUSED
théClassVar  ClassVaré  ÄClassVar  ClassVaŕ  MyClassVar             field
```

That also retires finding 6 without a separate fix. There is no encode left to fail: a lone surrogate used to make `PyUnicode_AsUTF8AndSize` fail, the error was cleared, and the annotation became a field. `chr(0xD800) + "ClassVar"` is now refused like any other spelling — and the NUL case stops being special rather than being specially bounded.

**8, 9, 10 — the test gaps, all taken, and 9 is the one I should have written first.** Every text-path case in this file was a string the test built itself, so nothing proved the path is reachable the way a *user* reaches it. There are now two real `from __future__ import annotations` modules compiled and executed — one refusing, one building ordinary fields — because refusing everything would have satisfied the first on its own. Same reasoning for 8: every `Annotated` object-path test was a refusal, so a walk that reported a form for everything passed them all; there is a positive control now.

**13 — the refusal test never named the form.** Right, and it is the cheapest possible way for the two `names_form` calls to be swapped without a single failure. Now parametrized with the expected form and matching `annotated {form}`.

**12 — probing `__origin__` on a bare interpreter.** Taken; `special_form_of` returns before the walk when neither module is loaded, since no comparison in it could match.

**1, 2, 3, 7, 11 — unchanged, all on #57**, along with **4**, which I answered last round: quotes have to stay boundaries because `x: 'ClassVar[int]'` is a nested forward reference that must be refused, and loosening them trades a loud false refusal for a silent false accept.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues, answer each finding on its PR, and keep side-discussions on follow-up issues. This is round 6 on #50.

JPHutchins:
Finding 8 turned up something better than the test it asked for. Five taken, six on #57, one is #44's. `camas check` green at 25/25 on `b538bf0`.

**8 — ZWNJ, ZWJ, middle dot.** Added them and the suite went red on 3.10, 3.11 and 3.12 while passing on 3.13+. **CPython changed the rule:**

```
3.10  ZWNJ=False  ZWJ=False  middot=True  acute=True
3.12  ZWNJ=False  ZWJ=False  middot=True  acute=True
3.13  ZWNJ=True   ZWJ=True   middot=True  acute=True
```

So `ClassVar<ZWNJ>` is one name on 3.13 and two tokens on 3.12 — and salix refuses it on 3.12 and accepts it on 3.13, for exactly the right reason. That is the payoff of asking `PyUnicode_IsIdentifier` rather than carrying a table, and I would not have known it without your list.

A list of expected answers would have pinned one side and called the other a regression, so the test asserts the **agreement**: for eleven trailing characters it compares salix's verdict against `("a" + character).isidentifier()` on whatever interpreter is running. Nothing to update when the rule moves again.

**6 — the text path still failed open.** Right, and it is the same fail-open I fixed in `module_attribute` two rounds ago, in the path deciding the same question. `PyUnicode_FromString`, `PyUnicode_Find` and `PyUnicode_New` failures were cleared and answered "not a form", which reads as "this is a field". Nothing clears what it did not expect now, and `append_declared` fails the class when the lookup itself failed rather than when it found nothing.

**10 — the `-1` return.** Folded into the same change; no branch clears an error it did not cause.

**7 — bare `Annotated[ClassVar, 'meta']`.** Taken, skipped below 3.11 where typing rejects it as an argument before salix is reached.

**12 — `ANNOTATION_FORMAT_FORWARDREF = 2`.** Correct and real on this branch, which is cut from main. It is **#44's fix** — that PR replaces the enum with all four of annotationlib's values and the right numbering, and this branch inherits it on merge. Confirmed by diffing the two.

**1, 2, 3, 4, 5, 9 — on #57.** Finding 1 broadens the `__args__` gap usefully (`list[ClassVar[int]]`, not just `Optional[...]`); I have not re-edited the issue for that, since option 1 there covers the whole class.

**11 — the needles.** Declined for now: it is two small-string allocations per string annotation, only on the future-annotations path, and hoisting them means threading a struct through three functions that #57's option 2 would rewrite anyway.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues, answer each finding on its PR, and keep side-discussions on follow-up issues. This is round 7 on #50.

JPHutchins:
One taken, one declined with a measurement, five on #57. `camas check` green at 25/25 on the new head.

**6 — the `__origin__` probe's blanket clear.** Taken, and your framing is why: it was the last one. Every other fail-open in this path went during this PR — `module_attribute`, then `names_form`, then `continues_identifier` — and this one stayed because it sits in the object path rather than the text path. Not having an `__origin__` still ends the walk quietly; anything else leaves the error for `append_declared` to fail the class on.

**7 — the needles.** Declined, with the number. Class creation, 20 000 iterations each:

```
text path    6.22 us/class
object path  5.68 us/class
difference   +0.54 us  -- and the two allocations are somewhere inside it
```

So the whole text path, including two `PyUnicode_FromString` calls, two `PyUnicode_Find` scans and every `PyUnicode_IsIdentifier` probe, costs 0.54 µs against the object path — and only for modules using `from __future__ import annotations`. Hoisting the needles means threading them through three functions that #57's option 2 rewrites. I would rather leave the code the shape it is until that ruling.

**1, 2, 3, 4, 5 — on #57**, unchanged, and each needs a design decision rather than a patch:

- `__args__` never walked (broadened here to `list[ClassVar[int]]` and `tuple[...]`, which is a better example than the `Optional[...]` on the issue)
- the `sys.modules` snapshot
- the renamed import
- `Annotated[int, ClassVar]` as metadata
- the same shape quoted, which cannot be loosened without turning a loud false refusal into a silent false accept — measured two rounds ago

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues, answer each finding on its PR, and keep side-discussions on follow-up issues. This is round 8 on #50.

JPHutchins:
One taken, six unchanged. `camas check` green at 25/25 on the new head.

**6 — the fail-closed probe had no test.** Fair, and a good habit to hold me to: the fix rested on the same reading of the code that produced it. `test_a_failing_origin_probe_fails_the_class` covers both directions — an annotation object whose `__getattr__` raises `RuntimeError` for `__origin__` fails the class with that error, and one that raises `AttributeError` for everything is an ordinary field.

**1, 2, 3, 4, 5 — all on #57**, unchanged and unchanged for the same reason: each needs a design decision, and the issue sets out the three that would settle them. Nothing new in this round's phrasing of them.

**7 — the needles.** Declined again, and the measurement from last round stands: the whole text path costs **+0.54 µs/class** against the object path, allocations included, and only for modules using `from __future__ import annotations`.

For what it is worth, rounds 8 and 9 on this PR found one actionable item each, both of them tests for code I had just changed. The C is where it was two rounds ago.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues, answer each finding on its PR, and keep side-discussions on follow-up issues. This is round 9 on #50.

JPHutchins:
Three taken, six unchanged. `camas check` green at 25/25 on the new head.

**7 — the error check was on the wrong side of the verdict.** Correct. `append_declared` only asked `PyErr_Occurred()` when no form was named, so an allocation failing inside `continues_identifier` *while the scan matched* would have had its error replaced by the refusal — a real failure reported as a ClassVar the user never wrote. The check is ahead of both branches now.

**8 — the oracle test only exercised one side.** Also correct, and it undercut the point of the test: the form was always at offset 0, so `opens` took the `found == 0` short-circuit every time and the leading boundary was never checked against Python at all. Each character now runs on both ends.

**9 — the bare-interpreter branch had no test.** Fair; the test module imports both modules, so the early return added two rounds ago was never reached. Dropping `typing` and `dataclasses` from `sys.modules` shows the walk skipped entirely — an annotation object's `__getattr__` is never asked for `__origin__` — and the class still builds.

**1, 2, 3, 4, 5 — on #57**, unchanged. **6 — the needles**, declined for the third time on the standing measurement (+0.54 µs for the whole text path).

Three rounds running now where every actionable finding has been a test or an error-handling ordering rather than the matcher itself. The C that does the deciding has not changed since round 7.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues, answer each finding on its PR, and keep side-discussions on follow-up issues. This is round 10 on #50.

JPHutchins:
One new escape recorded, one declined with reasoning, six unchanged. No code change; head stays `3d1d758`, green at 25/25.

**2 — PEP 695 aliases.** Confirmed, and it is reachable with no trick at all:

```python
type CV = ClassVar[int]

class Q(Struct):
    registry: CV
    name: str

Q.__struct_fields__   # ('registry', 'name')
Q(5, 'x').registry    # 5
```

`TypeAliasType` has no `__origin__` and does have `__value__`, so the walk stops at the first hop.

Added to **#57** rather than patched, and the reason is worth stating because it is now the third of its kind: the object path follows exactly one attribute, and each escape found asks it to follow one more — `__args__` for `Optional[Annotated[...]]` and `list[ClassVar[int]]`, `__value__` for this. Adding them one report at a time is how the *text* matcher's boundary rule got fixed twice and broken twice in this very PR. #57's option 1 should settle which attributes a form can hide behind; `SPECIAL_FORM_HOPS` already bounds the depth, and what is missing is the breadth.

**7 — the opens side uses the closes oracle.** Declined, and I think the alternative would be wrong. The natural start-side oracle is `(character + "a").isidentifier()`, which for a digit says False — so it would expect salix to refuse `1ClassVar`. salix accepts it as a field, and so does Python in the only sense that matters: `1ClassVar` is not something a source file can contain, so there is no annotation for the two to disagree about. Asserting on it would pin behaviour for an input that cannot occur.

The oracle that *is* meaningful on both sides is "does this character join to the form", which is the continue predicate, and it is the one in use. You are right that the two sides therefore cannot fail independently — but the branch is exercised, and I would rather that than an assertion about `1ClassVar`.

**1, 3, 4, 5, 6 — on #57. 8 — the needles**, declined on the standing +0.54 µs measurement.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through the automated review rounds on salix's pre-publish issues, answer each finding on its PR, and keep side-discussions on follow-up issues. This is round 11 on #50.

JPHutchins:
**One taken with numbers, one refuted with numbers, six still waiting on #57.** `a7ec6e2`, green at 25/25.

**8 — taken.** `names_form` allocated its needle on every call, and it is called once per field per form. Both needles now live in a `struct form_probes` that `append_declared` builds beside the two module attributes it already builds once per class.

Measured on the current head — 16 fields, 2000 classes, min of 25 runs, the two builds swapped back and forth so the machine's drift lands on both:

| annotations | before | after |
| --- | --- | --- |
| text (`from __future__ import annotations`) | 10.41–10.56 µs/class | **9.17–9.94** |
| objects | 9.78–10.06 µs/class | 9.86–9.95 |

<details>
<summary><b>Why they are built conditionally, which was not the first version</b></summary>

Building them for every class cost the object path about **110 ns/class** — the version before this one did that, and the same A/B showed it as a clean separation rather than noise:

```
before 9.782    after 9.950
before 9.831    after 9.864
before 9.859    after 9.967
before 9.942    after 10.081
before 9.925    after 10.052
before 9.981    after 10.063
```

1.1% on the path almost every class takes, to save 5% on the one almost none do, is the wrong direction for a library whose pitch is class-creation speed. `annotates_with_text` asks per class the same question the matcher asks per field — one `PyUnicode_Check` per annotation — and the object-path row in the table above is what is left of the cost: the two ranges overlap.

The needles come from `CLASS_VAR_FORM.name` and `INIT_VAR_FORM.name`, the same two strings the refusal message quotes, so each form is still spelled once in the file.

</details>

<details>
<summary><b>7 — the opens side does catch a divergence. Three mutations say so.</b></summary>

The finding is that the opens-side iteration "exercises the branch but cannot catch a divergence" because it passes the same continue-predicate oracle as the closes side. I broke the opens side three ways and re-ran `test_the_boundary_agrees_with_python_about_identifiers`:

| mutation | result |
| --- | --- |
| `bool const opens = true;` | **11 failed**, incl. `[letter] [underscore] [digit]` |
| opens reads the *closing* character instead of the opening one | **11 failed**, same three ids |
| opens asks "starts an identifier" rather than "continues one" | caught by `[digit]`: `1ClassVar` |

The oracle being the same on both sides is the point rather than the gap: the C code applies `continues_identifier` to both neighbours, so an oracle that asked something else on one side would assert the wrong answer. What makes the two iterations independent is that each is a separate class creation with a separate assertion — `"sClassVar"` and `"ClassVars"` fail separately, and the `[letter]`, `[underscore]`, `[digit]` ids above are exactly the characters where a wrong predicate on one side and not the other shows up.

Happy to be shown a divergence that survives all three; I could not construct one.

</details>

**1, 2, 3, 4, 5, 6 — unchanged, and they are #57.** Six of eight findings this round are entries on that issue, and they have come back unchanged for four rounds now because the fix for all of them is one design decision:

> **#57 option 1 — walk `__args__` as well as `__origin__` on the object path.** It closes the `Optional[Annotated[ClassVar[int], 'm']]` accept and makes the two paths agree — by having the object path adopt the text path's false refusal of `Annotated[int, ClassVar]`, which is legal code. That trade is the decision, and it is not mine to make.

@JPHutchins — this branch cannot converge without that one. Everything in it that is fixable without a ruling is now fixed.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked that the open PRs keep iterating against the automated reviewer now that its runner problem is fixed; this round answers review 11 on this branch and re-raises the one ruling it is blocked on.

JPHutchins:
**Finding 8 is gone, 7 is refuted with a build, and 1–6 are #57.** `e9b1dcc`, green at 25/25.

**7 — the sharpened version is a better argument, and it is still wrong.** The proposal is to make the opening-side oracle `(trailing + "ClassVar").isidentifier()`, pinning the leading character against Python's name-START rule. Only an implementation that asks *can this character begin an identifier* can satisfy that oracle, so I built one:

| | `x1ClassVar` | `1ClassVar` | `MyClassVar` | `ClassVar[int]` |
| as written | field | field | field | REFUSED |
| name-START on the opening side | **REFUSED** | REFUSED | field | REFUSED |

`x1ClassVar` is a legal Python identifier. The character before the form is a digit, which continues a name without being able to start one, so the proposed rule reads `ClassVar` as standing alone and refuses a name Python is perfectly happy with. It also fails five of the existing boundary cases.

Whether a prefix is a valid identifier is not decidable one character at a time — that is the whole reason both sides ask the same question. `x1ClassVar` is now in the merely-contains-the-form parametrization, so the next attempt at this fails a test rather than a review round.

The remaining half of the finding — that `1ClassVar` stays a field — is documented rather than changed: `x: 1ClassVar` is a syntax error, so it reaches the text path only through a hand-written `__annotations__`, and the rule that gets `x1ClassVar` right gets this one "wrong" as its consequence.

**1–6 are #57, and this branch has nothing left that is not.** Six findings, four rounds, unchanged, because the fix for all of them is one decision:

> **#57 option 1.** I re-raised it there with a correction to how I framed its cost: the text path **already** refuses `Annotated[int, ClassVar]` and ships that way, so option 1 does not adopt a new false refusal — it stops salix having two answers. Both shapes in the trade are already rejected by mypy and pyright, since PEP 526 requires `ClassVar` at the top level of an annotation.

@JPHutchins — either answer unblocks this. Take option 1 and findings 1 and 2 close and 5 and 6 become one agreed answer instead of two; decline it and #50 merges with the six recorded on #57, where a reader can find them. What it cannot do is keep waiting: the reviewer re-derives the same six every round, and it has now done so four times.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked that the open PRs keep iterating against the automated reviewer now that its runner problem is fixed; this answers the review that landed on `a7ec6e2`.

JPHutchins:
**Eight taken, one declined with numbers, five are #57.** `da11c55`, green at 25/25. Two commits: the ruling, then this review.

## The ruling landed — #57 option 1 is in

`728ced4` walks `__args__` as well as `__origin__`, which closes **finding 1** and, as it turns out, **finding 7** — the off-by-one in `SPECIAL_FORM_HOPS` is gone because the budget is no longer a depth. Five shapes that were fields on the object path while the text path refused the identical source:

```
Annotated[ClassVar[int], "m"] | None    field  ->  REFUSED
list[ClassVar[int]]                     field  ->  REFUSED
dict[str, ClassVar[int]]                field  ->  REFUSED
tuple[ClassVar[int]]                    field  ->  REFUSED
list[InitVar[int]]                      field  ->  REFUSED
```

**And the cost I told #57 the ruling would have does not exist.** I said option 1 buys agreement by having the object path adopt the text path's refusal of `Annotated[int, ClassVar]`. It does not — `Annotated` keeps metadata out of the arguments:

```console
>>> Annotated[int, ClassVar].__args__, Annotated[int, ClassVar].__metadata__
((<class 'int'>,), (typing.ClassVar,))
```

so the walk never reaches it and that shape is still a field. Finding 6's disagreement therefore survives untouched, and is still #57's.

A queue rather than a recursion, because `misc-no-recursion` is an error here; a fixed array of owned references with a 32-node budget, on the stack, because an allocation would be one per subscripted annotation per class.

<details>
<summary><b>What the walk costs, measured</b></summary>

16 fields, 2000 classes, min of 20, builds swapped back and forth:

| annotation | before | after |
| --- | --- | --- |
| `int` | 10.36–10.45 µs/class | 10.41–10.56 |
| `list[int]` | 11.26–11.40 µs/class | **12.50–12.62** (+11%) |
| `int \| None` | 11.56–11.68 µs/class | **12.42–12.65** (+8%) |

An annotation with no subscript pays nothing — `PyType_Check` ends the visit before either lookup. One with a subscript pays a second attribute lookup and a visit per argument. That is the price of the ruling and it is worth reporting rather than burying.

</details>

## 8 — taken, and I could not make it misbehave

You are right that the walk runs the class author's `__getattr__` inside a `PyDict_Next` loop over the dict that code can reach. I tried to corrupt it — adding 64 keys from inside the walk — and it built the class cleanly. That is not the same as being safe, so the names are taken first with `PyDict_Keys` and each annotation is looked up again as it is reached. One list per class.

The behaviour is defined now instead of incidental, and pinned: a field added during the walk is not declared, one deleted during it is skipped.

<details>
<summary><b>10, 11, 12, 13, 14 — all taken, and 14 was mine</b></summary>

- **14.** Correct, and it is a regression I introduced in `e9b1dcc`: I replaced an accurate "The last five are legal identifiers" with "Five of these", and eleven of the fourteen are. Counted with `str.isidentifier` this time rather than by eye.
- **13.** The docstrings explained each case by the bug it used to be, which none of them can establish — they would pass on an implementation that never had those bugs. The history is now marked as history and the cases stay as pins.
- **12.** It said "on a bare interpreter" and claimed salix imports neither module. It runs on a populated interpreter with two entries removed, and establishes nothing about what salix imports. Both claims gone.
- **11.** InitVar had never been through the compiler. Both forms are parametrized through the real `from __future__ import annotations` module now — which matters precisely because what the compiler stores is text, and the two forms are different text.
- **10.** Matched only "salix does not support", so it would have passed with the two labels swapped. It asserts the form's own label now.

</details>

<details>
<summary><b>9 — declined, with what is left of it</b></summary>

The needle allocation was the bulk and `a7ec6e2` moved it out; what remains is the second scan and `continues_identifier`'s per-boundary probe. Bounding it from the earlier measurements on this branch: the whole text path was 54 ns/field, the needles were 34 of that, so everything you name here is inside the remaining ~20 ns/field — and only on the `from __future__ import annotations` path.

Merging the two scans means one pass recognising two needles, and removing the probe allocation means a per-class scratch buffer written in place. Both are real; neither is worth the surface at 20 ns on the uncommon path while #57 still has open questions about what this matcher should even accept. Happy to take it if you disagree — it is the kind of thing better done once the semantics stop moving.

</details>

**2, 3, 4, 5, 6 are #57**, unchanged. Finding 2 in particular needs a `__value__` hop rather than an `__args__` one, so it is not what was ruled on — I have asked there whether to add it, since it is one hop and closes another live #14.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She ruled option 1 on #57 and asked that the open PRs keep iterating against the automated reviewer; this lands that ruling and answers the review on `e9b1dcc`.

JPHutchins:
**The critical one is real, it is mine, and it is fixed.** `8e397a7`, green at 25/25.

## 1 — segfault, reproduced before fixing

```console
$ python segv.py
building...
$ echo $?
139
```

Two commits that were each correct combined into a NULL dereference:

- `728ced4` made the walk read `__args__`, which runs more of the class author's `__getattr__`.
- `da11c55` made the loop re-look-up each annotation, so a value that code injects is now *seen*.
- And the needles came from `annotates_with_text(annotations)` **before** the loop — a question asked of a dict the walk can change.

Answer "no text here", inject a str from `__getattr__`, and `names_form` gets a NULL needle. Exactly as you described it, including that the second commit is what made the first one reachable.

The needles are built at the first str annotation now, inside the loop, which is where the value is used — a decision made where it is used cannot go stale between the decision and the use. `annotates_with_text` is gone, and with it the extra pass over the dict that **finding 10** counted. The injected annotation is now *refused*, which is also the right answer: it says `ClassVar[int]`.

## 12 — the same class of mistake as the segfault's cause, one comment over

`form_within`'s comment claimed the walk "makes the object path refuse `Annotated[int, ClassVar]`" and gave a `__args__` str example. Neither happens — Annotated keeps metadata in `__metadata__` — and `test_the_form_as_annotated_metadata_is_still_a_field_on_the_object_path`, which I wrote **in the same commit**, pins the opposite. Corrected; the disagreement is #57's and the comment says so.

**14** — `module_attribute`'s tail was `optional_attribute`'s body written out again, in the PR that added the helper. It calls it. **13** — the five new `__args__` cases matched only "salix does not support", so they would have passed with the labels swapped; parametrized with the form.

<details>
<summary><b>2, 3, 11 — real, and each is a fail-open I would rather record than guess at</b></summary>

**2, the 32-node budget.** A form past the frontier becomes a field. True, and the old `SPECIAL_FORM_HOPS = 4` had the same shape with a tighter bound — a chain of five hops failed open before this PR too. What changed is the number, not the direction. Raising it trades one arbitrary bound for another; refusing at the budget turns a deep-but-legal annotation into a `TypeError`. Recording it on #57 with the other known escapes rather than picking one here.

**3, a non-tuple `__args__`.** Same shape: `PyTuple_Check` with no else. Handling it properly means `PySequence_Fast` over something a `__getattr__` synthesised, which is more user code inside the walk — the thing that produced the segfault above. Also #57.

**11, a key deleted mid-walk.** `__struct_fields__` omits it while the class's `__annotations__` still lists it. Agreed that they disagree; what they should do instead is a real question — refuse the class, or keep the field with no value — and both are worse than the current answer for a class nobody meant to write. Recording it.

</details>

**4, 5, 6, 7, 8 are #57**, unchanged, and **9** (the borrowed-ref window under free threading) I want to take properly with `PyDict_GetItemRef` rather than squeeze into this round — it is a real free-threading point on a project that ships 15 free-threaded wheels, and it deserves its own measurement.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked that the open PRs keep iterating against the automated reviewer; this answers the `changes` verdict on `da11c55`, whose critical finding was a segfault introduced by the two preceding commits.

JPHutchins:
**Round 15: two taken, four refuted with the case each describes, nine recorded on #57 or measured and declined.** `0a5a864`, `fa66584`, `75df06b` — three commits, green at 25/25.

Two of the fifteen were already fixed in commits held back so they would not land under the review in flight:

| # | | |
| --- | --- | --- |
| 2 | PEP 695 alias never unwrapped | **fixed** — `fa66584`, the `__value__` hop [@JPHutchins approved on #57](https://github.com/JPHutchins/salix/issues/57#issuecomment-5186827993) |
| 10 | borrowed reference across a free window | **fixed** — `0a5a864`, `PyDict_GetItemRef` under the dict's lock |
| 14 | the attribute probe written twice | **fixed** — `75df06b`, it lives in `owned.h` and `meta.c` uses it |
| 15 | `form_probes` double-books every pointer | **fixed** — `75df06b`, it owns them through a cleanup; four mirrors gone |
| 12 | plain-type annotations pay for the frontier | **measured, declined** — the patch is slower, and there is no memset |
| 11 | probes re-fetched per annotated class | **declined** — same reason 12's patch is wrong |
| 7 | any object with `__origin__`/`__args__` is walked | **refuted** — refusal needs the form to be *in* them |
| 8 | an `AttributeError` subclass reads as absent | **declined** — `hasattr` reads it the same way |
| 13 | two scans, one allocation per boundary | **declined** — the allocation happens at a hit, not per character |
| 1, 3, 4, 6, 9 | budget, renamed import, quotes, metadata position, non-tuple `__args__` | **#57**, where each is already in the remaining-work table |
| 5 | the guard turns off if `typing` leaves `sys.modules` | **reproduced, added to #57** |

## 12 — I built it, and it costs rather than saves

Two things are wrong with it, and the second is the interesting one.

**The patch as written reopens #14.** `if (PyType_Check(annotation)) return no-form;` at the top of `special_form_of` makes `x: InitVar` a field, because the bare `dataclasses.InitVar` **is** a class — `test_a_bare_init_var_is_refused_like_a_bare_class_var` fails on it. The behaviour-preserving version is `return form_named_by(annotation, probes)` there, which is exactly what the walk does for a type anyway.

**And that version does not pay.** Two prebuilt `.so` files, swapped between runs, three interleaved rounds, 16 plain `int` fields per class:

```console
before 7.976 us/class      after 8.005 us/class
before 7.915 us/class      after 7.985 us/class
before 8.034 us/class      after 8.068 us/class
```

The early-out is *slower* in all three pairs. So the premise is worth checking too, and it does not hold:

```console
$ gcc -S -O2 @compile_flags.txt src/fields.c -o - | grep -c memset
0
```

**There is no 256-byte memset.** The compiler can see that `nodes[]` above `count` is never read, so zeroing the frontier costs one store to `count`. What the walk actually spends on `x: int` is a `Py_NewRef` and a `Py_DECREF`, and adding a branch to skip them is not free either.

Reverted, and reporting the numbers instead.

## 11 — the same answer

The two module lookups cannot move behind a "no annotation can be a form" test, because a plain class *can* be one: that is what 12's patch got wrong. They are two `sys.modules` lookups and two attribute reads **per class**, in a class creation that costs ~8 µs.

## 7 — the refusal needs the form to be in them, not the name on them

Built all three cases:

```console
instance with __args__ = (int, str)        field
instance with __origin__ = list            field
instance with __args__ = (ClassVar,)       REFUSED
```

So an object is not refused for *having* the attributes — it is refused for the form being reachable through them. The third row is the one the finding calls a false refusal, and I think it is the right answer: an annotation that hands back `ClassVar` when asked for its arguments is indistinguishable from one, and this guard is fail-closed by construction.

## 8 — absent is what Python says it is

A `__getattr__` raising any `AttributeError` is saying the name is not there, and everything else that asks reads it that way:

```console
hasattr(Refuses(), "__origin__")                False
getattr(Refuses(), "__origin__", "absent")      'absent'
```

Matching the exact type would make salix the only reader of the protocol that treats an `AttributeError` subclass as a hard failure, and would break a library that subclasses it for a message. The comment now says absent means any of them, and why.

## 13 — the allocation is per hit, not per character

`continues_identifier` runs only after `PyUnicode_Find` has already located the needle, so for the annotation that has no `ClassVar` in it at all — every annotation, nearly always — it never runs. Two `PyUnicode_Find` passes over a short string is the whole cost. What the allocation buys is PEP 3131: `théClassVar` is not the form standing alone, and no ASCII test gets that right.

## 5 — reproduced, and it belongs with the others

It is real, and it takes removing the module after the annotation exists:

```console
ClassVar[int] with typing out of sys.modules    field
ClassVar[int] with typing back                  REFUSED
```

The fix you name — holding the objects from salix's own import state — means importing `typing` when salix is imported, which is a cost on every program that never writes a `ClassVar`. The alternative is caching the first one found, which is module state and a lifetime question at round 15 of this PR. Recorded on #57 beside the other four, where the whole set can be answered at once.

## 1, 3, 4, 6, 9 — #57's table, and it is accurate

Confirmed the budget repro, including that the first argument is not spared:

```console
tuple[ClassVar[int], int]              REFUSED
tuple[ClassVar[int], *([int] * 39)]    field
```

Worth saying why, because it is not "the 40th argument is dropped": `ClassVar[int]` is visited — it is the second node — but expanding it pushes `ClassVar` past the 32nd slot, and the match is on `ClassVar`, not on `ClassVar[int]`. A bigger budget moves the wall; only an unbounded frontier removes it.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, answering round 15's fifteen findings on her PR.

JPHutchins:
**Two taken; the other ten are the escapes #57 already carries, and they have not changed.** Green at 25/25, `nix flake check` green, and `main` is merged in — #49 landed a `reject_unsafe_default` into the same file this PR rewrites, which git could not align. Resolved by keeping both; the suite is 812 passed, 8 skipped, 1 xfailed on the merged tree.

## 11 — a scan that could not look ran the next one on top of its exception

Taken. `PyUnicode_Find` answers `-1` for "not in the text" and `-2` for "could not look", and `names_form` spelled both as `false`, so the second needle scan started with an exception pending. `append_declared` still failed the class — it reads `PyErr_Occurred()` before it reads the verdict — but it did so after running a scan it should not have.

`named_special_form` asks between the two now, and both `names_form` and the boundary check say in one place that `false` means "no, or could not tell":

```c
if (names_form(text, probes->class_var_name)) {
    return CLASS_VAR_FORM;
}

if (PyErr_Occurred()) {
    return (struct special_form){0};
}

return names_form(text, probes->init_var_name) ? INIT_VAR_FORM : (struct special_form){0};
```

## 12 — the comment described the walk it used to be

Taken, and it is the third time a comment here has outlived the code it describes. "The `__origin__` chain is walked" and "a self-referential `__origin__` cannot spin" were both true before #57's ruling and the `__value__` hop; the enum ten lines below already said so, which is how you found it. One paragraph now, describing the tree: `__origin__`, every element of `__args__`, and a PEP 695 alias's `__value__` are all edges, and the frontier is what stops a cycle *whichever* edge points back.

## 1, 3, 5, 6, 7, 9 — the recorded set, unchanged

The budget, the renamed import, the quoted string, `Annotated[int, ClassVar]`, the sys.modules window and the non-tuple `__args__` are the table on [#57](https://github.com/JPHutchins/salix/issues/57), each with what closing it would take. I confirmed the budget repro again last round, including that the first argument is not spared. Nothing here has moved them; they need the tokeniser or a growable frontier, and #57 is where that is argued as one decision rather than six.

## 2 and 8 — the `__value__` hop's own two rows

Both are consequences of the hop [@JPHutchins approved](https://github.com/JPHutchins/salix/issues/57#issuecomment-5186827993), and both are the disagreement #57 exists to hold: the object path now refuses `type CV = ClassVar[int]` where the text path still accepts it, and an object that answers `__value__` with the form is treated as the form. Added to #57's table rather than argued here.

## 4 and 10 — noted, not done

4 is a test-isolation risk in a test that does not exist yet, so there is nothing to fix; if a sys.modules test is written it will need `monkeypatch` rather than a bare `del`. 10 is right that the two error branches in `form_within` have no Python-level test — they need an allocation failure or a `__getattr__` that raises a non-AttributeError at the second probe. The second is reachable and is worth a test; it is not worth another round on this PR.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, answering the latest round on her PR.

PR #52, opened 2026-08-01
test: cover methods on a struct, in the shapes people write them

JPHutchins:
Nothing in the suite exercised a struct as a class *with behaviour*. A codec — `to_bytes`, `to_string`, a `from_bytes` classmethod — is the shape that motivates reaching for a struct, and none of it was pinned.

Eighteen tests over a wire-frame struct: instance methods, a classmethod that builds through the generated constructor, a staticmethod, a property, a body-defined `__len__` coexisting with the generated `__repr__`/`__eq__`/`__hash__`, the class docstring surviving, and all of it inherited, overridable, and correct on a subclass.

**Tests only.** Every gap is recorded as it behaves; none is fixed.

<details>
<summary><b>What does not work: <code>functools.cached_property</code></b></summary>

A struct's fields are its slots and there is no instance `__dict__`, so there is nowhere to cache:

```
TypeError: No '__dict__' attribute on 'Cached' instance to cache 'expensive' property.
```

That is **parity, not a shortfall**. Measured across the alternatives:

```
salix                      TypeError: No '__dict__' attribute ...
dataclass(frozen, slots)   TypeError: No '__dict__' attribute ...
NamedTuple                 TypeError: No '__dict__' attribute ...
dataclass(frozen) no slots 200
```

Only a dataclass that carries a `__dict__` can do it, and carrying one is the thing salix exists not to do. Two alternatives that *do* work are pinned next to it:

- `functools.cache` on a method — caches on the function, so the absent `__dict__` is not in its way, at the cost of holding the struct alive.
- **a declared field written by `__post_init__` through `set_field`** — which a slotted dataclass cannot do at all, since it has no way through the frozen wall.

</details>

<details>
<summary><b>What is sharp: a method named after a field</b></summary>

The slot descriptor wins and the method is silently discarded:

```python
class Collide(Struct):
    x: int
    def x(self): return "method"

Collide(1).x        # 1
```

Also parity — `dataclass`, `dataclass(slots=True)` and `NamedTuple` all give `1` here. Pinned so a future change to `drop_class_variables` cannot alter it unnoticed.

</details>

`camas check` green at 25/25, across 3.10–3.15 and 3.14t.

> [!WARNING]
> **LLM Disclosure**
>
> This PR was authored by claude-opus-5 on behalf of @JPHutchins. She asked whether users can add methods to structs as they would to a dataclass or NamedTuple, then asked for tests covering that, `property` and `cached_property`, with gaps filed rather than fixed.

### DIFF
--- a/tests/test_methods.py
+++ b/tests/test_methods.py
@@ -77,15 +77,13 @@ class TestMethods:
 
         frame = Frame.from_bytes(b"\x02\x01\x07\x01")
 
-        assert type(frame) is Frame
         assert (frame.length, frame.kind, frame.flags) == (258, 7, 1)
 
     def test_a_round_trip_returns_an_equal_struct(self):
         """Composing pack with unpack holds by the struct module's own inverse
         property, so this pins the classmethod reaching `__eq__` with a real
-        instance and nothing about the wire format. The two that pin the format
-        are the ones checking `to_bytes` and `from_bytes` against independent
-        bytes; neither of them is the test directly above.
+        instance and nothing about the wire format. What pins the format is the
+        pair that check `to_bytes` and `from_bytes` against independent bytes.
         """
 
         original = Frame(4096, 3, 2)
--- a/tests/test_methods.py
+++ b/tests/test_methods.py
@@ -7,12 +7,16 @@ that motivate reaching for a struct in the first place -- `to_bytes`,
 `to_string`, a `from_bytes` classmethod -- plus the descriptors that sit
 alongside them, and the dunders that salix would otherwise generate.
 
-Methods are the subject; `__slots__` and `__match_args__` are salix's whatever
-the body says, and `TestBindingsSalixOwns` is where that is pinned.
+Methods are the subject. `__slots__` is salix's whatever the body says, and
+`__match_args__` is unless the class opts out of it; `TestBindingsSalixOwns`
+pins both, including where the body wins.
 """
 
+import dataclasses
 import functools
 import struct as struct_module
+import sys
+from typing import NamedTuple
 
 import pytest
 
@@ -73,14 +77,15 @@ class TestMethods:
 
         frame = Frame.from_bytes(b"\x02\x01\x07\x01")
 
+        assert type(frame) is Frame
         assert (frame.length, frame.kind, frame.flags) == (258, 7, 1)
-        assert isinstance(frame, Frame)
 
     def test_a_round_trip_returns_an_equal_struct(self):
         """Composing pack with unpack holds by the struct module's own inverse
         property, so this pins the classmethod reaching `__eq__` with a real
-        instance and nothing about the wire format -- the two tests above are
-        what would catch `<HBB` changing.
+        instance and nothing about the wire format. The two that pin the format
+        are the ones checking `to_bytes` and `from_bytes` against independent
+        bytes; neither of them is the test directly above.
         """
 
         original = Frame(4096, 3, 2)
@@ -103,7 +108,9 @@ class TestMethods:
         assert not Frame(1, 0).is_empty
 
     def test_a_dunder_the_body_defines_is_kept(self):
-        assert len(Frame(258, 7)) == 258
+        frame = Frame(258, 7)
+
+        assert (len(frame), frame.kind) == (258, 7)
 
     def test_a_body_repr_and_init_displace_the_generated_ones(self):
         """Two different mechanisms reaching the same place. `__repr__` is a
@@ -204,11 +211,9 @@ class TestCaching:
         """A struct's fields are its slots and it has no instance dict, so there
         is nowhere for cached_property to put the value.
 
-        `dataclass(slots=True)` fails with the same message. NamedTuple fails
-        too, but for its own reason below 3.13 -- `Cannot use cached_property
-        instance without calling __set_name__ on it`, because the NamedTuple
-        machinery never makes that call -- and with the missing-dict message
-        from 3.13 on. Only a dataclass carrying a __dict__ succeeds.
+        The siblings are asserted below rather than described here, because
+        every parity claim this file made in prose has turned out to be wrong at
+        least once.
         """
 
         class Cached(Struct):
@@ -220,6 +225,50 @@ class TestCaching:
 
         with pytest.raises(TypeError, match="No '__dict__' attribute"):
             _ = Cached(2).expensive
+
+    def test_a_slotted_dataclass_fails_the_same_way(self):
+        @dataclasses.dataclass(slots=True)
+        class Cached:
+            x: int
+
+            @functools.cached_property
+            def expensive(self) -> int:
+                return self.x * 100
+
+        with pytest.raises(TypeError, match="No '__dict__' attribute"):
+            _ = Cached(2).expensive
+
+    def test_a_namedtuple_fails_too_for_its_own_reason_below_3_13(self):
+        """Below 3.13 the NamedTuple machinery never calls `__set_name__`, so
+        the descriptor complains about that rather than about the missing dict.
+        """
+
+        class Cached(NamedTuple):
+            x: int
+
+            @functools.cached_property
+            def expensive(self) -> int:
+                return self.x * 100
+
+        expected = (
+            "No '__dict__' attribute"
+            if sys.version_info >= (3, 13)
+            else "without calling __set_name__"
+        )
+
+        with pytest.raises(TypeError, match=expected):
+            _ = Cached(2).expensive
+
+    def test_only_a_dataclass_carrying_a_dict_succeeds(self):
+        @dataclasses.dataclass
+        class Cached:
+            x: int
+
+            @functools.cached_property
+            def expensive(self) -> int:
+                return self.x * 100
+
+        assert Cached(2).expensive == 200
 
     def test_functools_cache_on_a_method_still_works(self):
         """It caches on the function rather than the instance, so the absent
@@ -249,8 +298,9 @@ class TestCaching:
 
 
 class TestBindingsSalixOwns:
-    """Not every body binding is the body's to keep. These two are salix's
-    whatever the class writes, and neither collides with a field name.
+    """Not every body binding is the body's to keep, and neither of these
+    collides with a field name. `__slots__` is taken unconditionally;
+    `__match_args__` only while the class wants one.
     """
 
     def test_a_body_slots_is_replaced_by_the_fields(self):
@@ -266,6 +316,19 @@ class TestBindingsSalixOwns:
             __match_args__ = ("nope",)
 
         assert Matched.__match_args__ == ("x",)
+
+    def test_opting_out_of_match_args_leaves_the_body_its_own(self):
+        """`match_args=False` means salix writes none, not that it writes an
+        empty one -- so the body keeps what it wrote, and `__slots__` does not.
+        """
+
+        class Matched(Struct, match_args=False):
+            x: int
+            __match_args__ = ("nope",)
+            __slots__ = ("extra",)
+
+        assert Matched.__match_args__ == ("nope",)
+        assert Matched.__slots__ == ("x",)
 
 
 class TestNameCollisions:
--- a/tests/test_methods.py
+++ b/tests/test_methods.py
@@ -1,10 +1,11 @@
 """Methods on a struct, in the shapes people actually write them.
 
-A struct's body is an ordinary class body: the generated `__init__`, `__eq__`,
-`__hash__` and `__repr__` do not displace anything it defines. These are the
-codec-shaped methods that motivate reaching for a struct in the first place --
-`to_bytes`, `to_string`, a `from_bytes` classmethod -- plus the descriptors that
-sit alongside them.
+A struct's body is an ordinary class body, and what it defines is kept -- with
+one exception, which is a name that collides with a field, where the slot
+descriptor wins. These are the codec-shaped methods that motivate reaching for a
+struct in the first place -- `to_bytes`, `to_string`, a `from_bytes` classmethod
+-- plus the descriptors that sit alongside them, and the dunders that salix
+would otherwise generate.
 """
 
 import functools
@@ -82,6 +83,27 @@ class TestMethods:
     def test_a_dunder_the_body_defines_is_kept(self):
         assert len(Frame(258, 7)) == 258
 
+    def test_a_body_repr_and_init_displace_the_generated_ones(self):
+        """The claim in this file's docstring, made testable: the two dunders
+        salix would have written are the body's when the body writes them.
+        """
+
+        class Custom(Struct):
+            x: int
+
+            def __repr__(self) -> str:
+                return "custom repr"
+
+            def __init__(self, x: int) -> None:
+                set_field(self, "x", x * 10)
+
+        instance = Custom(1)
+
+        assert repr(instance) == "custom repr"
+        assert instance.x == 10
+        assert Custom(1) == Custom(1)
+        assert hash(Custom(1)) == hash((10,))
+
     def test_the_generated_dunders_survive_alongside_them(self):
         frame = Frame(258, 7, 1)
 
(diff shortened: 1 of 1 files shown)


JPHutchins:
All three taken. Every one is about prose I wrote, and every one was wrong.

**1 — an untested headline claim, contradicted by the file's own tests.** Right on both counts. `__len__` was the only body dunder anywhere in the file, so the promise about `__init__`/`__eq__`/`__hash__`/`__repr__` was unenforced. It is a real guarantee, so it is a real test now — a body `__repr__` and a body `__init__` both win, while the `__eq__` and `__hash__` the body left alone are still salix's. The docstring also names the exception it previously ignored, which `TestNameCollisions` was demonstrating two screens below.

**2 — the NamedTuple mechanism.** Correct, and version-dependent, which is why I missed it measuring only on 3.14:

```
3.10  Cannot use cached_property instance without calling __set_name__
3.11  Cannot use cached_property instance without calling __set_name__
3.12  Cannot use cached_property instance without calling __set_name__
3.13  No '__dict__' attribute on 'N' instance to cache 'e' property.
3.14  No '__dict__' attribute on 'N' instance to cache 'e' property.
```

The NamedTuple machinery never calls `__set_name__` on the descriptor — a different failure that CPython's message stopped distinguishing in 3.13. The docstring says so now.

**3 — "what a slotted dataclass cannot".** False, and I checked it this time:

```
frozen+slots via object.__setattr__: 200
mutable+slots directly            : 200
```

A slotted dataclass fills a declared field in `__post_init__` perfectly well. What `set_field` buys is being the *supported* path rather than an escape hatch around the class's own rules — a smaller claim, and a true one.

Worth noting all three survived my own review because I checked the behaviour and then wrote a sentence that went further than what I had measured. The tests were right; the prose around them was not.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, answering the automated review on this tests-only PR.

JPHutchins:
Both taken; both were about prose, and both are now assertions instead. `camas check` green at 25/25 on `602cc2a`.

**1 — "with one exception" undercounts.** Confirmed, and neither of the other two is a collision:

```
class S(Struct):
    x: int
    __slots__ = ("extra",)          ->  S.__slots__ == ('x',)
    __match_args__ = ("nope",)      ->  M.__match_args__ == ('x',)
```

Taking your first branch rather than the second, because a header that says "methods are kept except collisions" and leaves `__slots__` unmentioned is the thing that misled you: the sentence is now scoped to methods *and* points at `TestBindingsSalixOwns`, which pins both replacements. A prose scope that no test enforces is how this file got its previous three wrong claims.

**2 — `__init__` is not a dunder salix writes.** Right. `__repr__` is a mixin binding the body's rebinding beats in the MRO; the constructor is a `tp_vectorcall` that `defines_own_init` swaps for `PyType_GenericNew`. Same observable, two mechanisms, and the docstring now separates them.

The `set_field` half is the more useful catch — as written, the test read as though `set_field` were a stylistic choice. It is not, and there is now a test saying so rather than a sentence:

```
def __init__(self, x): self.x = x
-> TypeError: I object does not support attribute assignment
```

`test_a_body_init_cannot_assign_through_the_frozen_wall`. A reader who copies the natural spelling hits exactly that, and the suite now fails if it ever stops being true.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for unit tests covering methods on a struct after checking that users can add them the way they would to a dataclass, and I am iterating this PR against the automated reviewer. This is round 3.

JPHutchins:
Both taken. `camas check` green at 25/25 on `1090be4`.

**1 — duplicate coverage.** Correct, and I should have looked. `test_custom_init.py:101` already pins the identical scenario, and lines 114-123 already pair it with the `set_field` companion I was adding narrative for. Deleted; `test_a_body_repr_and_init_displace_the_generated_ones` now points at that file for both halves.

**2 — the plain-dataclass parity claim.** Wrong, and worse than wrong in the direction that matters — a plain dataclass does not discard the method, it promotes it:

```
@dataclass
class DC:
    x: int
    def x(self): return "method"

DC.x                         ->  <function DC.x>
DC.__dataclass_fields__['x'] ->  default=<function DC.x>
DC()                         ->  DC(x=<function DC.x>)
DC(1).x                      ->  1        # instance attribute shadowing, not a descriptor
```

`dataclass(slots=True)` gives a `member_descriptor` and NamedTuple a `_tuplegetter`, so those two clauses held. The docstring now names those two and says what the plain dataclass does instead, which is the more useful sentence anyway: salix's collision loses the method, and a plain dataclass's silently constructs instances holding it.

You are right to hold this file to that standard specifically. Four claims in it have now been wrong, all in the same direction: prose asserting parity or completeness that the measurement did not cover.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for unit tests covering methods on a struct, and I am iterating this PR against the automated reviewer. This is round 4.

JPHutchins:
Finding 1 is right, and it makes last round's correction wrong too. Five taken, one declined. `camas check` green at 25/25 on `c0eea7b`.

**1 — the collision is not a discard.** Built it:

```
Collide.x                    <member 'x' of 'Collide' objects>
Collide.__struct_defaults__  (<function Collide.x>,)
Collide()                    BUILT: x = <function Collide.x>
Collide(1).x                 1
```

`append_declared` reads the class-body value bound to the field name as that field's default, and a `def` is a class-body value bound to a name. So a **required field silently becomes optional** and its default is a function; `@property` stores the `property` object the same way. `Collide(1).x == 1` was an assertion that could not see any of it.

And it inverts what I told you last round. I moved "promotes it to the field's default" onto a plain dataclass and *away* from salix — salix does it too, and unlike a dataclass it has a defaults table for the wrong value to leak into as well. That is the second time the parity claim in this file has been backwards.

`TestNameCollisions` now pins both halves separately, and **#54 is retitled and corrected** — I had filed it as "silently discarded", which is the smaller half.

**3, 4 — assertions that could not fail on value.** Both taken.

| was | now |
| --- | --- |
| `to_bytes() == HEADER.pack(258, 7, 1)` | `== b"\x02\x01\x07\x01"` |
| `header_size() == HEADER.size` | `== 4` |

The `to_bytes` one is the better catch: written out, the assertion depends on the three fields reaching the method *in the right order*, which re-deriving from the method's own expression could never test.

**5 — the shared cache entry.** Taken, and it deserved a test rather than a docstring line: `functools.cache` keys on `self`, a struct hashes by field values, so `Computed(2)` and a second `Computed(2)` are one key. `test_that_cache_is_shared_by_every_equal_instance` builds two, asserts `first is not second`, and asserts one call.

**6 — the `type: ignore` comments.** Taken, removed. You are right that pyright flags them — though not in CI, since the gate points pyright at `tests/typing` only. Confirmed by running it directly:

```
tests/test_methods.py:260:36 - error: Unnecessary "# type: ignore" comment (reportUnnecessaryTypeIgnoreComment)
```

An inert suppression that names a mypy-only code is worse than none.

**2 — the `__post_init__` duplicate.** Declined, narrowly. It is the same *mechanism* as `test_set_field_derives_a_frozen_field_on_every_interpreter`, but the assertion here is the caching claim — that the value is computed **once**, which the `calls` counter checks and the sibling does not. The docstring now points at test_post_init.py for the mechanism so a reader is not left thinking this is where `set_field` is pinned. If you still read it as redundant next round I will drop it; I took the identical finding last round and want to be consistent about which half is new.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for unit tests covering methods on a struct and issues opened for the gaps, and I am iterating this PR against the automated reviewer. This is round 5.

JPHutchins:
All three taken, and two of them are the same mistake: I corrected a claim in one place and left it standing in another. `camas check` green at 25/25 on `a21e559`.

**2 — the module docstring.** Right, and it was four lines from the top of the file whose `TestNameCollisions` docstring I had just rewritten to say the opposite. Fixed: "a name that collides with a field, which is not kept and not simply lost either: it becomes that field's default."

**1 — `test_a_subclass_inherits_them`.** Same tautology, same commit, one test over. I fixed `test_a_staticmethod_needs_no_instance` and did not grep for the pattern. Changing `"<HBB"` to `"<HHI"` would have moved both sides together and passed. Now `== 4`, matching the sibling.

**3 — the caching test.** Taken this time, and your framing is what changed my mind: the docstring claimed the value is "computed once, at construction, and read from a slot afterwards", and the test read the field **once**, so the second half was never exercised. It now reads it three times against one `__post_init__` call — which is the only assertion in it that `test_post_init.py` does not already make, so the duplication objection and the coverage gap resolve together.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for unit tests covering methods on a struct and issues opened for the gaps, and I am iterating this PR against the automated reviewer. This is round 6.

JPHutchins:
All four taken, including the one I declined twice. `camas check` green at 25/25 on `6fe1edb`.

**2 — the caching test.** You were right and I was wrong on the specifics both times. `test_every_field_is_written_before_it_runs` does pin exactly-once:

```python
assert seen == [(1, "default"), (1, 2)]
```

A list equality fails if `__post_init__` runs zero times, twice, or more — so `len(calls) == 1` was never the unique assertion I claimed it was. And the three-read version I added last round was decorative, for the reason you give: a slot cannot recompute.

Deleted. What survives is a sentence in `TestCaching`'s docstring pointing at the file that does the pinning. Three rounds, two wrong defences — the pattern was that I kept looking for what my test *could* claim rather than checking what the sibling already claimed.

**1 — `tag=9` unasserted.** Taken, `assert tagged.tag == 9`. The one piece of data unique to that test was the one thing it did not check.

**3 — the self-consistent round trips.** Taken for `from_bytes`, which now takes the same independent `b"\x02\x01\x07\x01"` that `to_bytes` is checked against, so each direction is pinned on its own rather than against the other. The round-trip test stays — it does pin the classmethod reaching `__eq__` with a real instance — but its docstring now says that is *all* it pins, since composing pack with unpack holds whatever `HEADER` is.

**4 — the duplicated fixture.** Taken. `counted_cache()` builds it, and it has to be per-test rather than module-level because `functools.cache` outlives the instances that fill it — which is the same lifetime cost the tests are about.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for unit tests covering methods on a struct and issues opened for the gaps, and I am iterating this PR against the automated reviewer. This is round 7.

JPHutchins:
All five taken. `camas check` green at 25/25 on `d424817`.

**2 — `__match_args__` is not unconditionally salix's.** Right, and I overstated it in two places. Measured:

```python
class Matched(Struct, match_args=False):
    x: int
    __match_args__ = ("nope",)   -> ('nope',)   the body keeps it
    __slots__ = ("extra",)       -> ('x',)      salix takes it anyway
```

`set_match_args` returns without touching the namespace when the option is off, so opting out means salix writes *none* rather than writes an empty one. Both existing tests only ran the default. Header corrected, and `test_opting_out_of_match_args_leaves_the_body_its_own` pins the asymmetry.

**4 — untested stdlib parity in the cached_property docstring.** Taken, and this is the right standard for this file specifically: every parity claim it has made in prose has been wrong at least once. All three are tests now — `dataclass(slots=True)`, `NamedTuple` including the pre/post-3.13 message split, and the plain dataclass that succeeds. The docstring says they are asserted below rather than restating them.

**1 — `kind=7` unasserted.** Same pattern as `tag=9`, one test along, and I fixed that one last round without grepping for the rest. Now `(len(frame), frame.kind) == (258, 7)`.

**5 — dead `isinstance`.** Correct: `from_bytes` returns `cls(...)` and the assertion above already forces the attributes. `type(frame) is Frame` at least fails if a subclass ever comes back from it.

**3 — the imprecise cross-reference.** Taken; the test directly above formats field values and never touches `HEADER`, so it names the two that actually pin the format.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for unit tests covering methods on a struct and issues opened for the gaps, and I am iterating this PR against the automated reviewer. This is round 8.

JPHutchins:
Both taken. `camas check` green at 25/25 on `f74cadf`.

**1 — the new clause is false.** Correct, and it is the same mistake in a smaller space: I fixed an imprecision by writing something I had not checked. The test directly above the round trip *is* the `from_bytes` one, which is one of the two the sentence names.

The docstring no longer says where those tests are, only what they do. A positional claim about a test file has to be re-verified every time anything moves, and it earns nothing.

**2 — `type(frame) is Frame` is still unreachable.** Also correct, and a fair reading of what I did: I changed the assertion's form and not its substance. `from_bytes` returns `cls(...)`, `cls` is `Frame` at that call site, and the line below already forces the fields. Deleted — `test_a_classmethod_on_a_subclass_builds_the_subclass` is where the `cls` binding actually matters.

Both findings this round were nits about prose and dead assertions rather than anything the tests get wrong, which is the first time that has been true here.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for unit tests covering methods on a struct and issues opened for the gaps, and I am iterating this PR against the automated reviewer. This is round 9.

JPHutchins:
All six taken, and finding 1 found a real defect. `camas check` green at 25/25 on `a36e793`.

**1 — no test pinned a body `__eq__`/`__hash__`.** Correct, the header promised it and the file did not have it. Writing one turned up **#58**:

```
salix     : a == b -> True   a != b -> True
dataclass : c == d -> True   c != d -> False
plain     : e == f -> True   e != f -> False
```

The body's `__eq__` wins for `==`, and salix keeps binding the mixin's `__ne__` beside it — so `a == b` and `a != b` are both true. Python derives `__ne__` from `__eq__` when a class defines only `__eq__`, which is why every other construction is consistent and this one is not. `bind_hash` already computes `body_defines_eq`; `__ne__` presumably wants the same treatment `__hash__` gets.

Out of scope for a tests-only PR, so it is filed and the test pins the current behaviour with the issue named on the assertion, so a fix has to come back and change that line.

**2 — the cache docstring is backwards.** `functools.cache` holds its keys alive; it does not outlive them. The sibling test says exactly that one paragraph down ("at the cost of holding the struct alive"), so the file contradicted itself. Fixed — the reason a fresh class is needed per test is that entries one test made would answer for the next.

**3 — the byte assertion overclaimed.** Right: a method ignoring its fields and returning the same constant would pass, so it does not depend on the fields *reaching* it. What it catches is a wrong **order**, and it says that, pointing at the round trip for the rest.

**4, 5 — `Collide` and `Tagged` declared twice each.** Taken, one builder apiece.

**6 — `Custom(1)` four times for three assertions.** Taken.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for unit tests covering methods on a struct and issues opened for the gaps, and I am iterating this PR against the automated reviewer. This is round 10.

JPHutchins:
All three taken, and finding 1 is the right objection to how I recorded #58. `camas check` green at 25/25 on the new head.

**1 — pinning the bug as correct.** Agreed. Asserting `first != second` with an issue number in a comment says the broken thing is right and relies on someone reading the comment. `xfail(strict=True)` on the assertion that *should* hold is the better shape: red today for the right reason, and an unexpected-pass error the moment #58 lands — which a fix cannot walk past.

**3 — and the docstring did the file's own worst habit.** Right, and pointedly: I wrote `d424817` ("turn the last prose parity claims into assertions") two rounds ago and then made a fresh parity claim about dataclass and plain classes in prose, with neither constructed. `test_every_other_construction_derives_ne_from_eq` builds both.

**2 — `match_args=False` overstated.** Confirmed:

```
class Off(Struct, match_args=False): x: int
Off.__match_args__            ->  ()          inherited from Struct
dataclass(match_args=False)   ->  no attribute at all
```

Opting out means salix writes none into *this* namespace — which is exactly why a body binding survives it — not that the class ends up without one. The header and the test docstring both say that now, and both spellings are asserted.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for unit tests covering methods on a struct and issues opened for the gaps, and I am iterating this PR against the automated reviewer. This is round 11.

PR #65, opened 2026-08-05
ci: import the two wheels nothing has ever loaded

### LINKED ISSUE #37: 84% of every wheel is DWARF, and 14 of the 50 are never loaded before shipping

JPHutchins:
Closes #37 — this is its second half.

`tools/check_wheel.py` inspects all 50 wheels — RECORD hashes and sizes, the WHEEL tag against the filename, the payload's container and machine against the platform tag. What it cannot do is load one. That was left to the `coverage` matrix, and the matrix had no runner for two of the platforms it builds:

| wheel | who could import it before | now |
| --- | --- | --- |
| `macosx_10_13_x86_64` | nobody — `macos-latest` is Apple silicon | `macos-15-intel` × 3.10, 3.15 |
| `win_arm64` | nobody — `windows-latest` is x86_64 | `windows-11-arm` × 3.11, 3.15 |

Both are cross-compiled by zig, which is the half of the build with nobody standing behind it, and both went to PyPI inspected and never run.

**Why `windows-11-arm` asks for 3.11 rather than 3.10.** python-build-standalone has no Windows ARM64 build below 3.11, so there is no `cp310-cp310-win_arm64` wheel for that leg to import:

```console
$ ls result-wheels/*win_arm64* | cut -d- -f3 | sort -u
cp311 cp312 cp313 cp314 cp315
```

The wheel set is not square and the matrix should not pretend it is.

**This cannot be exercised locally**, which is why the issue kept it out of #42 and why it needs its own PR — the check is whether these four legs go green here.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked that the open salix issues keep being worked through; this is the half of #37 that was deliberately deferred because it can only be verified in CI.


JPHutchins:
**`macos-13` is retired — swapped for `macos-15-intel`.** `3ec92a6`.

The two legs 98fa254 added could never be scheduled. On run [30965520951](https://github.com/JPHutchins/salix/actions/runs/30965520951) they sat `queued` from 01:17:40Z while the twelve legs around them finished:

```
coverage (macos-13, 3.10)        queued      <- 80 minutes, never started
coverage (macos-13, 3.15)        queued      <- 80 minutes, never started
coverage (windows-11-arm, 3.11)  success
coverage (windows-11-arm, 3.15)  success
...twelve others                 success
```

[actions/runner-images](https://github.com/actions/runner-images) lists no macOS 13 image; the x86_64 macOS runners now are `macos-15-intel` and `macos-26-intel`, both standard and free for public repositories. This takes the older one — the wheel is tagged `macosx_10_13_x86_64`, so either is far past its floor.

A label GitHub accepts but cannot schedule is worse than no leg at all: the run never completes, so nothing waiting on it reports. I cancelled the stuck run.

~~The other new leg is unaffected — `windows-11-arm` passed on both 3.11 and 3.15, which is the first time anything has imported a `win_arm64` salix wheel.~~

**Correction (see below):** that last sentence was wrong. Those legs passed while importing `win_amd64` under emulation — uv resolves a bare `--python 3.15` on that runner to the x86_64 build. Fixed in `75898ab`; nothing had imported a `win_arm64` wheel at the time of this comment.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, while driving her open PRs toward merge.

JPHutchins:
**Finding 1 is answered by a run that had not happened when you looked, and finding 2 is real. Finding 3 was right, and chasing it found something worse than all three.** `75898ab`.

## 1 — both `macos-15-intel` legs have now run, and they went green

[Run 30969907429](https://github.com/JPHutchins/salix/actions/runs/30969907429) on `3ec92a6`:

```
coverage (macos-15-intel, 3.10)   pass   48s
coverage (macos-15-intel, 3.15)   pass   46s
```

And they imported what they were added for:

```
Downloading cpython-3.10.20-macos-x86_64-none (download)
Downloading cpython-3.15.0b4-macos-x86_64-none (download)
```

## The one you found by accident: the ARM legs were importing `win_amd64`

Checking finding 2 against the logs turned up that the `windows-11-arm` legs **passed twice without importing a `win_arm64` wheel**:

```
Downloading cpython-3.11.15-windows-x86_64-none (download)
Downloading cpython-3.15.0b4-windows-x86_64-none (download)
```

uv resolves a bare `--python 3.15` on that runner to the x86_64 build, Windows ARM64 runs it under emulation, and an emulated interpreter installs `win_amd64` — the wheel `windows-latest` already covers. Half of #37's second half was green and empty. I have struck the claim I made [above](https://github.com/JPHutchins/salix/pull/65#issuecomment-5186894339).

The builds do exist and the floor matches the wheel set:

```console
$ uv python list --all-versions --all-platforms --all-arches | grep -c windows-aarch64
42
$ ... | grep -v freethreaded | cut -d. -f1,2 | sort -uV
3.11  3.12  3.13  3.14  3.15
```

So the leg names the architecture now, and CI is the check:

```
--python cpython-3.11-windows-aarch64
--python cpython-3.15-windows-aarch64
```

## 2 — camas builds from its sdist on `macos-15-intel`, and that is what the run shows

Confirmed on both counts. No camas release ships either wheel — 0.1.29 is the latest, and .27, .28 and .29 all publish the same six platform tags:

```
macosx_11_0_arm64  manylinux2014_{x86_64,aarch64}  musllinux_1_2_{x86_64,aarch64}  win_amd64
```

The macOS Intel leg therefore builds it, and the log says how long that took:

```
02:49:07.882  Building camas==0.1.29
02:49:12.443     Built camas==0.1.29        <- 4.6 s
02:49:12.481  Installed 1 package in 4ms
```

`windows-11-arm` did not build it — 54 ms, wheel — because that job's own `uv sync` runs the emulated x86_64 interpreter, which is the same thing the section above is about. Once the leg is native, that one will build from sdist too.

There is nothing to pin to, so this is accepted for now and belongs upstream: filed as [JPHutchins/camas#287](https://github.com/JPHutchins/camas/issues/287). A 4.6-second source build on a leg that takes 48 seconds is a cost worth paying to import a wheel nobody had ever loaded; it is not a cost worth keeping once camas can ship the two wheels.

## 3 — right, and fixed

"The first four" were the six ubuntu legs. The comment names `macos-latest` and `windows-latest` now.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, answering the first review round on this PR.

JPHutchins:
**Both ARM legs are green on `75898ab`, and this time they ran on ARM.** [Run 30971847823](https://github.com/JPHutchins/salix/actions/runs/30971847823):

```
coverage (windows-11-arm, cpython-3.11-windows-aarch64)  success  | interpreter: cpython-3.11-windows-aarch64
coverage (windows-11-arm, cpython-3.15-windows-aarch64)  success  | interpreter: cpython-3.15-windows-aarch64
```

The interpreter line is read out of each job's log, not assumed — it is the same line that gave the emulated x86_64 away on the previous head.

**On what is measured and what follows from it:** the logs name the interpreter, not the wheel. That an aarch64 CPython imported salix from `--find-links ../result-wheels` leaves one possibility, since a `win_amd64` payload cannot load into an ARM64 process and the sdist path is closed on that runner — MSVC cannot build this source, which is why the leg installs rather than compiles. So the wheel is the `win_arm64` one, by elimination from a measured interpreter rather than by reading its name.

I considered adding an in-process assertion that the installed wheel's tag matches the machine, and did not: on the emulated leg `platform.machine()` reports `AMD64` and the `win_amd64` tag matches it, so the test would have passed on the bug. What closes it is the explicit request — uv cannot silently answer `cpython-3.15-windows-aarch64` with another architecture, where it can and did answer `3.15` with one.

All four new legs now do what #37 asked:

| wheel | leg | interpreter it ran |
| --- | --- | --- |
| `macosx_10_13_x86_64` | `macos-15-intel` × 3.10, 3.15 | `cpython-*-macos-x86_64` |
| `win_arm64` | `windows-11-arm` × 3.11, 3.15 | `cpython-*-windows-aarch64` |

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, reporting the CI result of the fix in the comment above.

PR #70, opened 2026-08-08
fix: read a slot under the lock the writer takes

### LINKED ISSUE #46: repr, hash and == read slots unsynchronised, so a concurrent write segfaults on a free-threaded build

JPHutchins:
Closes #46.

`#7` made the write side safe by routing `set_field` through `PyMember_SetOne`,
which holds a critical section on the instance over the store and defers the
release past the end of it. The read side was never touched: every reader loaded
a slot and then increffed what it found, which is two steps with a window
between them, and on a free-threaded build a concurrent write frees the pointer
inside it.

`struct_slot_ref` now takes the reference under a critical section on the
instance, and every reader goes through it — `repr.c`, `hash.c` and both
comparison paths in `compare.c`. One function rather than a section at each of
four sites, so a fifth reader cannot forget it.

## The measurement

3.14t, four writer threads and three readers of one kind on one shared
`frozen=False` struct, 40k rounds, three runs each. The control is CPython's own
member descriptor, which does hold the lock:

| reader | exits |
| --- | --- |
| `descriptor` (`shared.value`, i.e. `PyMember_GetOne`) | 0, 0, 0 |
| `repr` | 134, 139, 134 |
| `==` | 139, 134, 134 |
| `hash` | 0, 0, 0 — reads the same way, did not manifest |

The regression test is `SIGABRT` 3/3 without the critical section and clean with
it.

## Not the issue's preferred option 2

<details>
<summary>Routing reads through <code>PyMember_GetOne</code> would inherit the guarantee rather than restate it, which is the right instinct — but a NULL slot is an ordinary state here.</summary>

`del m.x` on a mutable struct leaves a slot NULL, and `repr` renders it
`M(x=<unset>)`. `Py_T_OBJECT_EX` answers a NULL slot with an `AttributeError`,
so option 2 turns an ordinary answer into raise-then-clear at every reader, and
needs `PyMemberDef.name` populated or CPython formats the message from a NULL
pointer.

So `struct_slot_ref` hands NULL back — `repr` is the caller that wants to see it
— and `struct_slot_ref_or_none` is the total version `hash` and `==` use, which
is what `struct_slot_or_none` was before this.

</details>

<details>
<summary>Why the critical section is a macro pair rather than a function.</summary>

A critical section is a lexical scope holding a frame the interpreter links into
a per-thread stack, so no function can open one and return with it still open —
the same reason `owned.h` is macros. `Py_BEGIN_CRITICAL_SECTION` arrived in 3.13
and expands to a bare scope on a build with the GIL, so `types.h` carries a shim
beside the `SLOT_MEMBER_TYPE` one and nothing changes on a GIL-ful build.

</details>

<details>
<summary>The test took three tries to mean anything — it passed on the unfixed build twice.</summary>

Once because the reader's per-iteration assertion was wide enough to hide the
window between the load and the incref, and once because the comparison found an
early out. The version here asserts every 1000th pass rather than every pass,
runs `ITERATIONS * 5` rounds, and compares against a struct holding a bare
`object()` so the walk can never short-circuit and can never be equal by
accident. Survival is the assertion, because a segfault takes pytest with it —
same shape as the write-path tests above it.

</details>

Writers are unchanged: `construct.c:112` runs before the object escapes, and
`write_slot` keeps `PyMember_SetOne`.

> [!WARNING]
> **LLM Disclosure**
>
> each PR to be iterated with the automated reviewer; this is #46, which was
> written and held while #48 was in flight.

### DIFF
--- a/src/compare.c
+++ b/src/compare.c
@@ -74,8 +74,9 @@ static PyObject * ordering_result(PyObject * const self, PyObject * const other,
 	}
 
 	for (Py_ssize_t i = 0; i < self_type->struct_field_count; ++i) {
-		PY_OWNED(mine, Py_NewRef(struct_slot_or_none(self_type, self, i)));
-		PY_OWNED(theirs, Py_NewRef(struct_slot_or_none(other_type, other, i)));
+		struct slot_pair const pair = struct_slot_pair_ref(self_type, self, other_type, other, i);
+		PY_OWNED(mine, pair.mine);
+		PY_OWNED(theirs, pair.theirs);
 		int const equal = PyObject_RichCompareBool(mine, theirs, Py_EQ);
 
 		if (equal == COMPARISON_ERROR) {
@@ -114,8 +115,9 @@ static enum comparison names_equal(
 	);
 }
 
-/* Borrowed slots are sound here and not in ordering_result: the comparison is
- * the last use of either pointer. */
+/* Owned for the same reason ordering_result's are: the equality probe runs
+ * arbitrary Python, which can rebind either slot, and on a free-threaded build
+ * can free what a borrowed pointer names. */
 static enum comparison values_equal(
 	StructType const * const self_type,
 	PyObject * const self,
@@ -123,11 +125,10 @@ static enum comparison values_equal(
 	PyObject * const other
 ) {
 	for (Py_ssize_t i = 0; i < self_type->struct_field_count; ++i) {
-		int const equal = PyObject_RichCompareBool(
-			struct_slot_or_none(self_type, self, i),
-			struct_slot_or_none(other_type, other, i),
-			Py_EQ
-		);
+		struct slot_pair const pair = struct_slot_pair_ref(self_type, self, other_type, other, i);
+		PY_OWNED(mine, pair.mine);
+		PY_OWNED(theirs, pair.theirs);
+		int const equal = PyObject_RichCompareBool(mine, theirs, Py_EQ);
 
 		if (equal != COMPARISON_EQUAL) {
 			return equal;
--- a/src/hash.c
+++ b/src/hash.c
@@ -44,9 +44,7 @@ Py_hash_t Struct_hash(PyObject * const self) {
 		return HASH_ERR;
 	}
 
-	for (Py_ssize_t i = 0; i < type->struct_field_count; ++i) {
-		PyTuple_SET_ITEM(values, i, Py_NewRef(struct_slot_or_none(type, self, i)));
-	}
+	struct_slots_ref_or_none_into(type, self, values);
 
 	if (Py_EnterRecursiveCall(" while hashing a struct") != 0) {
 		return HASH_ERR;
--- a/src/repr.c
+++ b/src/repr.c
@@ -61,7 +61,7 @@ static PyObject * field_repr(
 	PyObject * const self,
 	Py_ssize_t const index
 ) {
-	PyObject * const value = *struct_slot(type, self, index);
+	PY_OWNED(value, struct_slot_ref(type, self, index));
 
 	PY_OWNED(rendered, value != NULL ? PyObject_Repr(value) : PyUnicode_FromString("<unset>"));
 
--- a/src/types.h
+++ b/src/types.h
@@ -87,19 +87,127 @@ static inline PyObject * * struct_slot(
 }
 
 /*
- * A slot stays NULL until something writes it, which is observable on a
- * half-built struct. Reading it as None keeps hash and == total instead of
- * making each one re-derive the guard; repr wants to see the NULL, so it
- * reads the slot directly.
+ * Py_BEGIN_CRITICAL_SECTION arrived in 3.13, and expands to a bare scope on a
+ * build with the GIL -- which is every build below it.
+ *
+ * A macro because a critical section is a lexical scope holding a stack frame
+ * the interpreter links into a per-thread stack; no function can open one and
+ * return with it still open. Same reason owned.h is macros.
  */
-static inline PyObject * struct_slot_or_none(
+#if PY_VERSION_HEX < 0x030D0000
+#	define STRUCT_BEGIN_CRITICAL_SECTION(object) {
+#	define STRUCT_END_CRITICAL_SECTION() }
+#	define STRUCT_BEGIN_CRITICAL_SECTION2(first, second) {
+#	define STRUCT_END_CRITICAL_SECTION2() }
+#else
+#	define STRUCT_BEGIN_CRITICAL_SECTION(object) Py_BEGIN_CRITICAL_SECTION(object)
+#	define STRUCT_END_CRITICAL_SECTION() Py_END_CRITICAL_SECTION()
+#	define STRUCT_BEGIN_CRITICAL_SECTION2(first, second) Py_BEGIN_CRITICAL_SECTION2(first, second)
+#	define STRUCT_END_CRITICAL_SECTION2() Py_END_CRITICAL_SECTION2()
+#endif
+
+/* One field from each of two instances, for the readers that walk them in
+ * pairs. */
+struct slot_pair {
+	PyObject * mine;
+	PyObject * theirs;
+};
+
+/*
+ * A new reference to a field, taken under the same lock the writer takes.
+ *
+ * #7 made the write side safe by going through PyMember_SetOne, which holds a
+ * critical section on the instance over the store and defers the release past
+ * the end of it. Every reader here loaded the slot and then increffed what it
+ * found, which is two steps with a window between them: on a free-threaded
+ * build a concurrent write frees the pointer inside that window, and repr and
+ * == segfault. Measured, 3.14t, four writers and three readers of one kind:
+ * `repr` exited 134/139/134 and `==` 139/134/134, while the same loop reading
+ * through CPython's own member descriptor survived every time. #46.
+ *
+ * Every acquisition is in this file, so a fifth reader cannot forget one. The
+ * macros are a bare scope on a build with the GIL, so all three of these are
+ * the same load and incref they always were there.
+ *
+ * What a reader gets is per-slot atomicity, not a snapshot of the struct: a
+ * write landing between two of repr's fields renders a pair the object never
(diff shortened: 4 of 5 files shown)


JPHutchins:
Round 1: all four taken, and the granularity one measured rather than argued.

## 🔵 `compare.c:117` — stale comment

Correct, and the comment said the opposite of what the code now does. Replaced
with the reason the references are owned, which is the same reason
`ordering_result` gives.

## 🔵 `types.h:132` — N acquisitions per read, and no snapshot

Measured before changing anything. 3.14t, uncontended, median of 7 × 20k
`timeit` samples, an eight-field frozen struct:

| | main (no lock) | round 1 (per field) | now |
| --- | --: | --: | --: |
| `hash(eight)` | 77.5 ns | 111.4 ns | **82.2 ns** |
| `eight == other` | 54.6 ns | 134.5 ns | **110.6 ns** |
| `repr(eight)` | 974.7 ns | 1055.8 ns | 1032.0 ns |
| `hash(one)` | 55.8 ns | 60.6 ns | 61.0 ns |

So the per-field lock was costing `hash` 44% and `==` 146%, which is worth the
two changes it can be reduced by:

- **`hash` now takes one acquisition for the whole tuple** —
  `struct_slots_ref_or_none_into`. Its loop is `PyTuple_SET_ITEM` over a load
  and an incref and never calls back into Python, so the section can span it.
  That recovers 29 of the 34 ns and makes `hash` the one reader that genuinely
  does get a snapshot.
- **the comparison paths take one `Py_BEGIN_CRITICAL_SECTION2(self, other)` per
  field-pair** instead of two separate sections —
  `struct_slot_pair_ref`. 134.5 → 110.6 ns.

<details>
<summary>Why the comparison loop is not hoisted the rest of the way, and why <code>repr</code> is not hoisted at all.</summary>

Both call arbitrary Python per field — `PyObject_RichCompareBool` and
`PyObject_Repr`. A critical section held across that is not a snapshot: CPython
suspends a thread's sections when it detaches, so the guarantee lapses exactly
where it would be needed, while the code would read as though it held.

The remaining alternative is to snapshot every field into a buffer under one
section and compare afterwards. That is an allocation per `==` — and on the 35
GIL-ful wheels of the 50, where the macros are a bare scope and there is nothing
to optimise, it would be a pure regression. 8 acquisitions at ~7 ns is cheaper
than two tuples.

So `struct_slot_ref` and `struct_slot_pair_ref` promise per-slot atomicity and
say so in the comment; `struct_slots_ref_or_none_into` promises a snapshot and
says why it can.

</details>

## 🔵 `test_free_threading.py:164` — the changed `hash` path was untested

Correct. The reader loop hashes now, and it needed a second shared struct to do
it: a `frozen=False` struct is unhashable, so the race the finding describes
needs a **frozen** one written through `set_field` — which is precisely the
combination `hash.c:48` is reachable in.

## ⚪ `test_free_threading.py:173` — odd `THREADS`

Fixed by slicing to the thread count rather than doubling half of it, so an odd
`THREADS` is one role short of a writer rather than one thread short of a role:

```python
roles = iter(([write, read] * THREADS)[:THREADS])
```

## The regression test still bites

Re-scored after the rework, since the readers changed shape. Sections stripped
from all three helpers, three runs each, distinct `.so` hashes both ways:

```
unlocked   exit=139  exit=134  exit=134
locked     exit=0    exit=0    exit=0
```

`camas check` 25/25 and `nix flake check` green, including the Windows cross
builds.

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Round 2: both taken, and the minor came back with a result I did not expect —
worth stating plainly rather than shipping the coverage as though it holds.

## 🔵 the ordering path had no concurrent coverage

Right that nothing reached `ordering_result`. It has a test now — and it is the
weaker of the four, which the test says so itself.

**Adding `<` to the existing reader loop catches nothing.** Reverting only
`ordering_result`'s pair read to the borrowed pattern survived 3/3 there. The
reason is the loop it was added to: the three readers beside it each take the
instance lock, which serialises them against the writer and leaves almost no
window for the one that does not.

**A dedicated test at the same density does not catch it either.** Four writers,
four readers doing nothing but `floor < shared`, borrowed pair read:

```
  run 1 exit=0
  run 2 exit=0
  run 3 exit=0
```

and 3/3 again at five times the rounds. So I cannot claim this test pins the
crash, and it does not say it does.

<details>
<summary>This is the same result <code>hash</code> gave, which is why it is a shape rather than a surprise.</summary>

The measurement in the PR description has `hash` at `0, 0, 0` unfixed while `==`
was `139, 134, 134` — and `hash` reads exactly as `==` does. `ordering_result`
is one racing load per call behind a `names_equal` tuple comparison and the
richcompare dispatch, which is apparently not often enough to land in a
two-instruction window.

The path is still fixed, and the crash is still pinned — by the test above,
through `struct_slot_pair_ref`, which both comparison paths now share. What the
new test pins is that ordering keeps *working* under concurrent writes, and it
would catch a functional regression in that loop. That is the honest scope.

</details>

## ⚪ the "two readers" sentence in `types.h`

Correct, and it was left over from a draft where the comparison loop was also
going to be hoisted. Only `struct_slots_ref_or_none_into` keeps its loop in C,
so only `hash` gets a snapshot. Reworded to say exactly that, and to name the
two calls — `PyObject_Repr` and `PyObject_RichCompareBool` — that make hoisting
the other two read as a guarantee without being one.

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Merged at round 3 on the convergence signal: **score 1 ≤ 1 — converged**, CI
green on `8f19cc6`, all fifteen legs.

I had been judging convergence by eyeballing finding counts and had not read the
badge at all — three of the four PRs in this batch had already reached it, and
answering a converged round on #72 came back *iterating*, so it cost ground.
Filed as JPHutchins/code-review#141: the signal is unreachable from
`findings-json`, which is the only channel the sticky's own `AGENTS: STOP` line
points at.

## The one finding merged over

Round 3's remaining minor is right that `assert ordered is True` in
`test_a_shared_ordered_struct_is_safe_to_compare_while_another_thread_writes_it`
cannot fail on the race — `floor` is `(-1, -1)` and the writer only ever stores
`(i, i)`.

It is not there to. It is the same device the sibling test names in its own
docstring — *"the value assertions are there so that a reader which quietly
stopped reading would fail rather than pass"* — and the straddling writer the
recommendation suggests would weaken exactly that, since the reader then cannot
know which of the two values is current and can assert nothing about the result.
The docstring already says the test does not pin the crash; what it does not say
is that the assertion is constant by design.

That last sentence is the fair part of the finding and it is a docstring, on a
test whose limitation is already written down, in a PR the reviewer scores as
converged. It goes with the next change to `test_free_threading.py` rather than
another round here.

> [!WARNING]
> **LLM Disclosure**
>
> why this merged with one open minor.

PR #71, opened 2026-08-08
fix: read the layout base from the base CPython chose

### LINKED ISSUE #9: find_struct_base takes the first struct base, dropping fields and defeating frozen inheritance

JPHutchins:
Closes #9.

`find_struct_base` returned the first `StructMeta` in `bases`, and every
question was then asked of that one object. A fieldless struct base standing in
front of one with fields answered them all wrongly:

```python
class C(Fieldless, WithFields):
    c: int

C.__struct_fields__         # ('c',)  -- a and b gone
C(1).a                      # AttributeError
```

The slots for `a` and `b` still existed; nothing reached them, so `repr`, `==`
and `hash` silently ignored two thirds of the object. Reversing the base order
fixed it, which was the tell.

It also defeated `frozen`, which is the part that matters.
`class M(Fieldless, WithFields, frozen=False)` was accepted, so a mutable
subclass of a frozen struct existed and could be handed to anything holding a
`WithFields`-typed reference — exactly what `options.h` says `frozen` is there
to prevent.

## The base to read is the one CPython chose

`tp_base` is where CPython placed the new slots, so it is where their offsets
are measured from. Among struct bases that is **the one with the most fields**,
not the widest one.

<details>
<summary>Why not <code>tp_basicsize</code> — it is wrong on 3.10 and 3.11, and the two rules agree everywhere else.</summary>

CPython discounts `__weakref__` and `__dict__` when it compares layouts. Below
3.12 a weakref slot still grows `tp_basicsize`, so a `weakref=True` fieldless
base measures exactly as wide as a one-field base and wins a comparison on size
while CPython gave `tp_base` to the other:

```
3.10.20   Fieldless=16/0  FieldlessWeak=24/0  One=24/1  Two=32/2
3.12.13   Fieldless=16/0  FieldlessWeak=16/0  One=24/1  Two=32/2
```

Over every buildable combination of two and three bases drawn from eight shapes
— fieldless, fieldless with weakref, a fieldless subclass of that, one field,
one field with weakref, a fieldless subclass of the one-field base, two fields,
and two reached by extending the one-field base — each candidate rule scored
against `__base__` itself:

| | buildable | `first` wrong | `widest` wrong | `most fields` wrong |
| 3.10.20 | 155 | 85 | 25 | 0 |
| 3.12.13 | 155 | 85 | 0 | 0 |
| 3.14.6 | 155 | 85 | 0 | 0 |

</details>

<details>
<summary>The frozen constraint does not need its own walk of <code>bases</code>, and it had one.</summary>

The first version of this answered `frozen` of *any* struct base rather than of
the layout one, on the reasoning that each base made the promise separately.
With the layout base fixed that is redundant rather than safer: the base with
the most fields has fields whenever any base does. The mutation says so — the
extra walk was removed only after removing it failed to break a test on either
3.10 or 3.14, and the case that would distinguish the two (a weakref-only base
tying on size) is the one the field-count rule already gets right.

</details>

<details>
<summary>Each candidate rule is a mutation of this commit, and the tests are scored against it.</summary>

```
first-match-layout-base  [3.10]: failed=9    [3.14]: failed=9
widest-by-basicsize      [3.10]: failed=4    [3.14]: failed=0
```

13 tests. `widest-by-basicsize` correctly fails nothing on 3.14, where it and
the field-count rule agree; on 3.10 the four it fails are the two weakref cases
and two of the parametrised `__base__` agreement checks.

</details>

## What is not here

The `__slots__` half of #12 was written alongside this and has been taken back
out. Refusing a body `__slots__` in `StructMeta_new` also refuses salix's own:
`create_class` hands the build to a winning metatype whose Python `__new__`
calls back into `StructMeta_new` with the namespace `build_class_namespace`
already wrote `__slots__` into, and three tests turn red on the round trip. The
refusal has to tell those two apart, which is its own change against #48's
hand-off. #12 stays open for it and for the `__new__` half, which is blocked on
#13.

> [!WARNING]
> **LLM Disclosure**
>
> each PR to be iterated with the automated reviewer; this is #9, held while #48
> was in flight and rewritten against the `create_class` it left behind.

### DIFF
--- a/src/meta.c
+++ b/src/meta.c
@@ -35,6 +35,9 @@ static PyObject * build_struct_class(
 	PyObject * keywords
 );
 static StructType * find_struct_base(PyObject * bases);
+static StructType * find_behaviour_base(PyObject * bases);
+static struct options inherited_options(PyObject * bases, StructType const * behaviour);
+static bool any_struct_base_is_mutable(PyObject * bases);
 static bool has_weakref_slot(StructType const * base);
 static struct options base_options(StructType const * base);
 static PyObject * build_class_namespace(
@@ -43,6 +46,8 @@ static PyObject * build_class_namespace(
 	PyObject * new_names,
 	struct options options,
 	StructType const * base,
+	struct options inherited,
+	bool frozen_across_bases,
 	bool body_defines_eq,
 	bool inherits_body_eq
 );
@@ -52,6 +57,7 @@ static enum result apply_options(
 	PyObject * namespace,
 	struct options options,
 	struct options inherited,
+	bool frozen_across_bases,
 	bool body_defines_eq,
 	bool inherits_body_eq
 );
@@ -220,7 +226,12 @@ static PyObject * build_struct_class(
 	PyObject * const original_namespace,
 	PyObject * const keywords
 ) {
-	struct options const inherited = base_options(base);
+	/* The behaviour base rather than the layout one: every dunder a struct base
+	 * carries is resolved by the MRO, which takes the first of them, so reading
+	 * the options off anything else makes the record and the behaviour two
+	 * different answers. */
+	StructType const * const behaviour = find_behaviour_base(bases);
+	struct options const inherited = inherited_options(bases, behaviour);
 	struct options_request const request =
 		options_read(keywords, inherited, base != NULL && base->struct_field_count > 0);
 
@@ -236,11 +247,12 @@ static PyObject * build_struct_class(
 
 	/* Read before build_class_namespace rebinds anything: afterwards every
 	 * comparison name is present whether the body wrote one or salix did. An
-	 * inherited one survives only if this body leaves equality alone. */
+	 * inherited one survives only if this body leaves equality alone -- and it
+	 * is the behaviour base's, because that is the __eq__ the MRO finds. */
 	bool const body_defines_eq = PyDict_GetItemString(original_namespace, "__eq__") != NULL;
 	bool const inherits_body_eq = (
-		base != NULL &&
-		base->struct_resolves_body_eq &&
+		behaviour != NULL &&
+		behaviour->struct_resolves_body_eq &&
 		request.options.eq == inherited.eq
 	);
 
@@ -252,6 +264,8 @@ static PyObject * build_struct_class(
 			plan.new_names,
 			request.options,
 			base,
+			inherited,
+			request.options.frozen && any_struct_base_is_mutable(bases),
 			body_defines_eq,
 			inherits_body_eq
 		)
@@ -289,9 +303,60 @@ static PyObject * build_struct_class(
 	return (PyObject *) struct_class;
 }
 
-/* Find the (single) struct base among ``bases``. A fieldless one still carries
- * the options a subclass inherits, so it counts. */
+/*
+ * The struct base this class extends the layout of: the one with the most
+ * fields, which is what CPython settles on as tp_base and so where the slots
+ * and their offsets come from. Taking the first match instead let a fieldless
+ * base stand in front of one with fields, and every field of the second went
+ * missing.
+ *
+ * Fields rather than tp_basicsize, because CPython discounts __weakref__ and
+ * __dict__ when it compares layouts and salix's fields are the only other slots
+ * a struct base adds. Below 3.12 a weakref slot still widens the type, so a
+ * `weakref=True` fieldless base measures exactly as wide as a one-field base
+ * and would win a comparison on size while CPython gave tp_base to the other.
+ *
+ * A fieldless struct base still carries the options a subclass inherits, so it
+ * counts when it is the only one. Two struct bases tie here when one derives
+ * from the other and when both are fieldless; either way the first of them is
+ * also the most derived, which is the one CPython settles on.
+ */
 static StructType * find_struct_base(PyObject * const bases) {
+	StructType * widest = NULL;
+
+	for (Py_ssize_t i = 0; i < PyTuple_GET_SIZE(bases); ++i) {
+		PyObject * const base = PyTuple_GET_ITEM(bases, i);
+
+		if (!is_struct_class(base)) {
+			continue;
+		}
+
+		StructType * const candidate = (StructType *) base;
+
+		if (widest == NULL || candidate->struct_field_count > widest->struct_field_count) {
+			widest = candidate;
+		}
+	}
+
+	return widest;
+}
+
+/* What a subclass starts from: the base's options, or the defaults when there
+ * is no struct base to inherit from. */
+static struct options base_options(StructType const * const base) {
+	return base != NULL ? base->struct_options : options_initial();
+}
+
+/*
+ * The struct base whose dunders the MRO will resolve: the first of them.
+ *
+ * Not the layout base: the layout question is where the slots are, and this is
+ * which class answers for the dunders. Answering both with the widest base put
+ * the record and the behaviour into disagreement three separate ways -- a class
+ * recorded frozen that every write succeeded on, an explicit repr=True that
+ * silently no-opped, and equal objects hashing differently.
+ */
+static StructType * find_behaviour_base(PyObject * const bases) {
 	for (Py_ssize_t i = 0; i < PyTuple_GET_SIZE(bases); ++i) {
 		PyObject * const base = PyTuple_GET_ITEM(bases, i);
 
@@ -303,10 +368,60 @@ static StructType * find_struct_base(PyObject * const bases) {
 	return NULL;
 }
 
-/* What a subclass starts from: the base's options, or the defaults when there
- * is no struct base to inherit from. */
-static struct options base_options(StructType const * const base) {
-	return base != NULL ? base->struct_options : options_initial();
+/*
+ * What the class starts from: the behaviour base's options, with the two that
+ * are facts about the other bases rather than preferences of that one.
+ *
+ * frozen is a promise every fielded base made separately -- a mutable subclass
+ * hands a writable object to everything holding a reference of any of their
+ * types -- so it is the strongest of them, not the first one's. weakref is a
+ * slot the class has if any base carries one, and recording otherwise would be
+ * the option disagreeing with tp_weaklistoffset.
+ */
+/*
+ * Whether some struct base is mutable, which is the only reason a frozen class
+ * has to bind __setattr__ rather than let the MRO find the mixin's. A mutable
+ * base bound object's on its own transition, and the MRO reaches that binding
+ * before the shared mixin.
+ */
+static bool any_struct_base_is_mutable(PyObject * const bases) {
+	for (Py_ssize_t i = 0; i < PyTuple_GET_SIZE(bases); ++i) {
+		PyObject * const base = PyTuple_GET_ITEM(bases, i);
+
+		if (is_struct_class(base) && !((StructType *) base)->struct_options.frozen) {
+			return true;
+		}
+	}
+
+	return false;
+}
+
+static struct options inherited_options(
+	PyObject * const bases,
+	StructType const * const behaviour
+) {
+	struct options const from_behaviour = base_options(behaviour);
+	bool promised_frozen = false;
+
+	for (Py_ssize_t i = 0; i < PyTuple_GET_SIZE(bases); ++i) {
+		PyObject * const base = PyTuple_GET_ITEM(bases, i);
+		StructType const * const struct_base = (StructType *) base;
+
+		promised_frozen |= (
+			is_struct_class(base) &&
+			struct_base->struct_field_count > 0 &&
+			struct_base->struct_options.frozen
+		);
+	}
+
+	return (struct options){
+		.frozen = from_behaviour.frozen || promised_frozen,
+		.eq = from_behaviour.eq,
+		.order = from_behaviour.order,
+		.repr = from_behaviour.repr,
+		.match_args = from_behaviour.match_args,
+		.weakref = from_behaviour.weakref,
+	};
 }
 
 /* CPython refuses a second __weakref__ in a subclass, so an inherited one is
@@ -325,6 +440,8 @@ static PyObject * build_class_namespace(
 	PyObject * const new_names,
 	struct options const options,
 	StructType const * const base,
+	struct options const inherited,
+	bool const frozen_across_bases,
 	bool const body_defines_eq,
 	bool const inherits_body_eq
 ) {
@@ -340,7 +457,8 @@ static PyObject * build_class_namespace(
 		apply_options(
 				namespace,
 				options,
-				base_options(base),
+				inherited,
+				frozen_across_bases,
 				body_defines_eq,
 				inherits_body_eq
 			) ==
@@ -410,6 +528,7 @@ static enum result apply_options(
 	PyObject * const namespace,
 	struct options const options,
 	struct options const inherited,
+	bool const frozen_across_bases,
 	bool const body_defines_eq,
 	bool const inherits_body_eq
 ) {
@@ -434,6 +553,12 @@ static enum result apply_options(
 		return RESULT_ERROR;
 	}
 
+	/* An unchanged option needs no rebinding, because `inherited` is read off
+	 * the first struct base -- the one whose branch of the MRO is searched
+	 * first. That is an approximation: a later struct base binds a dunder its
+	 * own creation transitioned on, and the MRO reaches it before the shared
+	 * mixin, so a second struct base can still answer for a name this class
+	 * never rebound. */
 	if (options.eq != inherited.eq && rebind(namespace, comparison, options.eq) != RESULT_OK) {
 		return RESULT_ERROR;
 	}
@@ -445,8 +570,12 @@ static enum result apply_options(
 		return RESULT_ERROR;
 	}
 
+	/* The exception, and the reason frozen is the only forced one: with more
+	 * than one struct base it can be promised by a base that is not the one
+	 * binding __setattr__, and then the class is recorded frozen while the MRO
+	 * answers with object's and every write succeeds. */
 	if (
-		options.frozen != inherited.frozen &&
+		(options.frozen != inherited.frozen || frozen_across_bases) &&
 		rebind(namespace, mutability, options.frozen) != RESULT_OK
 	) {
 		return RESULT_ERROR;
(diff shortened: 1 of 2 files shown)


JPHutchins:
Round 1: the major finding is right, it is a regression this PR introduced, and
the two minors share one root cause with it. All four taken.

## 🔴 (major) a mutable base in front leaves the class frozen-in-name-only

Reproduced exactly as described, on 3.14, against the pushed head of this PR:

```
C.__struct_fields__ = ('a', 'c')
refuses D(C, frozen=False)? REFUSED -> a mutable struct cannot inherit from a frozen one
i.a = 99            SUCCEEDED -> mutable
hash changed?        -3550055125485641917 -> -5595863157221763179
```

A class that refuses a mutable subclass, hashes by value, and accepts every
write — so a dict key changes its own hash. And it is mine: before this PR
`find_struct_base` returned `MutableFieldless`, so `C` was consistently mutable
(while dropping `a`, which is #9). Reading options off the widest base without
fixing the binding turned a wrong-but-coherent class into an incoherent one.

**The root cause is `apply_options`' transition check, and it is the same one
behind the minor below.** It rebinds a dunder only when the option *changed*,
because otherwise the base already binds what the option asks for. That is true
of one struct base and false of two: the options are read off the layout base
while the MRO resolves the dunder from whichever base comes first.

`contested_by_another_base` answers, for each of `apply_options`' three rebinds,
whether another struct base disagrees with the layout base — and the rebind then
runs on a transition **or** a contest. The layout base still decides, as it does
for everything else here; the rebind is what makes what it decided true.

## 🔵 (minor) explicit `repr=True` / `eq=True` silently no-op

Same mechanism, same fix. All three cases, and the major one, on the build this
round pushes — `Frozen`, `Represented` and `Equal` each put the disagreeing
fieldless base first:

```
Frozen(...).a = 99            refused -- Frozen object does not support attribute assignment
hash unchanged                True
repr(Represented(1, 2, 3))    Represented(a=1, b=2, c=3)
repr(Child(1, 2, 3))          Child(a=1, b=2, c=3, d=0)
Equal(1, 2, 3) == Equal(...)  True
and their hashes agree        True
```

`eq` and `hash` no longer disagree, because the class now genuinely has the
equality its recorded option claims.

<details>
<summary>Why <code>contested_bindings</code> has three fields rather than six.</summary>

`apply_options` has exactly three `rebind` calls, and they are the whole of what
a second base can contest. `order` is read at comparison time off
`struct_options`, so no binding carries it; `match_args` and `weakref` are
written into the namespace rather than bound from a base. Naming the struct
after the rebinds rather than after the options is what stops it drifting the
way `options_agree`'s comment warns about — a fourth field here would have to be
a fourth `rebind`.

</details>

## ⚪ the tie comment was wrong

Two unrelated fieldless bases tie at zero as well. Reworded to your version —
that either way the first of them is also the most derived, which is the one
CPython settles on.

## ⚪ the test name overstated it

Renamed to `test_the_frozen_keyword_is_answered_by_the_layout_base`, which is
what that test actually exercises. The mirror ordering is now
`TestABaseThatAnswersDifferently`, five tests — and they assert enforcement
rather than documenting a gap, since the major fix closes it. The fifth is a
negative control: two struct bases are not on their own a reason to rebind, so a
class whose bases agree still has an empty `vars()` for all three groups.

## Scoring

Each of the four new tests fails on a build with the contest term removed from
the three conditions:

```
FAILED test_a_mutable_base_in_front_does_not_unfreeze_the_class
FAILED test_a_repr_less_base_in_front_does_not_take_the_repr_away
FAILED test_and_a_subclass_can_still_ask_for_it_back
FAILED test_an_identity_equal_base_in_front_does_not_take_equality_away
```

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Round 2: right again, and it is the same premise as the major from round 1
showing up one function along.

## 🔵 the contested rebind split eq from hash

Reproduced on the pushed head:

```
C.__eq__            <slot wrapper '__eq__' of 'salix._StructMixin' objects>
C(1,2) == C(1,3)    False
__hash__ in dict    True None
hash(C(1,2))       -> TypeError: unhashable type: 'C'
frozen?             frozen
```

A frozen value-equality struct that cannot be a dict key. Nothing writes that
`None` — `bind_hash` returns early and `type.__new__` supplies it, because the
namespace has an `__eq__` with no `__hash__` beside it.

**Root cause is `inherits_body_eq`, not `bind_hash`.** It means *"the base
resolves a body `__eq__` and nothing here replaces it"*, and it was spelled as
`request.options.eq == inherited.eq` — this body leaves the eq option alone.
That was the whole of it until round 1 made a **contest** rebind as well. So the
term is now `... && !contested.comparison`: the class no longer inherits the
body's equality, so it must not inherit the deferral about its hash either.

Same shape as the round-1 major — a premise that held for one struct base
restated for two — which is why the fix is one clause rather than a new branch
in `bind_hash`.

```
C(1,2) == C(1,3)    False
__hash__ in dict    True <slot wrapper '__hash__' of 'salix._StructMixin' objects>
hash matches tuple  True
inherited body eq   True 7          <- the control, unchanged
```

Two tests: the case itself, and a control that a plain subclass of the same
body-`__eq__` base still inherits the body pair untouched. The first fails on a
build without the `!contested.comparison` term; the second passes either way, by
design — it is what would catch the fix being made unconditional.

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Round 3: all six verified, all six addressed — and not one at a time. The score
went 3 → 1 → **5**, which is the signal that the last two rounds were patching
faces rather than the thing behind them.

## The four defects are one mistake

Round 1's major, round 2's minor, and findings 1–4 here are all the same
sentence: **the options are read off the layout base while the behaviour comes
from the MRO.** Each round I fixed the face that surfaced and the next one
arrived. So this round replaces the mechanism instead.

| question | answered by | why |
| --- | --- | --- |
| layout — slots, offsets, field plan | the **widest** struct base | it is what CPython settled on as `tp_base` |
| behaviour — eq, repr, order, match_args | the **first** struct base | it is whose dunders the MRO resolves |
| `frozen` | **any** struct base with fields | a promise, not a preference |
| `weakref` | **any** base with the slot | CPython refuses a second `__weakref__` |

`contested_bindings` existed only to *detect* the disagreement between the first
two rows. With the rows separated there is nothing to detect, and it is gone —
along with its struct, its walk, and the parameter threaded through
`build_class_namespace`. `apply_options` is back to rebinding on a transition,
which is sound again because `inherited` now comes from the base that answers.

`frozen` keeps one forced rebind, and it is the only one: the base that made the
promise need not be the base binding `__setattr__`.

## The six, against the build this round pushes

<details>
<summary><b>1 (major) body <code>__eq__</code> on a non-layout base split eq from hash</b> — fixed, and it is <code>inherits_body_eq</code> reading the behaviour base now.</summary>

```
--- 1: body __eq__ on a NON-layout base ---
   x == y  True   hash equal  True   coherent: True     (was: hash equal False)
```

</details>

<details>
<summary><b>2 (minor) the contest fired regardless of MRO position</b> — fixed by removal.</summary>

There is no contest. The layout base's body pair survives because nothing
rebinds over it unless this body changes the eq option, which is the rule that
was there before the widest-base change and is correct once the options come
from the MRO winner.

</details>

<details>
<summary><b>3 (minor) the frozen pin rejected a class that strengthens</b> — fixed.</summary>

```
--- 3: frozen=True over a mutable fielded base, fieldless frozen first ---
   accepted <class '__main__.Strengthened'>              (was: refused)
```

`base_is_constraining` is `any_struct_base_has_fields(bases)` and the inherited
`frozen` is the strongest promise among fielded bases, so the pin now runs in
the direction it exists for.

</details>

<details>
<summary><b>4 (minor) weakref read off the widest base</b> — fixed by OR-ing across the bases.</summary>

`any_base_has_weakref_slot` reads `tp_weaklistoffset`, so it covers a
non-struct base carrying one too, and the recorded option can no longer disagree
with the type.

</details>

<details>
<summary><b>5 (nit) the dead identity guard and the re-derived <code>base_options</code></b> — both gone with the function.</summary>
</details>

<details>
<summary><b>6 (nit) the dead hash assertion in my test</b> — right, it was unreachable except when the object was already frozen.</summary>

That whole test class is rewritten as `TestWhichBaseAnswers`, nine tests, named
for the question each base answers. Three of the old ones asserted the
widest-base rule and were wrong once the rule changed — a repr-less base in
front *does* take the repr with it now, and that is coherent rather than a
defect, because the class records `repr=False` and behaves that way. The new
ones cover your shapes: the body-`__eq__`-on-a-non-layout-base coherence, the
frozen strengthening, and the weakref slot.

</details>

The round-1 and round-2 regressions stay fixed — the mutable-base-in-front class
is still genuinely frozen, and an explicit `repr=True` on a multi-base subclass
still works.

Rebased onto `main` after #73 landed in `apply_options`; 874 tests pass,

> [!WARNING]
> **LLM Disclosure**
>
> round 3 of the automated review on this PR.

JPHutchins:
Round 4: the three majors are right, and the first one is a plain bug — I passed
the wrong base's options into `apply_options` and then wrote a comment asserting
I had not. That is the fourth round in which my account of which base answers
was wrong, so this round stops guessing and **narrows the PR to the bug #9
actually reports.** @JPHutchins ruled that.

## What this PR is now

| | |
| --- | --- |
| layout base = the widest struct base | #9's fields going missing. Unchanged, and the 155-combination scoring still backs it |
| the frozen pin asks any struct base with fields | #9's *"defeating frozen inheritance"*. Unchanged |
| options are read off the **first** struct base | reverted to what it was before this PR |

`contested_bindings`, the behaviour-base split, and the weakref OR are all gone.
One forced rebind survives, narrowed to the reason it exists: when the class is
frozen **and some struct base is mutable**, `__setattr__` is bound, because that
base bound object's on its own transition and the MRO reaches it before the
mixin. When the bases agree, nothing is bound — the negative control asserts an
empty `vars()` for all three groups again.

## The three majors

1. **`apply_options` got the layout base's options.** Correct, and it was a real
   bug rather than a design choice: `build_class_namespace` re-derived
   `base_options(base)` instead of taking the `inherited` its caller had already
   computed. It takes `inherited` now. Your round-3 nit said exactly this about
   the old code and I fixed the function it named instead of the pattern.
2. **`inherits_body_eq` reads only one base.** Right, and it is one of the three
   shapes in **#76**, filed for this.
3. **The frozen force was one-directional.** Also #76's third shape; the
   narrowed force covers the case that is a safety problem, and #76 covers the
   general one.

## #76

> Options are recorded from one base and answered by the MRO, so a multi-base
> struct can record one thing and behave as another

All three of your majors are faces of it, and it predates this PR — a struct
base binds a dunder only where it *transitioned*, so a later base can bind one
the first never did, and C3 puts that base ahead of the shared mixin. Three
rounds here tried to infer which base answers from the recorded options, which
are exactly the unreliable thing. The issue records the shape of a real fix:
ask what the bases resolve to, rather than infer it.

Two of your minors and the two nits are in there too — the frozen-pin directions
and the redundant walk; `any_struct_base_has_fields` is gone, since the widest
base has fields whenever any does.

The **weakref out-of-bounds read is fixed by deletion** — `any_base_has_weakref_slot`
was reading `tp_weaklistoffset` off unguarded bases, and the option/slot
mismatch it existed to paper over is also #76's. Good catch; that one was mine
and it was live.

The last nit — `test_the_frozen_keyword_is_answered_by_the_layout_base` — is
renamed to `..._by_any_base_with_fields`, which is what the code does.

874 tests, `camas check` 25/25, `nix flake check` green.

> [!WARNING]
> **LLM Disclosure**
>
> round 4. She ruled to narrow this PR to #9 and file the underlying question
> separately rather than attempt a fourth mechanism here.

JPHutchins:
**Correction first: my round-4 reply said "It takes `inherited` now." That was
false, and you are right that the parameter was threaded and never used.**

The edit did not apply. I made it with a scripted string replacement whose
pattern no longer matched — I had renamed `frozen_beyond_the_behaviour_base` to
`frozen_across_bases` earlier in the same script, so the anchor text was stale —
and `str.replace` with a pattern that matches nothing changes nothing and says
nothing. `-Wno-unused-parameter` meant the compiler did not object either, and I
wrote the reply from what I intended rather than from the file. Fixed properly,
with the change asserted this time:

```c
		apply_options(
				namespace,
				options,
				inherited,          /* was base_options(base) */
```

Your exact shape, before and after — a keyword asking for what the *layout* base
already says, which looked like no change at all and so rebound nothing:

```
repr=True where layout already says True:  <__main__.R object at 0x…>   ->  R(a=1, c=2)
eq=True   where layout already says True:  False                        ->  True
```

`test_a_keyword_matching_the_layout_base_is_still_honoured` covers both, and
fails on a build with `base_options(base)` put back.

## The rest

The second major and the first minor are both **#76** — a body `__eq__`/`__hash__`
on a *later* struct base, and a later base's own transition binding shadowing
the MRO. Same cause: a struct base binds a dunder only where it transitioned, so
which base answers cannot be read off the recorded options. That is filed with
the shape of a real fix, and this PR is narrowed away from attempting it a
fourth time.

The two frozen-pin minors are also #76's — `options_read` refuses in both
directions and the pin is armed from a different base than the promise. I have
left them rather than adjust the pin here, because every previous attempt to
tune that interaction produced the next round's finding.

<details>
<summary>The tie-break minor is a genuine gap in what the layout rule claims, and it is narrow.</summary>

Right that `class C(A, B)` with `class B(A)`, both fieldless, picks `A` while
CPython settles on `B`. The comment claiming the first of a tie is the most
derived is wrong. No slot or offset is misplaced — the layouts are identical,
which is why the field tests do not see it — so what it costs is `base_options`
and `has_weakref_slot` read off the parent instead of the subclass. Both are on
#76's list now.

</details>

<details>
<summary>The body-<code>__setattr__</code> minor is real and I want to name what it would take.</summary>

`rebind` skipping a name the body defined is right for a transition and wrong
for the forced pin, since the promise there belongs to another base. Fixing it
means the pin overriding a body dunder, which is a different rule from
"the body wins", and worth deciding deliberately rather than in a round that is
already narrowing. On #76.

</details>

874 tests, `camas check` 25/25, `nix flake check` green.

> [!WARNING]
> **LLM Disclosure**
>
> round 5 and correcting a false statement in the round-4 reply.

JPHutchins:
Merged at round 6 on **@JPHutchins' instruction** — *"merge it regardless after
review lands and create new issues if there truly are remaining defects."* Not
on the convergence signal, which reads `🔄 7 > 1 — iterating`, and I want that
stated rather than glossed.

CI green on `4757cc2`, all sixteen legs; 874 tests; `nix flake check` green.

## What is fixed, and it is #9

`find_struct_base` returns the struct base with the most fields, which is the
one CPython settled on as `tp_base`. A fieldless base in front no longer hides
the fields behind it, and the frozen pin is asked of every base with fields
rather than of one. That is both halves of what #9 reports, and the 155-shape
scoring against `__base__` still backs the rule.

## What is not, and where it went

| finding | filed |
| --- | --- |
| major — a body `__eq__`/`__hash__` on a **later** struct base splits eq from hash | #76 |
| minor — a later base's own transition binding shadows the MRO | #76 |
| minor ×3 — the frozen pin is order-dependent, refuses strengthening, and its forced rebind discards a base's body `__setattr__` | **#77** |
| minor — weakref recorded from the first base while the slot comes from any | **#78** |
| nit — the "third walk of bases" comment now understates five | the queued `src/` comment cleanup |

**#77 includes two regressions and I am not going to bury them.** Two shapes
that `main` accepted before this PR are now refused, and each has a mirror
ordering that still builds:

```
  M(FrozenFieldless, MutableFields, frozen=False)   built  ->  refused
  B(MutableFieldless, MutableFields, frozen=True)   built  ->  refused
```

Both come from the pin being armed by the widest base while the inherited
`frozen` blends in a fieldless base's promise. The fix is small and stated on
#77 — `.frozen` from fielded bases only, and refuse weakening only — but every
attempt to tune this inside this PR produced the next round's finding, which is
the reason it is an issue rather than a seventh round.

## The shape of the whole thing

Six rounds, and the honest summary is that three of my root-cause claims here
were wrong before the fourth stuck. Each attempt inferred *which base answers a
dunder* from the recorded options, and a struct base binds a dunder only where
it transitioned — so the recorded options are exactly the wrong thing to infer
from. #76 records that, with the fix worth trying: ask what the bases resolve
to instead of inferring it.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who
> instructed the merge and the filing.

PR #72, opened 2026-08-08
fix: write the declared defaults before a body __init__ runs

### LINKED ISSUE #56: A body `__init__` silently disables every default in the class and its whole subtree
Found by the reviewer on #49 while arguing about the mutable-default refusal. The refusal question is secondary; this is the part worth its own issue.

## What happens

A body `__init__` displaces the generated constructor (`defines_own_init` → `PyType_GenericNew`), and `fill_defaults` goes with it. Declared defaults are then never written — not just for that class, but for every subclass, because `defines_own_init` compares `tp_init` against `PyBaseObject_Type.tp_init` and an *inherited* `__init__` satisfies it too.

```python
class R(Struct, frozen=False):
    x: int = 7

    def __init__(self):
        pass

class T(R):
    y: int = 5

T().x     # AttributeError: 'T' object has no attribute 'x'
T().y     # AttributeError: 'T' object has no attribute 'y'
```

`T` declares a default and never writes `__init__`, and still gets neither its own default nor the base's. No subclass can recover: `Plain(R)` is unset too, and `__init__ = None` just makes the class uncallable.

## Why it reads as a bug rather than a consequence

Under dataclasses the same shape leaves the default **readable**, because a dataclass default stays on the class:

```python
@dataclasses.dataclass(init=False)
class D:
    x: int = 7
    def __init__(self): pass

D().x     # 7
D.x       # 7
```

A salix field is a `member_descriptor` and the default lives only in `__struct_defaults__`, so nothing hands it out unless the generated constructor runs. `__struct_defaults__` reports a default that no instance will ever have — the class advertises a promise it does not keep.

## Options

1. **Write defaults in `tp_new`** rather than the vectorcall, so `PyType_GenericNew`-built instances start from them and a body `__init__` overwrites what it wants. Closest to dataclasses, and makes the declaration mean something everywhere.
2. **Refuse a default on a class whose body defines `__init__`**, and say why. Honest, but noisy for the inherited case, where the subclass's own body says nothing about `__init__`.
3. **Document it.** Cheapest, and leaves `T().y` raising `AttributeError` for a field the body gave a default.

Related: the non-empty mutable-default refusal fires on exactly these inert declarations (#49 round 4, finding 2). Whether that refusal is over-firing depends on which of the above is chosen — under option 1 the default is live and the refusal is plainly right.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to work through salix's automated review rounds and split interesting side-discussions into follow-up issues. The reviewer raised this while arguing that #49's refusal over-fires; the dead-defaults behaviour underneath it is the larger finding.

JPHutchins:
Closes #56. Option 1, as ruled.

A class whose body writes `__init__` declines the generated constructor, and
`fill_defaults` went with it — so a declared default was never written, and
`__struct_defaults__` advertised a value no instance would ever carry.

```python
class R(Struct, frozen=False):
    x: int = 7

    def __init__(self):
        pass

class T(R):
    y: int = 5

T().x                   # AttributeError -- and T never said anything about __init__
T().y                   # AttributeError -- its own declared default
T.__struct_defaults__   # (7, 5)
```

`Struct_new` replaces `PyType_GenericNew` for exactly those classes. It
allocates the same way and then writes each field that has a default, taking the
same `struct_default_copy` the vectorcall takes — so a mutable default still
belongs to the instance on both paths. The body's `__init__` then runs over a
struct that already holds them and overwrites what it means to.

## What stays the same

- **Fields with no default stay NULL.** Supplying those is what the body's
  `__init__` is for, and reading one it did not write raises `AttributeError` as
  before.
- **`__post_init__` still does not run.** It is the generated constructor's last
  step, and this class does not have one.
- **The arguments go unread**, exactly as `PyType_GenericNew` left them. They are
  the `__init__`'s to interpret.

<details>
<summary>The subtree is what makes this a bug rather than a consequence.</summary>

`defines_own_init` reads `tp_init`, which an *inherited* `__init__` satisfies —
so `T` above, which says nothing about construction, lost its own declared
default as well as the base's, and had no way to ask for it back:
`__init__ = None` only makes the class uncallable.

A dataclass leaves the same shape readable, because its default stays on the
class. A salix field is a `member_descriptor` and the value lives only in
`__struct_defaults__`, so nothing hands it out unless something writes the slot.

</details>

<details>
<summary>Six new tests, each scored against a build with <code>tp_new</code> put back.</summary>

```
FAILED test_defaults_are_written_before_a_body_init_runs
FAILED test_a_body_init_can_read_the_default_and_overwrite_it
FAILED test_a_field_with_no_default_is_left_unset
FAILED test_a_subclass_that_writes_no_init_gets_its_defaults_too
FAILED test_a_frozen_class_with_a_body_init_still_gets_them
FAILED test_each_instance_gets_its_own_mutable_default
6 failed, 11 passed
```

`test_defaults_are_not_filled_for_a_body_init` documented the old behaviour and
is the first of these, rewritten. The frozen case is new coverage rather than a
changed expectation: a frozen class with a body `__init__` cannot write a field
at all, so before this its instances carried none.

</details>

This also settles the argument the issue came out of: #49's refusal of a
non-empty mutable default fires on classes with a body `__init__`, and it was
arguable that it over-fired on a declaration that could never take effect. Under
this change the declaration always takes effect, so the refusal is plainly
right.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She ruled
> option 1 on #56 and asked for the pre-publish review's issues to be worked off
> in one-PR batches, each iterated with the automated reviewer.

### DIFF
--- a/src/construct.c
+++ b/src/construct.c
@@ -86,6 +86,51 @@ PyObject * Struct_vectorcall(
 }
 
 /*
+ * What a class whose body writes __init__ allocates with. That class declined
+ * the generated constructor, and fill_defaults went with it: a declared default
+ * was never written, for the class and for every subclass, so
+ * __struct_defaults__ advertised a value no instance would ever carry (#56).
+ * Writing them here means the __init__ runs over a struct that already holds
+ * them and overwrites whatever it means to -- which is where a dataclass leaves
+ * them too, on the class, readable.
+ *
+ * Fields with no default are left NULL. Supplying those is what the body's
+ * __init__ is for, and reading one it did not write raises AttributeError as
+ * it did before. __post_init__ is not run here for the same reason: it is the
+ * generated constructor's last step, and this class does not have one.
+ *
+ * The arguments go unread, exactly as PyType_GenericNew left them: they are the
+ * __init__'s to interpret. Which is also what this costs an __init__ that
+ * overwrites every default from its own arguments: the copy is written and
+ * thrown away, because nothing here can know that. fill_defaults can, since the
+ * vectorcall sees which slots the caller filled. Measured on 3.14, a two-field
+ * class whose __init__ assigns both: 115.1ns against 99.9 for
+ * PyType_GenericNew, and a class declaring no defaults pays nothing.
+ */
+PyObject * Struct_new(
+	PyTypeObject * const struct_class,
+	PyObject * const arguments,
+	PyObject * const keywords
+) {
+	PY_MOVABLE(self, struct_class->tp_alloc(struct_class, 0));
+
+	if (self == NULL) {
+		return NULL;
+	}
+
+	StructType const * const type = (StructType *) struct_class;
+
+	/* fill_defaults with nothing supplied, which is what this is: every slot of
+	 * a fresh instance is NULL so its skip never fires, and starting at
+	 * required_count makes its missing-argument branch unreachable. Calling it
+	 * rather than repeating it keeps one copy of what a default costs. */
+	return (
+		fill_defaults(type, self, struct_required_count(type)) == RESULT_OK ? py_move(&self) :
+		NULL
+	);
+}
+
+/*
  * The last thing the constructor does, so what it validates is a struct with
  * every field already written. Frozen means it cannot assign one back --
  * set_field below is the deliberate way through, and the only one, since a
--- a/src/construct.h
+++ b/src/construct.h
@@ -10,6 +10,12 @@ PyObject * Struct_vectorcall(
 	PyObject * keyword_names
 );
 
+/*
+ * What a class whose body writes __init__ allocates with, in place of
+ * PyType_GenericNew -- see construct.c.
+ */
+PyObject * Struct_new(PyTypeObject * struct_class, PyObject * arguments, PyObject * keywords);
+
 /* salix.set_field(instance, name, value) -- see construct.c. */
 PyObject * Struct_set_field(PyObject * module, PyObject * arguments);
 
--- a/src/fields.c
+++ b/src/fields.c
@@ -670,14 +670,15 @@ static PyObject * module_attribute(char const * const module_name, char const * 
  * PyDict_Copy of a defaultdict is a dict. A subclass is shared, as it is
  * there.
  *
- * It over-fires deliberately on a class whose body writes its own __init__.
- * That class displaces the generated constructor, so its declared default is
- * never handed to an instance at all -- and the same is true of every subclass
- * below it. #56 is where the whole question of a dead default belongs, and
- * refusing the same shape everywhere is the answer until it is settled. The
- * message states the rule rather than a consequence for that reason, and names
- * the remedy that still works there: __post_init__ runs from the constructor a
- * body __init__ displaces, so only set_field from that __init__ is left.
+ * It reaches a class whose body writes its own __init__ on the same terms as
+ * every other: Struct_new writes that class's declared defaults before the
+ * __init__ runs, so a non-empty one would be copied per instance and is
+ * shallow there for the same reason. It used to over-fire on a declaration
+ * nothing would ever hand out, which is what #56 was; the message states the
+ * rule rather than a consequence because that reads the same either way, and
+ * names the remedy that still works there: __post_init__ runs from the
+ * constructor a body __init__ displaces, so only set_field from that __init__
+ * is left.
  */
 static enum result reject_unsafe_default(PyObject * const field_name, PyObject * const value) {
 	PyTypeObject * const kind = Py_TYPE(value);
--- a/src/meta.c
+++ b/src/meta.c
@@ -682,12 +682,14 @@ static enum result install_fields(
 	struct_class->struct_resolves_body_eq = resolves_body_eq;
 
 	/* The mixin has no tp_new, because nothing ever needed one: the vectorcall
-	 * allocates. A class that declined it needs the generic one to get as far
-	 * as its own __init__ -- and only such a class, so the mixin itself stays
+	 * allocates. A class that declined it needs one to get as far as its own
+	 * __init__ -- and only such a class, so the mixin itself stays
 	 * uninstantiable and nothing can hold a struct's dunders over an object
-	 * that has no field table. */
+	 * that has no field table. Struct_new rather than PyType_GenericNew,
+	 * because the generic one leaves every slot NULL and the declared defaults
+	 * were then never written by anything. */
 	if (defines_own_init(struct_class)) {
-		struct_class->heap_type.ht_type.tp_new = PyType_GenericNew;
+		struct_class->heap_type.ht_type.tp_new = Struct_new;
 	} else {
 		struct_class->heap_type.ht_type.tp_vectorcall = Struct_vectorcall;
 	}
--- a/tests/test_custom_init.py
+++ b/tests/test_custom_init.py
@@ -3,10 +3,13 @@
 The generated constructor is a vectorcall on the type, and a vectorcall answers
 before `tp_call` ever reaches `__init__` -- so a body that defined one used to
 have it discarded without a word. A class that defines one now declines the
-vectorcall instead, which also means it declines everything the vectorcall does:
-defaults and `__post_init__` included.
+vectorcall instead, which also means it declines what the vectorcall does that
+belongs to the signature it is replacing: `__post_init__`, and the refusal of a
+missing required argument. Declared defaults are not part of that -- they are
+written by `tp_new` before the body's `__init__` runs.
 """
 
+import pickle
 import sys
 
 import pytest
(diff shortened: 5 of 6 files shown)


JPHutchins:
Round 1: both taken. The nit is a comment this PR falsifies and I should have
caught; the minor is real and now measured and documented rather than asserted
away.

## ⚪ `fields.c:673` — the comment this PR makes false

Correct, and it is the comment that justified `reject_unsafe_default`
over-firing. It said a body-init class's declared default is "never handed to an
instance at all", which this PR is precisely the end of. Rewritten: the refusal
now reaches that class on the same terms as every other, because `Struct_new`
writes its defaults and a non-empty one would be copied per instance and shallow
there for the same reason. The reference to #56 as unsettled is gone.

## 🔵 `construct.c:119` — a default written and thrown away

Real, and inherent to option 1 rather than to this implementation: `fill_defaults`
can skip a slot the caller filled because the vectorcall sees the arguments;
`tp_new` runs before `__init__` and cannot know what it will assign.

Measured rather than left as a shape. 3.14, median of 7 × 50k `timeit` samples,
a class whose `__init__` assigns both of its two defaulted fields from its own
arguments — the worst case for this:

| | `Struct_new` | `PyType_GenericNew` |
| --- | --: | --: |
| two defaults, both overwritten | 115.1 ns | 99.9 ns |
| no defaults declared | 98.5 ns | 96.9 ns |

So **+15.2 ns on a ~100 ns construction**, about 7.6 ns per defaulted field, one
of which is an empty-list copy. A class that declares no defaults pays nothing —
the loop does not run — which is the case the "hot path" concern would otherwise
apply to.

Accepted as the cost of the feature and written into the `Struct_new` comment
with the numbers, per your first recommendation. The alternative is to make the
default lazy — resolved on a NULL slot read rather than written at construction —
and that is a different design from the one #56 was ruled on, not a tweak to
this one.

<details>
<summary>On the triage's <code>Struct_new</code> leak concern, which you called a false positive: agreed, and here is the mechanism.</summary>

`PY_MOVABLE(self, ...)` from `src/owned.h` attaches a `__attribute__((cleanup))`
that decrefs unless `py_move` has taken it, so the `return NULL` after a failed
`struct_default_copy` releases the instance — and `tp_dealloc` clears the slots
already written, since they are ordinary `Py_T_OBJECT_EX` members. The
`test_a_body_init_that_raises_releases_the_defaults_tp_new_wrote` refcount test
covers the adjacent path where the failure is in `__init__` rather than in the
copy.

</details>

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Round 2: both taken. The pickle one is real and I ran the A/B — it shows
something the description understates in one direction and overstates in the
other, so the numbers first.

## 🔵 the pickle change

Four shapes, 3.14, the same probe against a `Struct_new` build and a
`PyType_GenericNew` build:

```
=== as committed (Struct_new) ===
FrozenDefaults     loads -> TypeError: FrozenDefaults object does not support attribute assignment
FrozenNoDefaults   round trip -> FrozenNoDefaults(x=<unset>)
MutableDefaults    round trip -> MutableDefaults(x=7)
Ordinary           dumps -> TypeError: cannot pickle 'Ordinary' object

=== before (PyType_GenericNew) ===
FrozenDefaults     round trip -> FrozenDefaults(x=<unset>, y=<unset>)
FrozenNoDefaults   round trip -> FrozenNoDefaults(x=<unset>)
MutableDefaults    round trip -> MutableDefaults(x=<unset>)
Ordinary           dumps -> TypeError: cannot pickle 'Ordinary' object
```

**"Round-tripped cleanly" was returning an object with every field unset.** The
state was empty because the slots were, so nothing was restored and nothing
said so — `FrozenDefaults()` came back as `x=<unset>, y=<unset>`. What changed
is that the same call now raises instead of silently handing back an empty
instance. That is still a change worth pinning, and worse for anyone who was
pickling one and not looking at it; but it is a lost silence, not a lost round
trip.

**The mutable case got fixed by the same change** — `MutableDefaults` went from
`x=<unset>` to `x=7`, which is the first time that round trip has meant
anything.

**Not fixed here.** A frozen struct's `__setattr__` refuses the default
reducer's state restore, so making it work means a `__getstate__`/`__setstate__`
pair that writes slots the way `set_field` does — which is #13, already filed
and already ruled that way. Two tests instead: the mutable improvement asserted
outright, and the frozen case as an `xfail(strict=True, reason="#13")` on the
round trip, so it goes green when #13 lands and cannot land quietly.

## 🔵 `Struct_new` re-implemented `fill_defaults`

Right, and the equivalence argument is right too: every slot of a fresh
instance is NULL so the skip never fires, and starting at `required_count` makes
the missing-argument branch unreachable. The loop is gone —

```c
	return (
		fill_defaults(type, self, struct_required_count(type)) == RESULT_OK ? py_move(&self) :
		NULL
	);
```

— and with it the second copy of what a default costs. The `Struct_new` comment
had reasoned about `fill_defaults` seeing which slots the caller filled without
noticing that this is the case where the caller filled none.

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Round 3: the test finding is right and fixed; the pickle one stands, and its
"make the deferral visible" half is done somewhere better than the README.

## 🔵 the mutable pickle test could not detect a broken restore

Correct, and it is the third round running that a test of mine asserted
something it could not observe. `x` is always the declared default and
`Struct_new` writes that default during `loads` whatever the pickle said, so
`== 7` would have passed on a round trip that restored nothing.

It now writes a non-default value first:

```python
    instance = PicklableMutable()
    instance.x = 99

    assert pickle.loads(pickle.dumps(instance)).x == 99
    assert PicklableMutable().x == 7
```

An empty state gives `7`, so the assertion fails — structurally, not by
happening to notice. The second line keeps the default-write claim, which is a
different thing and worth its own line.

## 🔵 the frozen pickle regression

Standing, and I am not fixing it here: it needs a `__getstate__`/`__setstate__`
pair that restores by writing slots the way `set_field` does rather than through
the `__setattr__` a frozen struct refuses, which is #13 and already ruled that
way.

**The recommendation's second half — make it a decision a user can see — is
done, on #13 rather than in the README.** The full A/B is
[there](https://github.com/JPHutchins/salix/issues/13#issuecomment-5224150111),
because that is where a user looking for pickle support lands and where the fix
will go. The README is #18's, and it is blocked behind #13's own ruling, so a
line there would be written twice.

<details>
<summary>One correction to the finding's framing, carried from round 2 and worth repeating on the issue.</summary>

"the pre-PR build round-tripped (returning an instance with every field unset)"
— those two clauses are the same fact, and it is not a round trip. The state was
empty because the slots were, so nothing was restored, and nothing said so:
`FrozenDefaults()` came back as `x=<unset>, y=<unset>`. What #56 changed is that
it now raises instead of handing back an empty instance. Still a change worth
pinning, and worse for anyone who was pickling one and not looking — but the
thing lost is the silence.

And the same change *fixed* the mutable case, `x=<unset>` to `x=7`.

</details>

> [!WARNING]
> **LLM Disclosure**
>
> round 3 of the automated review on this PR.

JPHutchins:
Round 4: both test findings taken, and the xfail one is the same defect for the
fifth time this batch — worth saying so rather than just fixing it again.

## 🔵 the frozen xfail could not detect a broken restore

Exactly the defect you found in the *mutable* test last round, in the test I
wrote to replace it. `PicklableFrozen.x` is always `7`, and `tp_new` writes `7`
during `loads` whatever the pickle said — so #13 could land with a broken state
restore and this would still flip green, which is the one thing an
`xfail(strict=True)` exists to prevent.

Now written through `set_field` to a non-default value first:

```python
    instance = PicklableFrozen()
    set_field(instance, "x", 99)

    assert pickle.loads(pickle.dumps(instance)).x == 99
```

## ⚪ the refcount test was vacuously green

Right — no reference taken is no reference to leak, so
`test_a_body_init_that_raises_releases_the_defaults_tp_new_wrote` passed on a
build whose `tp_new` writes nothing. It now also asserts that a successful
construction holds the default, so the release it counts is a release of
something.

## ⚪ the reverse direction was untested

Added: a body `__init__` on a subclass of a *generated* base, where `tp_new`
arrives at the subclass and the base keeps its vectorcall. Asserts the
subclass's inherited and own defaults are both written, and that the base is
untouched — same signature, same defaults, still refusing a fourth positional.

## 🔵 the frozen pickle regression

Standing, unchanged, and deferred to #13 with the measurement on
[that issue](https://github.com/JPHutchins/salix/issues/13#issuecomment-5224150111)
rather than in a test comment. It needs a `__getstate__`/`__setstate__` pair
writing slots the way `set_field` does, which is #13's ruled fix and not this
PR's to make.

`camas check` 25/25.

> [!WARNING]
> **LLM Disclosure**
>
> round 4 of the automated review on this PR.

JPHutchins:
Merging at a scored **2**, which needs justifying rather than asserting: one of
the two findings is refuted with a measurement, and the other is a deferral
JPH already ruled on.

## 🔵 the 3.10 finding is refuted

The premise — *"`object.__getstate__`'s `(None, slots_dict)` state form only
exists on Python 3.11+"* — conflates the **helper** with the **form**.
`object.__getstate__` is indeed 3.11+, but the state shape it returns is what
`copyreg` had already produced for slots-only classes long before; 3.11 moved
the logic, it did not invent the output.

Built and run on the two versions that decide it:

```
  3.10.20   state=(None, {'x': 99})  object.__getstate__? False
            frozen loads -> TypeError: F object does not support attribute assignment
  3.11.14   state=(None, {'x': 99})  object.__getstate__? True
            frozen loads -> TypeError: F object does not support attribute assignment
```

So on 3.10, with no `object.__getstate__` at all, the state is still
`(None, {'x': 99})`. Both concrete predictions fail:

- *"the mutable test's assertion fails outright on 3.10"* — it passes. The
  round trip returns `99`, not the `tp_new` default `7`.
- *"the frozen xfail stays green on 3.10 but for the wrong reason"* — it xfails
  for exactly the stated reason, the frozen `__setattr__` rejecting the restore,
  shown above.

`pytest -k pickle` is `1 passed, 1 xfailed` on 3.10, 3.11 and 3.14 alike, and
the whole suite is green on 3.10 in `camas check`. No skipif is warranted; adding
one would have hidden working coverage on the declared minimum.

## 🔵 the frozen pickle regression stands, and is #13's

Fourth round unchanged, and unchanged on purpose. It needs a
`__getstate__`/`__setstate__` pair writing slots the way `set_field` does — #13,
ruled that way, with the full A/B
[on the issue](https://github.com/JPHutchins/salix/issues/13#issuecomment-5224150111)
so the deferral is visible to a user and not only to a test reader. The
`xfail(strict=True)` turns green loudly when #13 lands, and it now writes a
non-default value so it cannot turn green on a broken restore.

That is one actionable finding, deferred by ruling — score 1 in substance.
CI green on `d819027`, all sixteen legs.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She merges
> as reviews converge and asked for evidence when judging one spent; the
> measurement above is that evidence for the finding this overrides.

Issue #47, opened 2026-08-01
A non-struct base's __eq__ shadows the struct base's, and salix binds a structural hash beside it


JPHutchins:
salix decides whether to bind a hash by asking its **struct** base. A non-struct base earlier in the MRO can supply the `__eq__` the class actually resolves, and salix never sees it.

```python
class Base(Struct):
    x: object

class Mixin:
    def __eq__(self, o): return True

class B(Mixin, Base):
    y: object = 0
```

```
B.__mro__[:3]            ['B', 'Mixin', 'Base']
B.__eq__ is Mixin.__eq__ True
B.__hash__               <slot wrapper '__hash__' of 'salix._StructMixin' objects>
B(1,0) == B(2,0)         True
len({B(1,0), B(2,0)})    2
```

Two equal instances, two slots. Same contract violation as #10, reached through a base salix does not inspect.

<details>
<summary><b>Why it is missed</b></summary>

`find_struct_base` returns the first `StructMeta` instance in `bases`, and every question about inheritance is then asked of that one object. `StructType.struct_resolves_body_eq` answers "did a class body supply the `__eq__` this struct resolves" correctly *for the struct chain*, and `bind_hash` trusts it — but `Mixin` is not in that chain, and its `__eq__` wins the MRO anyway.

The same blind spot is behind the first symptom in #9: what salix reads off "the base" is not always what the class actually gets.

</details>

<details>
<summary><b>Shape of a fix</b></summary>

The question to ask is not "what did my struct base do" but "is the `__eq__` this class will resolve one salix bound". For a struct base that is already stored; for anything else it is an attribute lookup against `_StructMixin.__eq__` and `object.__eq__`.

1. **Scan `bases`.** In `StructMeta_new`, treat the class as inheriting a body `__eq__` if any base does — the stored flag for struct bases, a lookup for the rest. Cheap, since non-struct bases are rare and the lookup only runs for them.
2. **Ask the created class instead.** After `type.__new__`, `struct_class`'s `tp_hash` and `__eq__` are both resolved, so the question can be answered exactly once against the finished MRO. But `bind_hash` writes into the namespace *before* the type exists, so this means moving the decision after creation and setting `__hash__` on the type — a larger change, and the reason it is bound in the namespace is that the slots are derived from what the body defines.
[...]

PR #73, opened 2026-08-08
fix: a body __eq__ carries the derived __ne__ with it

### LINKED ISSUE #58: A body `__eq__` wins for `==` but salix keeps its own `__ne__`, so `a == b` and `a != b` are both true
Found while adding a test for #52 that the module docstring promised and the file did not have: a body-defined `__eq__` surviving salix's generated dunders.

It survives for `==`. `__ne__` does not follow it.

```python
class Loose(Struct):
    x: int

    def __eq__(self, other):
        return isinstance(other, Loose)

    def __hash__(self):
        return 0

a, b = Loose(1), Loose(2)
a == b      # True
a != b      # True      <- both
Loose.__ne__  # <slot wrapper '__ne__' of 'salix._StructMixin' objects>
```

Every comparable construction gets this right:

| | `a == b` | `a != b` |
| --- | --- | --- |
| salix | True | **True** |
| `dataclass(eq=False)` + body `__eq__` | True | False |
| a plain class + body `__eq__` | True | False |

## Why

Python's default `__ne__` is derived from `__eq__` — `object.__ne__` calls `__eq__` and inverts unless it returns `NotImplemented` — so a class that defines only `__eq__` gets a consistent `!=` for free. salix binds the mixin's `__ne__` into the namespace alongside the body's `__eq__`, and the mixin's compares structurally, so the two answer different questions about the same pair.

`apply_options`/`rebind` in `src/meta.c` bind all six comparison names, and `bind_hash` already has the logic for noticing that the body defined `__eq__` (`body_defines_eq`). The fix is presumably to let `__ne__` go the way `__hash__` already does when the body writes `__eq__` — leave it unbound so `object.__ne__` derives it.

## Scope

Only when a body defines `__eq__` without also defining `__ne__`. A struct that leaves equality to salix is unaffected, and so is one that writes both. Related to #47, which is the same question one level out — a non-struct *base*'s `__eq__` shadowing the struct base's — but this one needs no unusual bases at all.

Pinned meanwhile in `tests/test_methods.py` (#52) as the current behaviour, so a fix has to update the test deliberately.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked for unit tests covering methods on a struct and issues opened for the gaps; the automated reviewer pointed out that no test pinned a body `__eq__`, and writing one turned this up.

JPHutchins:
Closes #58.

`object.__ne__` calls `__eq__` and inverts unless it returns `NotImplemented`,
so a class that defines only `__eq__` gets a consistent `!=` for free. salix did
not — the mixin is a base, so its **structural** `__ne__` is what the lookup
found beside a body `__eq__` answering a different question:

```python
class Loose(Struct):
    x: int

    def __eq__(self, other):
        return isinstance(other, Loose)

    def __hash__(self):
        return 0

Loose(1) == Loose(2)        # True
Loose(1) != Loose(2)        # True   <- both
```

| | `a == b` | `a != b` |
| --- | --- | --- |
| salix, before | True | **True** |
| salix, now | True | False |
| `dataclass(eq=False)` + body `__eq__` | True | False |
| a plain class + body `__eq__` | True | False |

`test_every_other_construction_derives_ne_from_eq` asserts the bottom two rather
than describing them, because this file has been wrong about a sibling's
behaviour more than once.

## The fix

`bind_not_equal` puts `object.__ne__` into the namespace when the body defines
`__eq__`. Binding is what the mixin being a *base* costs: leaving `__ne__` out of
the namespace does not reach `object.__ne__`, it reaches the mixin's. `rebind`'s
existing skip means a body that wrote its own `__ne__` keeps it.

<details>
<summary>Why it runs before the comparison rebind rather than after.</summary>

A class that also changed the `eq` option would otherwise have the mixin's
`__ne__` put in beside the body's `__eq__` by that rebind — the same bug through
the other door, since `rebind` skips `__eq__` (present in the body) but not
`__ne__` (absent). Running first means `__ne__` is present by the time the
comparison rebind looks, so it is skipped there too.

</details>

## Scoring

The `xfail` this replaces was `strict=True`, so the fix could not land quietly.
Both new tests fail on a build whose `bind_not_equal` returns without binding:

```
FAILED tests/test_methods.py::TestMethods::test_a_body_eq_carries_ne_with_it
FAILED tests/test_methods.py::TestMethods::test_a_subclass_of_it_derives_ne_the_same_way
```

Two more are negative controls that pass either way and would catch an
over-broad fix: a body that writes its own `__ne__` keeps it, and a struct that
leaves equality to salix is untouched.

> [!WARNING]
> **LLM Disclosure**
>
> each PR to be iterated with the automated reviewer; #58 came out of #52's
> review round.

### DIFF
--- a/src/meta.c
+++ b/src/meta.c
@@ -56,6 +56,7 @@ static enum result apply_options(
 	bool inherits_body_eq
 );
 static enum result rebind(PyObject * namespace, char const * const * names, bool from_mixin);
+static enum result bind_not_equal(PyObject * namespace, bool body_defines_eq);
 static enum result bind_hash(
 	PyObject * namespace,
 	struct options options,
@@ -427,6 +428,12 @@ static enum result apply_options(
 	static char const * const representation[] = {"__repr__", NULL};
 	static char const * const mutability[] = {"__setattr__", "__delattr__", NULL};
 
+	/* Before the rebind below, which would otherwise put the mixin's __ne__ in
+	 * beside the body's __eq__ for a class that also changed the eq option. */
+	if (bind_not_equal(namespace, body_defines_eq) != RESULT_OK) {
+		return RESULT_ERROR;
+	}
+
 	if (options.eq != inherited.eq && rebind(namespace, comparison, options.eq) != RESULT_OK) {
 		return RESULT_ERROR;
 	}
@@ -472,6 +479,28 @@ static enum result rebind(
 	}
 
 	return RESULT_OK;
+}
+
+/*
+ * A body that defines __eq__ gets Python's derived __ne__ with it, the way
+ * every other construction does -- object.__ne__ calls __eq__ and inverts,
+ * unless the body wrote a __ne__ of its own, which rebind leaves alone.
+ *
+ * It has to be bound rather than left to the MRO, because the mixin is a base:
+ * its structural __ne__ is what the lookup finds, and it answers a different
+ * question from the body's __eq__. `a == b` and `a != b` were both true. #58.
+ *
+ * Binding it into the dict also puts it ahead of a __ne__ a *co-base* supplies,
+ * which plain Python would let win -- the derived one is the end of the MRO
+ * there, not the front. That is the same shape as #47, where a non-struct
+ * base's __eq__ shadows the struct base's, and it is left with that issue: the
+ * mixin already discarded a co-base's __ne__ before this, so what changes here
+ * is which answer wins rather than whether the co-base's is heard.
+ */
+static enum result bind_not_equal(PyObject * const namespace, bool const body_defines_eq) {
+	static char const * const not_equal[] = {"__ne__", NULL};
+
+	return body_defines_eq ? rebind(namespace, not_equal, false) : RESULT_OK;
 }
 
 /*
--- a/tests/test_methods.py
+++ b/tests/test_methods.py
@@ -166,19 +166,122 @@ class TestMethods:
         assert hash(first) == hash(Loose(99)) == 0
         assert len({first, second}) == 1
 
-    @pytest.mark.xfail(strict=True, reason="#58: salix keeps the mixin's __ne__")
-    def test_a_body_eq_should_carry_ne_with_it(self):
-        """Asserted as it should be rather than as it is, so the assertion goes
-        green the moment #58 is fixed and strict=True makes that impossible to
-        miss.
+    def test_a_body_eq_carries_ne_with_it(self):
+        """The mixin is a base, so its structural __ne__ is what the lookup
+        found beside a body __eq__ that answered a different question, and
+        `a == b` and `a != b` were both true. #58.
         """
 
         Loose = self.loose_equality()
 
         assert (Loose(1) != Loose(2)) is False
 
+    def test_a_subclass_of_it_derives_ne_the_same_way(self):
+        """A subclass writing nothing about equality inherits both halves.
+
+        A regression guard rather than a proof of where the binding goes: the
+        assertion holds whether the derived __ne__ sits in Loose or in Child,
+        since either inverts the same inherited __eq__. What pins the location
+        is the class dict, asserted below.
+        """
+
+        class Child(self.loose_equality()):
+            y: int = 0
+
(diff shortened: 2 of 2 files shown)


JPHutchins:
Round 1: all five taken. Two of them said a test could not see what it claimed
to pin, and both were right — the fix now has three mutations scored against it
rather than one.

## 🔵 the ordering invariant was unenforced and untested

Right, and it is the branch I reasoned about hardest and covered least.
`test_a_body_eq_carries_it_through_an_eq_option_change` takes both directions:
`eq=False` with a body `__eq__`, and a subclass turning `eq=True` back on off
that base — the only shapes that reach the comparison rebind at all.

It fails on a build with `bind_not_equal` moved after that rebind, which is the
reorder the comment was the only thing preventing.

## 🔵 the negative control could not observe the gate

Right, and I had not checked. `object.__ne__` re-dispatches through the mixin's
own `tp_richcompare`, so a structural struct answers identically whether the
binding is there or not — making `bind_not_equal` unconditional passed every
assertion in it. It now asserts `"__ne__" not in vars(Structural)`, which is
the only place the gate is visible, and that assertion fails on the
unconditional build.

## 🔵 the binding shadows a co-base's `__ne__`

Reproduced, and pinned as it is rather than as plain Python has it —
`test_a_co_base_s_ne_is_shadowed_and_that_is_where_47_lives`.

I have not narrowed the binding, and the reason is your own observation that
this is a perpetuated divergence rather than a new one: the mixin's structural
`__ne__` already displaced a co-base's before this, so what changes is which
answer wins, not whether the co-base is heard. Narrowing it means deciding
whether a non-struct base's comparison dunders should beat the struct base's at
all — which is **#47**, filed, and not a question this PR should answer in
passing. The `bind_not_equal` comment now says so, and the test means a decision
on #47 has to change it deliberately.

## ⚪ `bind_hash`'s doc comment was orphaned

Entirely mine — I inserted the new function between a comment and the function
it documents. `bind_not_equal` and its comment now sit above the hash doc, so
each comment is directly over its function again.

## ⚪ the subclass test's docstring overclaimed

Right: the assertion holds whether the derived `__ne__` sits in `Loose` or in
`Child`. Reworded to say it is a regression guard, and the location claim is now
asserted rather than described — `"__ne__" in vars(Loose)` and not in
`vars(Child)`.

## Scoring

Three mutations, against the 34 tests in the file:

```
gate-removed  (bind unconditionally)   2 failed   <- incl. the negative control
not-bound     (never bind)             4 failed
reordered     (bind after the rebind)  1 failed   <- the new eq-option test
```

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Merged at round 2 on **score 1 ≤ 1 — converged**, CI green on `c10709d`, all
sixteen legs, and `git merge-tree` clean onto `main` (GitHub reported
`mergeable: UNKNOWN`, which it does while it recomputes).

## The three findings, and where they go

**The `__ne__`-as-a-field minor is real and not fixed here.** A class with both a
body `__eq__` and a field literally named `__ne__` now fails at creation with
`'__ne__' in __slots__ conflicts with class variable`, where before it built.
Fixing it properly means `bind_not_equal` consulting the field plan, which is
not available where it is called — so it is a small threading change rather
than a one-liner, on a shape nobody has written. Tracked with the two nits
below.

**Both nits are the same class of defect as the two you caught in round 1, and
both are mine:**

- `"__ne__" in vars(self.loose_equality())` builds a *third* `Loose` and
  inspects that. The claim it makes happens to remain true of the shape, but it
  is not the class `Child` was defined over, so it does not say what the
  docstring says it says.
- The `OptedOut` half of the eq-option-change test is invariant to every
  mutation, because `rebind(..., false)` installs `object.__ne__` for
  `eq=False` anyway. Only the `OptedBackIn` half observes the ordering.

That is three rounds running where a test of mine could not observe what its
docstring claimed. It is the thing to watch in this file, not a coincidence.

All three go into one test-accuracy follow-up together with the ordering-test
docstring left open on #70 — they are test-only, they do not belong in a
behaviour PR, and batching them costs one review round instead of three.

> [!WARNING]
> **LLM Disclosure**
>
> why this merged with three open findings and where each one went.

PR #75, opened 2026-08-08
test: say what these three assertions actually pin

JPHutchins:
Test-only. No source change; every behaviour these cover is unchanged.

**Four review rounds running produced the same finding against tests I had just
written** — the docstring claimed something the assertion could not observe.
Once is a slip, four times is a habit, so these land together rather than one
per PR.

| where | the claim | what the assertion could see |
| --- | --- | --- |
| `test_a_subclass_of_it_derives_ne_the_same_way` | "what pins the location is the class dict" | `vars(self.loose_equality())` builds a **third** `Loose` — not the class `Child` was defined over |
| `test_a_body_eq_carries_it_through_an_eq_option_change` | both halves guard the rebind order | only `OptedBackIn` does; `OptedOut` answers correctly either way, because `rebind(..., false)` installs `object.__ne__` for `eq=False` regardless |
| `test_a_shared_ordered_struct_is_safe_to_compare...` | reads as though it guards the race | `ordered is True` is constant — `floor` is below every tuple the writer stores |

<details>
<summary>The third one is a decline, and the reason matters.</summary>

The recommendation on #70 was to make the writer straddle `floor` so a stale or
torn pair read could be observed. That removes what the assertion is for: with a
straddling writer the reader cannot know which value is current, so it can
assert nothing about the result at all.

The constant assertion is the device the sibling test names in its own
docstring — *"the value assertions are there so that a reader which quietly
stopped reading would fail rather than pass"*. It was never a check on the race,
and the docstring now says that rather than leaving a reader to work it out.
The race on that path is not observable at any density I could reach: reverting
just `ordering_result` to a borrowed read survives 3/3, and 3/3 again at five
times the rounds, which is the same result `hash` gave before #46's fix.

</details>

Two `"__ne__" in vars(...)` assertions are new, on `OptedOut` and `OptedBackIn`,
because the dict is the only place the `eq=False` direction of the binding is
visible at all.

`camas check` 25/25.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. Three
> findings across #70's and #73's review rounds said a test could not observe
> what it claimed; both merged on the convergence signal with these recorded as
> follow-ups, and this is that follow-up.

### DIFF
--- a/tests/test_free_threading.py
+++ b/tests/test_free_threading.py
@@ -207,6 +207,14 @@ def test_a_shared_ordered_struct_is_safe_to_compare_while_another_thread_writes_
     often enough. So this pins that ordering keeps working under concurrent
     writes rather than that it would crash without the lock; the crash is pinned
     by the test above, through the helper both paths share.
+
+    `ordered is True` is constant by design and cannot fail in the race
+    direction: floor is below every tuple the writer stores, so a reader that
+    returns a stale but still-above-floor value answers True exactly as a live
+    one does. What it does catch is a read path that raises, or that returns
+    anything other than True. Making the writer straddle floor would not buy the
+    stale case either -- the reader then cannot know which value is current, so
+    it could assert nothing about the result at all.
     """
 
     class Ranked(Struct, frozen=False, order=True):
--- a/tests/test_methods.py
+++ b/tests/test_methods.py
@@ -185,11 +185,13 @@ class TestMethods:
         is the class dict, asserted below.
         """
 
-        class Child(self.loose_equality()):
+        Loose = self.loose_equality()
+
+        class Child(Loose):
             y: int = 0
 
         assert (Child(1) == Child(2), Child(1) != Child(2)) == (True, False)
-        assert "__ne__" in vars(self.loose_equality())
+        assert "__ne__" in vars(Loose)
         assert "__ne__" not in vars(Child)
 
     def test_a_body_eq_carries_it_through_an_eq_option_change(self):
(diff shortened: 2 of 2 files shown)


JPHutchins:
Both right, and there is a joke in it: this PR exists because three docstrings
overstated what their assertions could see, and two of the replacement
docstrings do the same thing.

## 🔵 the ordering checkpoint

Correct, and my paragraph contradicted itself inside three sentences. "The
answer does not depend on which one the reader sees" and "a reader that quietly
stopped reading fails here" cannot both be true — a reader returning a stale but
still-above-floor value answers `True` exactly as a live one does.

Reworded to what actually holds: the checkpoint cannot fail in the race
direction; what it catches is a read path that **raises, or returns anything
other than True**; and it cannot tell a live reader from a stale one. The
straddle note stays, because it does not buy the stale case either.

## 🔵 `"__ne__" in vars(OptedOut)`

Also right. Without the binding, the MRO reaches the mixin's structural `__ne__`
and `OptedOut(1) != OptedOut(2)` is `True` — so the pre-existing answer
assertion already fails in exactly the states the new dict assertion fails in.

Kept, with the credit moved: the answer assertion is what pins the `eq=False`
direction; the `vars()` assertions add only **where** the binding sits, which
the answers cannot show. That is a real if small thing to pin — it is the
difference between "the behaviour is right" and "the behaviour is right for the
reason we think" — but it is not the witness I claimed it was.

`camas check` 25/25.

> [!WARNING]
> **LLM Disclosure**
>

JPHutchins:
Merged at round 2 on **score 1 ≤ 1 — converged**, CI green on `3f66c8c`, all
sixteen legs.

All three findings are correct, and all three are about the precision of a
docstring I wrote about the precision of a docstring. That is the fourth level
of this and a good place to stop recursing — recorded here rather than spent on
another round.

**The minor is right and the correction is one clause.** "The `vars()`
assertions add only where the binding sits, which the answers cannot show" holds
for `OptedBackIn` — delete its own `__ne__` and the answer stays `(True, False)`
because the MRO reaches `OptedOut`'s inherited `object.__ne__`, so only the dict
assertion sees the location. It does not hold for `OptedOut`, whose answer
assertion already fails when its binding goes. So the sentence is true of one of
the two it is written over.

**Both nits are right too.** A raise inside `ordered = floor < shared` never
reaches `assert ordered is True` — `run_on_every_thread`'s failure list is what
catches a raising reader, with or without the assertion; and the assertion
samples one iteration in a thousand, so "catches a non-True answer" should say
"catches one that persists". And "could assert nothing at all" about a
straddling writer overstates: a monotonically increasing one admits a
once-True-stays-True invariant.

None of the three changes what any test does. They go into the `src/` comment
cleanup that JPH asked for in the same breath — she has since scoped that rule
to `src/` only, since tests, logs and commit messages are fair game for prose,
so that PR is the natural home for the wording.

> [!WARNING]
> **LLM Disclosure**
>
> why this merged with three open findings.

Issue #78, opened 2026-08-08
weakref is recorded from the first struct base while the slot can come from any of them


JPHutchins:
Found by #71's round-6 review, filed rather than fixed there on JPH's
instruction to merge and record what is left.

`inherited_options` returns `.weakref = from_behaviour.weakref` — the **first**
struct base's recorded option. The slot itself can come from any base: CPython
refuses a second `__weakref__`, so a `weakref=True` base anywhere in `bases`
gives the class the slot regardless of position. The two therefore disagree:

- a weakref-bearing base in a non-first position leaves the class with
  `tp_weaklistoffset` set and `struct_options.weakref` false, so a subclass
  reading the option is told the wrong thing;
- and because `build_slots` is driven by the option, an explicit `weakref=True`
  can be dropped when the behaviour base already records true.

There is a second route to the same place. `find_struct_base` breaks a
field-count tie by keeping the **first** base, and its comment claims the first
of a tie is the most derived. That is false for `class C(A, B)` with
`class B(A)`, both fieldless: salix picks `A` where CPython's `best_base`
settles on `B`. Layouts are identical so no slot or offset is misplaced, but
`has_weakref_slot(base)` is then asked of the parent rather than the subclass.

An earlier round of #71 fixed this by OR-ing a `tp_weaklistoffset` read across
every base. That was reverted, and rightly: it read the field off unguarded
bases, which is an out-of-bounds read on anything that is not a type. A fix
needs the guard, and needs to decide whether the option follows the slot or the
slot follows the option.

**The comment above `inherited_options` still describes the reverted behaviour**
— it says weakref "is a slot the class has if any base carries one". That is now
false, and goes with the queued `src/` comment cleanup rather than here.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, recording a
> defect left behind by #71 at her instruction to merge and file.

Issue #77, opened 2026-08-08
The frozen pin is armed from one base and directed by another: order-dependent, and it refuses two shapes that used to build


JPHutchins:
Found by the round-3 through round-6 reviews of #71 and **merged with the PR on
JPH's instruction**, so `main` carries it. Two of the three parts are
regressions against the `main` that preceded #71.

## Measured, same probe on both

```
=== main before #71 ===
  M(FrozenFieldless, MutableFields, frozen=False)      built
  M(MutableFields, FrozenFieldless, frozen=False)      built
  B(MutableFieldless, MutableFields, frozen=True)      built
  B(FrozenFieldless, MutableFields, frozen=True)       built

=== after #71 ===
  M(FrozenFieldless, MutableFields, frozen=False)      refused -- a mutable struct cannot inherit from a frozen one
  M(MutableFields, FrozenFieldless, frozen=False)      built
  B(MutableFieldless, MutableFields, frozen=True)      refused -- a frozen struct cannot inherit from a mutable one
  B(FrozenFieldless, MutableFields, frozen=True)       built
```

Both refusals have a mirror ordering that still builds, so the pin is
**order-dependent** — the same two bases accept or refuse depending on which
comes first.

## Three parts, one mechanism

The pin is armed from one base and directed by another.

1. **`base_is_constraining` reads the widest base** while `inherited.frozen` is
   blended: `inherited_options` computes `.frozen = from_behaviour.frozen ||
   promised_frozen`, where `from_behaviour` includes the behaviour base's own
   promise even when that base is *fieldless* and so promised nothing about a
   field. A fieldless frozen base in front therefore arms a refusal against a
   class whose only fielded base is mutable — the first line above.
2. **`options_read` refuses in both directions.** `base_is_constraining &&
   requested.frozen != inherited.frozen` rejects `frozen=True` over a mutable
   base as readily as `frozen=False` over a frozen one. Strengthening is safe
   and has no reason to be refused — that is the third line.
3. **The forced rebind discards a base's own `__setattr__`.** The
   `frozen_across_bases` term fires on "frozen class, some mutable struct base",
   and `rebind` skips only names in the *current* namespace, not inherited ones.
   A base that deliberately defined its own `__setattr__` in its body has it
   replaced by the mixin's whenever any mutable co-base is present.

## What a fix looks like

- `.frozen` should be the promise of the **fielded** bases only; a fieldless
  base has nothing to protect and `options.h` already says so.
[...]

Issue #76, opened 2026-08-08
Options are recorded from one base and answered by the MRO, so a multi-base struct can record one thing and behave as another


JPHutchins:
Split out of #71's round-4 review, where it produced three `major` findings in
one round. It predates that PR: the layout fix there made it easier to hit, not
possible.

## The mistake

salix records a class's options from **one** base and lets the **MRO** answer
for the dunders. Those are different questions, and with more than one struct
base they give different answers.

A struct class binds a dunder into its own dict only when its creation
*transitioned* — `apply_options` rebinds on `options.X != inherited.X` and
otherwise leaves the name to be inherited. So for `class C(A, B)`:

- `A` may bind nothing, because it never transitioned;
- `B` may bind `__repr__`, because it did;
- C3 puts `B` **before** the shared `Struct`/`_StructMixin`, so `B`'s binding
  wins the lookup — even though `A` is the first base and the one whose options
  C inherits.

The class then records one thing and behaves as another.

## Three shapes, each verified on a build

| shape | recorded | behaves |
| --- | --- | --- |
| explicit `eq=True`/`repr=True` where the two bases disagree | the keyword | the other base's binding — the keyword is silently dropped |
| a body `__eq__`/`__hash__` pair on a **later** struct base | salix's value hash | the body's `__eq__` — equal objects hash differently |
| a frozen first base with a mutable fieldless co-base | `frozen=True` | writes succeed |

The middle one is the dict-key invariant, which is the same failure #66 and
#71's round-1 major were about, reached by a third route.

## Why detection has failed three times

#71 tried, in order: rebind when another base's *option* disagrees; also flip
`inherits_body_eq` on that; then split "layout base" from "behaviour base". Each
fixed the face in front of it and the next round found another, because all
three infer *which base answers* from the recorded options — and the recorded
options are exactly what is unreliable.

## The shape of a real fix

Stop inferring and ask. For each name salix owns, walk `bases` and compare what
they actually resolve to against what the option requires:

```c
/* what the MRO will hand this class for `name`, before the class exists */
static PyObject * bases_resolve(PyObject * bases, char const * name);
```

Bind when the resolved object is not the one the option asks for. That is exact,
it collapses all three shapes to one rule, and it needs no reasoning about which
base is "the" base. The cost is a lookup per owned name per class creation,
which should be measured and can be gated on there being more than one struct
base — the single-base path is the common one and is not affected by any of
this.

[...]

PR #80, opened 2026-08-08
test: sweep every combination of two and three struct bases

### LINKED ISSUE #79: Multi-base struct behaviour is sampled, not covered: enumerate the base combinations and assert invariants
#71 took six review rounds and each one found a different shape of the same
mechanism. That is not the reviewer being thorough — it is the tests sampling a
space instead of covering it, so every round I fixed the shape I had thought of
and the next round found one I had not.

The shapes, in the order they arrived:

| round | shape |
| --- | --- |
| 1 | a mutable fieldless base in front of a frozen fielded one |
| 1 | an explicit `repr=True`/`eq=True` on a multi-base subclass |
| 2 | a body `__eq__` on the layout base under a disagreeing co-base |
| 3 | a body `__eq__` on a base that is not the layout base |
| 3 | `frozen=True` strengthening over a mutable fielded base |
| 3 | a weakref slot on a non-layout base |
| 5 | a keyword matching the *layout* base's value |
| 6 | a body `__eq__`/`__hash__` on a base that is not the *first* |
| 6 | a base whose own transition bound a dunder the first base did not |

Every one is two or three struct bases with different options. There are not
many of those — and I already enumerated them once, in a throwaway script that
scored the layout rule over **155 buildable combinations** of two and three
bases drawn from eight shapes. That script would have caught all nine at once.
It was deleted after it answered its one question.

## What to build

Turn the enumeration into a test. A catalogue of base shapes — fieldless,
fielded, weakref, frozen, mutable, `repr=False`, `eq=False`, a body `__eq__`,
a subclass of another — every buildable combination of two and three of them,
and for each the invariants asserted rather than the outputs pinned:

1. `__struct_fields__` is the layout base's fields followed by the new ones,
   and `__base__` is the base they were measured from.
2. What the class **records** matches what it **does** — `struct_options.repr`
   against `repr()`, `.eq` against `==`, `.frozen` against a write attempt,
   `.weakref` against `weakref.ref`.
3. Equal objects hash equal. This one alone covers #66, #71's round-1 major,
   its round-2 minor and its round-6 major.
4. A frozen class refuses every write, by every route.
5. A class that builds is coherent — no state where a recorded option and the
   MRO's answer disagree.

Invariants rather than expected values is the point: a table of expected outputs
for 155 combinations is unreadable and would have to be regenerated by the code
under test, which pins nothing. An invariant says what must be true of *any*
combination, so a shape nobody thought of is covered the day it becomes
buildable.

## Why it is worth doing before the fixes it would guide

#76, #77 and #78 are all this mechanism, and all three will need exactly these
assertions to know when they are done. Written first, they turn three open
issues from "fix and hope the next round agrees" into a red suite that goes
green. Written after, each will grow its own partial version.

The single-base path is not affected by any of this and is already well covered;
this is specifically about more than one struct base.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who observed
> after #71 merged that "test coverage needs to be better if we keep playing
> whack a mole on this". This is that observation made concrete.

JPHutchins:
#71 took six review rounds and every one found a different shape of the same
mechanism. That is the multi-base tests sampling a space instead of covering
it, so each round I fixed the shape in front of me and the next found one I
had not thought of.

This sweeps the space. Eleven base shapes, all **1100** orderings of two and
three of them, **784** of which build, and invariants asserted over every one.

## What holds today

Asserted unconditionally over all 784, so any regression is a red test naming
the combination:

| invariant |
| --- |
| the fields are `__base__`'s followed by the new one |
| every field reads back what it was given |
| `__match_args__` is the fields |
| a write and a delete agree about whether the class is frozen |
| a value that compares by value and can still move is unhashable |

## What does not, and how it is stated

The rest is a **differential**, which is salix's own rule restated: options are
inherited from the first struct base, so a class with several must behave like
the same class built on that base alone.

The combinations #76 can reach are split off by asking `vars()` which bases
bind the dunder — a structural question about the classes themselves, not a
list of known-bad examples — and the tests over that half are `xfail(strict)`.
The fix turns them red.

**The split is exact.** Every contested combination violates, so nothing sound
is parked in a bucket that is expected to fail:

| | sound half | violations | contested half | violations |
| `__repr__` | 620 | 0 | 164 | 164 |
| `__eq__` | 512 | 0 | 272 | 272 |
| `__lt__` | 771 | 0 | 13 | 13 |
| equal ⟹ equal hash | 512 | 0 | 272 | 112 |

<details>
<summary>Why <code>__lt__</code>'s split needs a second clause, and why that is not circular</summary>

Ordering is answered by the recorded option at comparison time — `compare.c`
reads `struct_options.order` off both types — rather than by a binding. So a
later base's `__lt__` only shadows an answer where there was one to shadow,
and the risk predicate needs `the first base alone orders` beside it.

That reads off `build((bases[0],))`, which is the *reference* side of the
differential, not the class under test. Without it the bucket is 164 wide to
cover 13 real violations, and the other 151 would sit in a test that is
expected to fail — where a regression in one of them changes nothing.

</details>

## #77 and #78

One strict xfail each, both measured:

- **#77** — `frozen=` accepted in one base order and refused in the other:
  **188** arrangements. Six unordered pairs at width 2, every one of them
  `MutableFields` against a fieldless base.
- **#78** — `weakref=False` silently ignored over an inherited slot: **251**.
  **This one also reproduces with a single base** (`class C(Weak, weakref=False)`
  builds, and `weakref.ref` still works), so it is wider than #78 as filed.
  CPython cannot take an inherited `__weakref__` away, which makes it a request
  salix should refuse rather than accept and drop.

## The sweep is not decorative

Three mutations of `src/meta.c`, each built and run:

| mutation | caught by | violations |
| --- | --- | --- |
| layout base back to the first struct base (#9's bug) | `test_the_fields_are_the_layout_bases_followed_by_the_new_one` | 357, first named `Fieldless + Fields: ('z',) after Fields('a', 'b')` |
| `bind_hash` drops the `options.eq && !options.frozen` arm | `test_a_value_that_compares_by_value_and_can_still_move_is_unhashable` | 48 |
| `apply_options` never rebinds `__repr__` | the split guard **and** an `XPASS(strict)` | — |

That third one is the interesting one. Removing the rebind means
`Unrepresented` no longer carries `__repr__` in its own dict, so the contested
bucket empties — and rather than going quietly green, both
`test_both_halves_of_the_split_are_reached[__repr__]` and the xfail fail. That
pair of guards exists for exactly this: a bucket that empties in silence is a
suite that stops testing without saying so. (The full suite also catches it
five other ways, in `test_options.py` and `test_multiple_bases.py` — the
differential is blind to a *uniform* break by construction, and those are what
cover it.)

<details>
<summary>Equality is observed three ways, not two</summary>

A body `__eq__` that answers `True` to everything is indistinguishable from the
structural one until it is shown two instances that **differ**. That is the
shape #71's round 6 found and the five rounds before it did not, and it is why
`observe` classifies equality as `everything` / `value` / `identity` rather
than as a boolean. Four of my own probes while building this file were wrong
for want of it, plus one that mutated the instance before comparing it.

</details>

874 tests before, 893 after. `camas check` 25/25 across 3.10–3.15 and 3.14t.

Closes #79.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who said
> after #71 merged that "test coverage needs to be better if we keep playing
> whack a mole on this". This is that made concrete, and it is the first item
> of the agreed next phase.

### DIFF
--- a/tests/test_base_combinations.py
+++ b/tests/test_base_combinations.py
@@ -411,9 +411,11 @@ def test_a_frozen_class_that_compares_by_value_hashes_by_value():
         if not (seen.frozen and seen.equality == "value"):
             continue
 
+        one, twin = instance(cls), instance(cls)
+
         if cls.__hash__ is None:
             violations.append(f"{named(bases)}: unhashable")
-        elif hash(instance(cls)) != hash(instance(cls)):
+        elif hash(one) != hash(twin):
             violations.append(f"{named(bases)}: equal instances hash differently")
 
     assert violations == []
@@ -550,13 +552,68 @@ def behaviour_differs_from_the_first_base_alone(
 
 
 def hash_disagreements(combinations: tuple[Combination, ...]) -> list[str]:
-    return [
-        named(bases)
-        for bases, cls in combinations
-        if cls.__hash__ is not None
-        and instance(cls) == instance(cls, 100)
-        and hash(instance(cls)) != hash(instance(cls, 100))
-    ]
+    """Both pairs, because either can be the equal one.
+
+    Two instances built from *differing* arguments are equal only where a body
+    __eq__ says everything is, and two built from the *same* arguments are
+    equal wherever equality reads the fields. Probing only the first left the
+    invariant firing on 111 of 672 sound combinations, and on 87 of those the
+    hash is a constant.
+
+    Both instances of a pair are held while they are compared: hashed one at a
+    time, the first is freed before the second is allocated and the allocator
+    hands back the same address, so an identity hash reads as a matching one.
+    """
+
+    disagreements = []
+
+    for bases, cls in combinations:
+        if cls.__hash__ is None:
+            continue
+
+        for left, right in ((instance(cls), instance(cls)), (instance(cls), instance(cls, 100))):
+            if left == right and hash(left) != hash(right):
+                disagreements.append(named(bases))
+                break
+
+    return disagreements
+
+
+def hashability_disagrees_with_equality(
+    combinations: tuple[Combination, ...],
+) -> list[str]:
+    """Whether a class can be hashed at all must follow the equality it
+    answers with -- identity equality is hashable, value equality is hashable
+    exactly when the value cannot move, and an equality a body supplied brings
+    its own rule.
+
+    The unhashable rule beside this one keys off *observed* value equality, so
+    it steps over a class that records eq=True, is made unhashable for it, and
+    then answers identity because a later base binds `__eq__`. That leaves an
+    identity-comparing class that cannot be put in a set, and nothing said so.
+    """
+
+    violations = []
+
+    for bases, cls in combinations:
+        seen = observe(cls)
+        hashable = cls.__hash__ is not None
+
+        if seen.equality == "identity" and not hashable:
+            violations.append(f"{named(bases)}: identity equality, unhashable")
+        elif seen.equality == "value" and hashable != seen.frozen:
+            violations.append(f"{named(bases)}: value equality, hashable={hashable}")
+
+    return violations
+
+
+def test_hashability_follows_the_equality_the_class_answers_with():
+    assert hashability_disagrees_with_equality(SOUND_EQ) == []
+
+
+@pytest.mark.xfail(strict=True, reason="#76: equality answered by a later base, hash settled from the record")
+def test_hashability_follows_equality_when_a_later_base_answers_it():
+    assert hashability_disagrees_with_equality(CONTESTED_EQ) == []
 
 
 def test_the_repr_is_the_first_struct_bases():
--- a/tests/test_base_combinations.py
+++ b/tests/test_base_combinations.py
@@ -1,0 +1,481 @@
+"""Every buildable combination of two and three struct bases, and what has to
+be true of all of them.
+
+The tests beside this one in test_multiple_bases.py pin shapes: this base in
+front of that one answers so. Nine of those shapes arrived one review round at
+a time while #9 was being fixed, each the same mechanism wearing a different
+hat, because a handful of examples samples the space rather than covering it.
+
+So these say what must hold of *any* combination, and sweep the whole space
+looking for somewhere it does not. A shape nobody thought of is covered the day
+it becomes buildable.
+
+The rule the differential tests state is salix's own: options are inherited
+from the **first** struct base, whose branch of the MRO is searched first, and
+only `frozen` and `weakref` are facts about every base rather than preferences
+of that one. So a class with several struct bases must behave like the same
+class built on its first struct base alone.
+
+Where that fails today it fails for one reason, filed as #76: a struct base
+binds a dunder only where its own creation transitioned an option, so a later
+base can bind a name the first one left to the MRO, and then the record and the
+behaviour are two different answers. Those combinations are split off by asking
+`vars()` which bases bind the name -- a structural question, not a list of
+known-bad examples -- and the tests over them are strict xfails, so #76 turns
+them red the day it lands.
+"""
+
+import itertools
+import operator
+import weakref
+from typing import Any, NamedTuple
+
+import pytest
+
+from salix import Struct
+
+
+class Fieldless(Struct):
+    pass
+
+
+class MutableFieldless(Struct, frozen=False):
+    pass
+
+
+class Weak(Struct, weakref=True):
+    pass
+
+
+class Fields(Struct):
+    a: int
+    b: int
+
+
+class OneField(Struct):
+    a: int
+
+
+class MutableFields(Struct, frozen=False):
+    m: int
+
+
+class Unrepresented(Struct, repr=False):
+    pass
+
+
+class ByIdentity(Struct, eq=False):
+    pass
+
+
+class Ordered(Struct, order=True):
+    o: int
+
+
+class BodyEq(Struct):
+    def __eq__(self, other: object) -> bool:
+        return True
+
+    def __hash__(self) -> int:
+        return 7
+
+
+class Derived(Fieldless):
+    pass
+
+
+SHAPES = (
+    Fieldless,
+    MutableFieldless,
+    Weak,
+    Fields,
+    OneField,
+    MutableFields,
+    Unrepresented,
+    ByIdentity,
+    Ordered,
+    BodyEq,
+    Derived,
+)
+
+ARRANGEMENTS = tuple(
+    arrangement
+    for width in (2, 3)
+    for arrangement in itertools.permutations(SHAPES, width)
+)
+
+_names = itertools.count()
+
+
+class Refused(NamedTuple):
+    """Salix declined the class."""
+
+    message: str
+
+
+class Impossible(NamedTuple):
+    """Python declined it, which is not this suite's business: an MRO or
+    lay-out conflict says these bases could never have been combined, whoever
+    was building them.
+    """
+
+    message: str
+
+
+Outcome = type | Refused | Impossible
+Combination = tuple[tuple[type, ...], type]
+
+PYTHONS_OWN = ("consistent method resolution", "lay-out conflict")
+NEW_FIELD = "z"
+
+
+def build(bases: tuple[type, ...], field: str = NEW_FIELD, **keywords: bool) -> Outcome:
+    try:
+        return type(Struct)(
+            f"Combination{next(_names)}",
+            bases,
+            {"__annotations__": {field: int}},
+            **keywords,
+        )
+    except TypeError as failure:
+        if any(reason in str(failure) for reason in PYTHONS_OWN):
+            return Impossible(str(failure))
+        return Refused(str(failure))
+
+
+def is_a_class(outcome: Outcome) -> bool:
+    return isinstance(outcome, type)
+
+
+def instance(cls: type, start: int = 0) -> Any:
+    return cls(*range(start, start + len(cls.__struct_fields__)))
+
+
+class Behaviour(NamedTuple):
+    """What a class does, as opposed to what it recorded. Only the second is
+    salix's to choose, and the whole of #76 is the two disagreeing.
+    """
+
+    repr: bool
+    equality: str
+    order: bool
+    frozen: bool
+    weakref: bool
+
+
+def observe(cls: type) -> Behaviour:
+    """`equality` is three-valued because equality has three possible sources.
+    A body `__eq__` that answers True to everything looks exactly like the
+    structural one until it is shown two instances that differ, which is the
+    shape #71's round-6 review found and four rounds before it did not.
+    """
+
+    one, twin, other = instance(cls), instance(cls), instance(cls, 100)
+
+    try:
+        operator.lt(one, twin)
+        orders = True
+    except TypeError:
+        orders = False
+
+    try:
+        weakref.ref(one)
+        weak = True
+    except TypeError:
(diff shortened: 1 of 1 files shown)


JPHutchins:
All four minors are right, and they are all the same kind of thing: the
invariants hold, and the *harness* had ways to stop covering the space without
saying so. That is the failure this file exists to prevent, so they are worth
more than their severity.

## The sweep could shrink in silence

Right, and the second trigger you found is the better one — `PYTHONS_OWN` is a
substring test over CPython's wording, so a release that rephrases an MRO or
lay-out error starts calling its own refusals salix's, and the combination
leaves `ALL` with nothing noticing.

Both are one assertion, because **salix refuses none of these 1100 arrangements
today**: 784 build, 316 are Python's own, 0 are salix's. So any refusal at all
is either a new constraint or a misclassified message, and either way it is
named rather than absorbed.

```
test_salix_refuses_no_arrangement_of_these_shapes
```

Verified by breaking the heuristic exactly as you describe —
`PYTHONS_OWN = ("a phrase a future CPython no longer uses",)`:

```
FAILED test_salix_refuses_no_arrangement_of_these_shapes
Left contains 316 more items, first extra item:
  'Fieldless + Derived: Cannot create a consistent method resolution order (MRO) for bases Fieldless, Derived'
```

The old guard stayed green through that.

## The contested buckets emitted one bit

Correct, and it is the tension I built in. An aggregate `== []` over a bucket
where everything already fails flips only when the bucket becomes *entirely*
clean.

The other bound closes it: each bucket now also asserts it is **wholly** broken.
Between the two, the count cannot move without a test noticing — a partial fix
drops it below the bucket size, a new violation among sound combinations moves
it out of the sound half's `== []`.

That required narrowing the hash bucket until it was exact, which your finding
is what prompted:

<details>
<summary>224 holding 112, and why an identity binder cannot produce the defect</summary>

A later base binding `__eq__` is what puts the record and the behaviour at odds,
but it is not sufficient for a *hash* disagreement. `ByIdentity` binds
`object.__eq__`, and object's equality still calls two differing instances
unequal — so there is no equal-and-differently-hashed pair to find. Only a
binder that answers **everything** equal produces one.

Adding that clause takes the bucket from 224 holding 112 real violations to
**112 holding 112**, and moves the other 112 into the sound half, which now
covers 672 rather than 512.

All four buckets are exact now:

| | sound | contested | violations |
| --- | --- | --- | --- |
| `__repr__` | 620 | 164 | 164 |
| `__eq__` | 512 | 272 | 272 |
| `__lt__` | 771 | 13 | 13 |
| `__hash__` | 672 | 112 | 112 |

</details>

The new tests say to delete themselves alongside the xfails when #76 lands —
they describe one defect from opposite sides.

## frozen was recorded and never checked against behaviour

The sharpest of the four. You are right about each half of why nothing caught
it: the differential excludes `frozen` deliberately, the write/delete test only
compares the two to each other, and the unhashable rule fires only for value
equality.

Asserted directly now — frozen if the first base alone is, or if any base with
fields is, which is the rule `inherited_options` implements:

```
test_a_frozen_promise_is_kept_by_every_arrangement_that_inherits_one
```

Dropping `frozen_across_bases` from `apply_options` makes it name **255**
combinations, first `Fieldless + MutableFieldless: frozen=False against a
promise`. (The unhashable rule fires on that mutation too, for the value-
equality subset — so it was partially caught, but not named and not completely.)

## #78's single base

No coverage anywhere, exactly as you say — the sweep is widths 2 and 3 and
`class C(Weak, weakref=False)` is width 1. It is now in the xfail's sweep *and*
in a test of its own, so a fix aimed only at the multi-base reading cannot
leave it quietly passing.

## The nit

Bound once.

---

21 passed, 6 xfailed. `camas check` 25/25 across 3.10–3.15 and 3.14t.

> [!WARNING]
> **LLM Disclosure**
>
> round 1.

JPHutchins:
Both minors are right, and they are the same shape as round 1's second finding:
a way the sweep can cover less than it says it does. Score 4 → 2, and these
close the last two paths I know of.

## The weakref bucket had no size

Correct, and your trigger (2) is the one that bites: a partial #78 fix makes
`is_a_class(without)` false for the arrangements it refuses, those drop out of
the violation list, and the strict xfail stays failing exactly as expected.

Bounded the same way the #76 buckets were — the bucket is the combinations
where some base carries the slot (**252**, including the single-base `(Weak,)`
case), and the violation count must equal its size.

**Simulated your scenario to check it fires.** A partial fix that refuses
`weakref=False` only where the *first* base carries the slot:

```
XFAIL test_the_weakref_option_and_the_slot_agree   <- unchanged, invisible
FAILED test_every_class_carrying_an_inherited_slot_really_does_ignore_the_request
```

Which is the finding, demonstrated.

<details>
<summary>The propagation direction is asserted too, and it is the weaker half — worth saying so rather than claiming two guards</summary>

I added `test_a_weakref_slot_on_any_base_reaches_the_class` for your trigger
(1), then tried to break it from salix and could not. Both directions are
caught by CPython before any test sees them:

- **dropping the option** (`.weakref = false` in `inherited_options`) changes
  nothing observable — the slot is inherited through the MRO, not granted by
  the option, which is #78's whole point;
- **adding one anyway** (`build_slots(new_names, true)`) is
  `TypeError: __weakref__ slot disallowed: we already got one` at class
  creation, before collection finishes.

So it states the documented "any base carries one" rule and would catch a #78
fix that changed inheritance semantics, but it is not what closes your finding.
The bound is.

</details>

## The Impossible side still absorbed

Right, and the specific risk you name is the real one: salix is a layout
library, so `lay-out conflict` is a phrase it could plausibly use in a refusal
of its own, and that refusal would be filed as Python's and vanish.

`len(ALL) == 784` is now asserted, with the split: **1100 arrangements, 784
buildable, 316 Python refuses outright.** Measured identical on 3.10, 3.11,
3.12, 3.13, 3.14 and 3.15, so it is not a version artefact.

**This is the one expected value in the file, and I want to be explicit that it
is one**, because the issue this PR closes argues against pinning outputs. The
distinction I would draw: it pins the *denominator*, not an outcome. Every
other assertion here is `violations == []` over `ALL`, and a sweep that quietly
covers 600 combinations reports exactly the same green as one that covers 784 —
so without a bound on the space, the invariants are only as strong as their
input, unverified. Absorbing 38 combinations into `Impossible` makes it read
`746 == 784`.

The docstring says a change here is something to look at rather than to
re-baseline.

## The nit

Right, and it was a regression of round 1's own fix — I removed a double
`observe` and then wrote a test doing up to five. All the per-shape
observations are hoisted to one `BEHAVIOUR_ALONE`, with `ORDERS_ALONE`,
`EQUALITY_ALONE` and `FROZEN_ALONE` derived from it.

---

24 passed, 6 xfailed. `camas check` 25/25 across 3.10–3.15 and 3.14t.

> [!WARNING]
> **LLM Disclosure**
>
> round 2.

JPHutchins:
All four are right. Three of them are this file's own methodology applied back
at it, which is fair — I set the pattern in round 2 and then did not finish it.

## #77 was the one bucket without a bound

Correct, and your partial-fix case is the right one to worry about: settle the
six width-2 pairs, leave the 182 width-3 arrangements, and the xfail keeps
failing exactly as expected.

Bounded — and **structurally rather than as a pinned 188**, which took finding
the predicate:

> an ordering is at risk when reversing it changes the `frozen` it inherits,
> **and** some base has fields

Both clauses are load-bearing. The second is the one I would have missed:
`options_read` only refuses when the layout base is *constraining*, so an
all-fieldless arrangement can inherit a different `frozen` reversed and still
be accepted both ways. Without it the bucket is **324 wide to hold 188**. With
it, 188/188 exact.

Simulated the partial fix — `XFAIL` unchanged,
`test_every_order_dependent_frozen_refusal_really_is_one` red. All four buckets
now carry the bound.

## The hash direction nothing could reach

This is the sharpest of the four and I had not seen it. Both guards are
one-way, for the two different reasons you give: `hash_disagreements` needs two
*equal* instances and a frozen value-equal class never produces a pair from
differing arguments, and the unhashable rule skips frozen classes outright.
**290 combinations with nothing said about them.**

`test_a_frozen_class_that_compares_by_value_hashes_by_value` asserts it in both
halves — hashable at all, and equal instances hashing equal. Mutating
`bind_hash` to `if (body_defines_eq || options.eq)` — unhashable for every
eq=True class, frozen included — fails it, where before it passed the sweep.

You are also right that my mutation table only covered the mutable direction.

## The redundant re-derivation

Right, and it is the same hoist I did last round without carrying it into the
main helper. `BEHAVIOUR_ALONE[bases[0]]` now, across all eight call sites.

## The nit — and it corrects something I said, not something I wrote

You have caught me overstating a guard **in last round's reply**. I said the
propagation test covered your trigger (1). It does not, and I had already found
that myself when the mutation would not take — I wrote "both directions are
caught by CPython before any test sees them" and then left the docstring
claiming the direction CPython owns.

The test only ever caught the opposite: a class weak-referenceable when no base
made it so — `build_slots` appending a slot over a base that has one, or a
struct class growing a `__dict__`. It now says that and asserts only that.

---

26 passed, 6 xfailed. `camas check` 25/25 across 3.10–3.15 and 3.14t.

> [!WARNING]
> **LLM Disclosure**
>
> round 3.

JPHutchins:
All three are tests that pass and cannot detect what they claim, which is worse
than not having them. The first is the sharpest finding this PR has had.

## The identity-hash probe could not see an identity hash

You are right and the mechanism is exactly as you describe. Reproduced:

```
identity-hash detected: sequential 0/500, both-held 500/500
```

`hash(instance(cls)) != hash(instance(cls))` frees the first temporary before
allocating the second, pymalloc hands back the same address, and the two
id-based hashes match. Holding both while they are compared fixes it, and that
is now written down in `hash_disagreements` too, since it is a trap for any
future probe here.

## Equal-implies-equal-hash was asserted almost nowhere it could fail

Measured, and it is as bad as you say: of 672 sound combinations the probe
fired on **111**, and 87 of those hash to the constant 7.

The cause is that it only ever built a pair from *differing* arguments, which
is equal only where a body `__eq__` says everything is. Two instances built
from the **same** arguments are equal wherever equality reads the fields — the
561 combinations that were being skipped. Both pairs are probed now:

| | before | after |
| --- | --- | --- |
| sound combinations the probe fires on | 111 | **377** |
| contested violations | 112 | 112 (bucket still exact) |

## The unhashable rule stepped over 24 broken combinations

This one turned out to be more than a nit, and thank you for it. Those 24 —
`(MutableFieldless, ByIdentity)` and friends — record `eq=True`, are made
unhashable for it, then answer identity because a later base binds `__eq__`.
An identity-comparing class that cannot go in a set.

They were sitting in the **sound** half, so it was a hole in the split rather
than only a gap in coverage.

The invariant that closes it states hashability against the equality the class
answers with:

> identity equality is hashable; value equality is hashable exactly when the
> value cannot move; an equality a body supplied brings its own rule

**0 violations across the whole sound-eq half (512), all 24 in the contested
one** — so it splits the way everything else in this file does, and #76 turns
it red. The bucket bound comes with it.

---

27 passed, 7 xfailed. `camas check` 25/25 across 3.10–3.15 and 3.14t.

> [!WARNING]
> **LLM Disclosure**
>
> round 4.

PR #81, opened 2026-08-08
feat: name the metadata _struct_fields_, keep msgspec's name as an alias

### LINKED ISSUE #74: __struct_fields__
did we invent this dunder? that is not allowed, standard practice is a sunder `_struct_fields_`

JPHutchins:
> did we invent this dunder? that is not allowed

We did, and it is not. So the pair salix defines is now `_struct_fields_` /
`_struct_defaults_`, and `__struct_fields__` / `__struct_defaults__` stay
beside them as **msgspec's** names, honoured rather than invented.

## Why the two are not the same kind of thing

The language reference reserves `__name__` for "the interpreter and its
implementation, including the standard library", and says any other use is
"subject to breakage without warning". PEP 8 puts the same rule as: **never
invent such names; only use them as documented.**

That is an asymmetry, not a prohibition, and it is the whole of the ruling
here. Minting `__struct_fields__` for salix is inventing one. Answering to
msgspec's `__struct_fields__` is using a documented one, which is the half PEP 8
allows — and it is what makes code written against msgspec work unchanged.

<details>
<summary>msgspec's names and semantics, measured rather than assumed</summary>

```
msgspec 0.21.1
class attrs: ['__struct_config__', '__struct_defaults__',
              '__struct_encode_fields__', '__struct_fields__']
__struct_fields__:   ('x', 'y')
__struct_defaults__: (2,)
on an instance:      ('x', 'y')
sunder present?      False
```

Same two names, same semantics salix has — the field-name tuple and the
trailing-defaults tuple, on both the class and the instance. msgspec also
carries `__struct_config__` and `__struct_encode_fields__`; salix has no
equivalent and does not invent one.

</details>

## What changed

| | |
| --- | --- |
| `src/mixin.c` | the instance-side getset table grows from two entries to four |
| `src/meta.c` | the class-side table likewise |
| `salix/__init__.pyi` | all four declared `Final` |
| `tests/` | 83 uses renamed to the sunder — 59 `__struct_fields__`, 24 `__struct_defaults__` |
| `tests/test_metadata_names.py` | new: the aliases are a surface with its own tests, not a spelling |

The two parametrised tests in `test_struct_identity.py` that sweep the metadata
attributes now sweep all four, so the impostor's `AttributeError` and the
"never handed the metaclass" guard both cover the aliases.

<details>
<summary>Why the mixin needs four getters where the metaclass needs two</summary>

The metaclass getters take the class and read it. The mixin's have to refuse a
non-struct — the mixin is a permitted base and a subclass of it need not be one
— and the refusal names the attribute asked for:

```
_struct_fields_ is defined on structs, and Impostor is not one
```

The getset machinery gives the getter no way to tell which of its names was
used, and the only channel that would (`closure`) needs a cast to `void *`. Four
one-line getters is the cast-free version, and it keeps the message honest:
ask for the dunder and the dunder is what the error names.

</details>

## One line of unrelated cleanup, declared

`src/construct.c`'s comment said `__struct_defaults__ advertised a value no
instance would ever carry (#56)`. The rename touches that sentence, so the
stale issue reference went with it rather than being left in a line being
rewritten. The other 13 issue references in `src/` are still queued for their
own src-only PR, untouched here.

902 tests pass, 10 skipped, 2 xfailed. `camas check` 25/25 across 3.10–3.15 and
3.14t; `camas flake_check` green, which is the only local thing that
cross-builds Windows.

Closes #74.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked
> "did we invent this dunder?" on #74 and then ruled: keep the compliant name,
> alias the non-compliant one.

### DIFF
--- a/salix/__init__.pyi
+++ b/salix/__init__.pyi
@@ -4,6 +4,8 @@ from typing_extensions import dataclass_transform
 
 @dataclass_transform(frozen_default=True)
 class Struct:
+    _struct_fields_: Final[tuple[str, ...]]
+    _struct_defaults_: Final[tuple[Any, ...]]
     __struct_fields__: Final[tuple[str, ...]]
     __struct_defaults__: Final[tuple[Any, ...]]
     def __init_subclass__(
--- a/src/construct.c
+++ b/src/construct.c
@@ -89,7 +89,7 @@ PyObject * Struct_vectorcall(
  * What a class whose body writes __init__ allocates with. That class declined
  * the generated constructor, and fill_defaults went with it: a declared default
  * was never written, for the class and for every subclass, so
- * __struct_defaults__ advertised a value no instance would ever carry (#56).
+ * _struct_defaults_ advertised a value no instance would ever carry.
  * Writing them here means the __init__ runs over a struct that already holds
  * them and overwrites whatever it means to -- which is where a dataclass leaves
  * them too, on the class, readable.
--- a/src/fields.c
+++ b/src/fields.c
@@ -754,7 +754,7 @@ static PyObject * build_defaults(PyObject * const all_names, PyObject * const de
 		 * declaration, so of course it is) and refused there, so the race costs
 		 * the work the first check exists to skip and not the invariant.
 		 *
-		 * `__struct_defaults__` still hands the stored object out, so filling it
+		 * `_struct_defaults_` still hands the stored object out, so filling it
 		 * through there defeats this. That route is out of contract and #51 is
 		 * where the whole type list is argued. */
 		if (reject_unsafe_default(field_name, value) != RESULT_OK) {
--- a/src/meta.c
+++ b/src/meta.c
@@ -131,17 +131,30 @@ PyTypeObject StructMeta_Type = {
 
 /* The mixin answers these for an instance; the metaclass answers the same
  * questions of the class, which is where msgspec puts them and so where a
- * reader looks first. */
+ * reader looks first. Both spellings, for the reason src/mixin.c gives: the
+ * sunder is salix's, and the dunder is msgspec's name honoured rather than a
+ * dunder of salix's own invention. Unlike the mixin's, these getters need no
+ * name of their own -- there is no non-struct for them to refuse. */
 static PyGetSetDef StructMeta_getset[] = {
 	{
-		.name = "__struct_fields__",
+		.name = "_struct_fields_",
 		.get = StructMeta_get_field_names,
 		.doc = "tuple of field names",
 	},
 	{
-		.name = "__struct_defaults__",
+		.name = "_struct_defaults_",
 		.get = StructMeta_get_defaults,
 		.doc = "tuple of trailing defaults",
+	},
+	{
+		.name = "__struct_fields__",
+		.get = StructMeta_get_field_names,
+		.doc = "tuple of field names, under msgspec's name for it",
+	},
+	{
+		.name = "__struct_defaults__",
+		.get = StructMeta_get_defaults,
+		.doc = "tuple of trailing defaults, under msgspec's name for it",
 	},
 	{.name = NULL},
 };
--- a/src/mixin.c
+++ b/src/mixin.c
@@ -10,6 +10,8 @@
 static int Struct_set_attribute(PyObject * self, PyObject * name, PyObject * value);
 static PyObject * Struct_get_field_names(PyObject * self, void * closure);
 static PyObject * Struct_get_defaults(PyObject * self, void * closure);
+static PyObject * Struct_get_fields_as_msgspec(PyObject * self, void * closure);
+static PyObject * Struct_get_defaults_as_msgspec(PyObject * self, void * closure);
 static PyObject * metadata_of(PyObject * self, enum struct_metadata which, char const * name);
 static PyGetSetDef Struct_getset[];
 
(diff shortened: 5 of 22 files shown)


JPHutchins:
**The minor is right and the important part of it is that I wrote a false
claim.** The docstring said the metaclass/mixin arrangement "stops a subclass
shadowing one spelling and leaving the other to report something the first no
longer agrees with". It does not, and your reproduction is exact:

```
class Weird(Struct):
    _struct_fields_: int
    x: int

Weird._struct_fields_        ('_struct_fields_', 'x')   <- the metadata
Weird(5, 6)._struct_fields_  5                          <- the field
```

The claim is gone. The behaviour is pinned instead —
`test_a_body_that_takes_one_of_these_names_is_read_two_ways`, over all four
spellings — so it is a decision rather than a discovery.

**Filed as #82 rather than fixed here.** Refusing a colliding field name is a
behaviour change beyond this PR: it would also refuse `__struct_fields__` as a
field name, which works today. The issue lays out the three options and
recommends the contained one.

Worth separating the two halves, because only one is this PR's doing: the
asymmetry is **pre-existing** — verified on the same build, `Dunder.__struct_fields__`
is `('__struct_fields__', 'x')` and `Dunder(5, 6).__struct_fields__` is `5`.
What #74 adds is that two names that were ordinary now behave this way, which
is the cost of reserving them and is the part worth a ruling.

## The nits

Three fixed as described: the `two getsets` comment now says four and names
which two are which; the `hasattr` guard is gone; `msgspecs_names` →
`msgspec_names`.

**The fourth is right and I have made it carry more than you asked for.** You
are correct that comparing the two class-side spellings is true by construction
— `src/meta.c` wires both to `StructMeta_get_field_names` — and that a fields
getter returning the defaults tuple for *both* would have survived it. Rather
than delete it, both spellings are now checked against the value itself, on the
class and on the instance:

```python
EXPECTED = (
    ("_struct_fields_", ("x", "y")),
    ("__struct_fields__", ("x", "y")),
    ("_struct_defaults_", (2,)),
    ("__struct_defaults__", (2,)),
)
```

The instance-side twin is where it earns the most, for exactly the reason you
give: the mixin wires the four names to four distinct functions, so a crossed
`which` enum is visible there and nowhere else.

<details>
<summary>One small correction on nit 3 — the docstring you quoted is not about the guard</summary>

You read `"the name says which of the two it pins, now that both exist"` as
implying the attribute's presence is conditional. It is not about presence: "the
two" is the *copying* path and the *sharing* path, and "both exist" is since
#19 gave empty containers a copy. The test name says which of those two it
pins.

The guard is gone regardless — you are right that it could never be false — but
the comment predates this PR and reads correctly, so it stays.

</details>

902 tests pass, 10 skipped, 2 xfailed. `camas check` 25/25 across 3.10–3.15 and
3.14t.

> [!WARNING]
> **LLM Disclosure**
>
> round 1 and correcting a false claim in a docstring the review found.

JPHutchins:
Finding 2 is a real one I had not seen, and it is the interesting half of #82
rather than a separate problem. Findings 1 stands as deferred; the five nits
are fixed.

## The stub already refuses what the runtime allows

Confirmed, and it is four errors rather than two — both sunders and both
dunders, two error codes each:

```
error: Incompatible types in assignment (expression has type "int",
       base class "Struct" defined the type as "tuple[str, ...]")  [assignment]
error: Cannot assign to final name "_struct_fields_"  [misc]
```

You are right that the *sunder* half is new: before this PR those two names
were undeclared and a field of that name checked cleanly.

**I want to keep it, and the reason is the direction it errs in.** The runtime
builds that class and then misreads it — metadata from the class, the field
from an instance — so a checker refusing it is the reservation being real
somewhere. Erring toward refusing a shape that misbehaves is the safe way for
the two to disagree.

What it should not be is an *accident*, which is what your finding correctly
identifies it as. So it is stated where the repo states this kind of thing —
`tests/typing/rejected.py`, run under mypy and pyright with
`--warn-unused-ignores`, so the `type: ignore`s fail the build if any of the
four stops being refused:

```python
class AFieldMayNotTakeOneOfTheMetadataNames(Struct):
    _struct_fields_: int  # type: ignore[assignment,misc]
    ...
```

**Recorded on #82 as the static half**, because it narrows the ruling: option 1
is already half-implemented statically, and whichever way it goes the two
halves have to be made to agree — rule for refusing and the runtime catches up;
rule for allowing and the *stub* is what changes, because it currently refuses
code salix accepts.

## Finding 1

Unchanged and deferred to #82, as you note. Refusing a colliding field name at
runtime would also refuse `__struct_fields__`, which works today, so it is a
ruling rather than a fix.

## The nits

All five, and two of them were mine from round 1 rather than from the original
PR:

| | |
| --- | --- |
| the plain class-variable shape described but unpinned | pinned — it reaches the same disagreement by a different route, since `drop_class_variables` strips only *annotated* names, so an unannotated assignment stays in the class dict |
| dead `default` local | gone. Round 1 removed the `hasattr` guard around it and left the local; you are right that the `del` did nothing, because `Defaulted` holds the tuple |
| dead `PAIRS` | gone, leftover from the first draft |
| subclass alias assertions in the relative form | absolute. Round 1 removed exactly this elsewhere and I left these two |
| four names enumerated three times | the two in `test_struct_identity.py` are one `METADATA_NAMES` constant now |

On the third enumeration: sharing across modules needs a `conftest.py` fixture,
and `tests/conftest.py` exists for one specific reason — keeping the in-place
build importable without putting the working tree on `sys.path` ahead of an
installed wheel. Putting test data in it seemed a worse trade than one
duplicated tuple, so I left that one. Happy to be told otherwise.

---

`camas check` 25/25 across 3.10–3.15 and 3.14t.

> [!WARNING]
> **LLM Disclosure**
>
> round 2.

PR #83, opened 2026-08-08
fix: ask every base which __eq__ the class will resolve

### LINKED ISSUE #47: A non-struct base's __eq__ shadows the struct base's, and salix binds a structural hash beside it

JPHutchins:
salix decides whether to answer for `__hash__` by asking whether the `__eq__`
the class resolves is one it bound. That question was put to the **struct** base
alone, and a non-struct base earlier in the MRO can supply the `__eq__` the
class actually gets:

```python
class Base(Struct):
    x: object

class Mixin:
    def __eq__(self, o): return True

class B(Mixin, Base):
    y: object = 0
```

| | before | after |
| --- | --- | --- |
| `B.__eq__ is Mixin.__eq__` | True | True |
| `B(1,0) == B(2,0)` | True | True |
| `B.__hash__` | salix's structural hash | `None` |
| `len({B(1,0), B(2,0)})` | **2** | `TypeError: unhashable type: 'B'` |

Two objects that compared equal and took two slots in a set, so neither could
find the other in a dict. Same contract break as #10, reached through a base
salix never looked at.

## The change

The question is now put to every base, in the order C3 searches them, and
answered by the first whose branch supplies an `__eq__` at all:

- a **struct base** always supplies one — the mixin's, or `object`'s where that
  base opted out — so the search stops there and the flag it already stores is
  the answer;
- a **non-struct base** answers only when it resolves something other than
  `object`'s or the mixin's; otherwise it adds nothing to the lookup and the
  next base decides.

This is the issue's option 1, and the shape it recommended. The lookup is
fallible — asking a base runs whatever its metaclass defines — so it returns a
tagged result and `build_struct_class` propagates rather than demoting, the
same call the #44 review settled.

`B` becoming unhashable is **Python's own rule**, not a new one salix invented:
a class body that defines `__eq__` and not `__hash__` is unhashable, and
`class Plain(Mixin): pass` has been `__hash__ is None` all along. A co-base that
defines both keeps its pair — `hash` 7, one set slot — because equality and the
hash then agree, which is the entire requirement.

<details>
<summary>The regression this caused once, and why the test for it is in the file</summary>

The first version compared each base's `__eq__` against `object.__eq__` alone.
`_StructMixin` is the only base `Struct` itself has, **and its metaclass is
plain `type`** — so `is_struct_class` is false for it and it fell through to
that comparison. The mixin's `__eq__` is not `object`'s, so `Struct` was built
recording "my equality came from a class body", and every struct in the world
inherited it. Five tests across `test_mutability.py`, `test_options.py` and
`test_construction.py` went red on every interpreter, all of them
`assert X.__hash__ is None` on an ordinary mutable struct.

The fix is to count the mixin's `__eq__` as salix's, which is what the issue
said — "an attribute lookup against `_StructMixin.__eq__` **and**
`object.__eq__`" — and which I only half implemented.
`test_a_struct_with_no_co_base_is_untouched` pins it directly rather than
relying on those five to catch it again.

</details>

## Coverage

`tests/test_non_struct_bases.py`, 10 tests. Mutation-checked: reverted to the
struct-base-only rule and rebuilt, **5 of the 10 fail** — the reported shape,
the paired co-base, the inert-base-first ordering, the subclass deferral, and
the `eq=True` case. The other 5 pass either way and are the controls: a plain
Python class, an inert co-base, the reverse base order, `eq=False`, and a
struct with no co-base at all.

<details>
<summary>Shapes that are <i>not</i> changed, asserted so the fix cannot be read as "a non-struct base always wins"</summary>

- **`class B(Base, Mixin)`** — `_StructMixin` is an ancestor of the struct base
  and so precedes an unrelated co-base in the MRO. salix's equality and hash
  both answer, unchanged.
- **`class B(Inert, Base)`** — `object.__eq__` is the absence of one, so the
  search passes over it and the struct base still answers.
- **`class B(Mixin, Base, eq=False)`** — a class that changes the eq option has
  salix's binding written into its own namespace, which the lookup reaches
  before any base. Identity equality, identity hash, consistent.
- **`class B(Mixin, Base, eq=True)`** — not a transition, so nothing is
  rebound and the co-base's `__eq__` still answers. That a keyword matching the
  inherited value does nothing is **#76** and is not this PR's to change; what
  does change is the hash beside it, which no longer contradicts the equality
  that won.

`list` and `tuple` as co-bases are not reachable at all — `multiple bases have
instance lay-out conflict`.

</details>

892 tests pass, 10 skipped, 2 xfailed. `camas check` 25/25 across 3.10–3.15 and
3.14t; `camas flake_check` green, which is the only local thing that
cross-builds Windows.

Closes #47.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, working
> through the agreed list of highest-priority bugs one PR at a time.

### DIFF
--- a/src/meta.c
+++ b/src/meta.c
@@ -22,6 +22,14 @@ struct member_lookup {
 	Py_ssize_t offset;
 };
 
+/* Whether the __eq__ a class resolves came from a class body, and whether the
+ * question could be answered -- asking a base runs whatever its metaclass
+ * defines, so the lookup can raise. */
+struct equality_source {
+	enum { EQUALITY_RESOLVED, EQUALITY_FAILED } tag;
+	bool from_a_body;
+};
+
 static int StructMeta_traverse(PyObject * self, visitproc visit, void * arg);
 static int StructMeta_clear(PyObject * self);
 static void StructMeta_dealloc(PyObject * self);
@@ -36,6 +44,8 @@ static PyObject * build_struct_class(
 );
 static StructType * find_struct_base(PyObject * bases);
 static StructType * find_behaviour_base(PyObject * bases);
+static struct equality_source resolves_body_equality(PyObject * bases);
+static struct equality_source base_equality(PyObject * base);
 static struct options inherited_options(PyObject * bases, StructType const * behaviour);
 static bool any_struct_base_is_mutable(PyObject * bases);
 static bool has_weakref_slot(StructType const * base);
@@ -260,12 +270,20 @@ static PyObject * build_struct_class(
 
 	/* Read before build_class_namespace rebinds anything: afterwards every
 	 * comparison name is present whether the body wrote one or salix did. An
-	 * inherited one survives only if this body leaves equality alone -- and it
-	 * is the behaviour base's, because that is the __eq__ the MRO finds. */
+	 * inherited one survives only if this body leaves equality alone, because a
+	 * class that changes the eq option has salix's binding written into its own
+	 * namespace and that is what the lookup finds first. */
 	bool const body_defines_eq = PyDict_GetItemString(original_namespace, "__eq__") != NULL;
+	struct equality_source const inherited_equality = resolves_body_equality(bases);
+
+	if (inherited_equality.tag == EQUALITY_FAILED) {
+		field_plan_clear(&plan);
+
+		return NULL;
+	}
+
 	bool const inherits_body_eq = (
-		behaviour != NULL &&
-		behaviour->struct_resolves_body_eq &&
+		inherited_equality.from_a_body &&
 		request.options.eq == inherited.eq
 	);
 
@@ -379,6 +397,67 @@ static StructType * find_behaviour_base(PyObject * const bases) {
 	}
 
 	return NULL;
+}
+
+/*
+ * Whether the __eq__ this class will resolve came from a class body, rather
+ * than from the mixin or from object.
+ *
+ * Every base is asked, in the order C3 searches them, and the first whose
+ * branch supplies an __eq__ at all is the answer. A struct base's branch always
+ * supplies one -- the mixin's, or object's where that base opted out of
+ * equality -- so the search stops there and the flag it already stores is the
+ * answer. A non-struct base answers only when it resolves something other than
+ * object's; otherwise it adds nothing to the lookup and the next base decides.
+ *
+ * Asking the struct base alone let a non-struct base ahead of it supply the
+ * equality while salix bound a structural hash beside it, so two instances
+ * compared equal and still took two slots in a set.
+ */
+static struct equality_source resolves_body_equality(PyObject * const bases) {
+	for (Py_ssize_t i = 0; i < PyTuple_GET_SIZE(bases); ++i) {
+		PyObject * const base = PyTuple_GET_ITEM(bases, i);
+
+		if (is_struct_class(base)) {
+			return (struct equality_source){
+				.tag = EQUALITY_RESOLVED,
+				.from_a_body = ((StructType *) base)->struct_resolves_body_eq,
+			};
+		}
+
+		struct equality_source const answer = base_equality(base);
+
+		if (answer.tag == EQUALITY_FAILED || answer.from_a_body) {
+			return answer;
+		}
+	}
+
+	return (struct equality_source){.tag = EQUALITY_RESOLVED, .from_a_body = false};
+}
+
+/*
+ * The two __eq__ that are not a body's: object's, which is the absence of one,
+ * and the mixin's, which is salix's own. A base resolving either contributes
+ * nothing a class body wrote, so the next base is what decides.
+ *
+ * The mixin has to be here rather than assumed unreachable, because it is the
+ * only base Struct itself has and its metaclass is plain `type` -- so the
+ * struct-base arm above does not catch it, and counting its __eq__ as a body's
+ * recorded that against Struct and every struct inheriting from it.
+ */
+static struct equality_source base_equality(PyObject * const base) {
+	PY_OWNED(equality, PyObject_GetAttrString(base, "__eq__"));
+	PY_OWNED(objects, PyObject_GetAttrString((PyObject *) &PyBaseObject_Type, "__eq__"));
+	PY_OWNED(salixs, PyObject_GetAttrString((PyObject *) &StructMixin_Type, "__eq__"));
+
+	if (equality == NULL || objects == NULL || salixs == NULL) {
+		return (struct equality_source){.tag = EQUALITY_FAILED};
+	}
+
+	return (struct equality_source){
+		.tag = EQUALITY_RESOLVED,
+		.from_a_body = equality != objects && equality != salixs,
+	};
 }
 
 /*
--- a/src/meta.c
+++ b/src/meta.c
@@ -58,7 +58,11 @@ static struct equality_source equality_from_the_co_bases(PyObject * bases, Py_ss
 static struct equality_source base_equality(
 	PyObject * base,
 	PyObject * objects_equal,
-	PyObject * mixin_equal,
+	PyObject * mixin_equal
+);
+static struct equality_source supplies_not_equal(
+	PyObject * bases,
+	Py_ssize_t first,
 	PyObject * objects_not_equal,
 	PyObject * mixin_not_equal
 );
@@ -513,20 +517,71 @@ static struct equality_source equality_from_the_co_bases(
 			};
 		}
 
-		struct equality_source const answer = base_equality(
-			base,
-			objects_equal,
-			mixin_equal,
-			objects_not_equal,
-			mixin_not_equal
-		);
+		struct equality_source const answer = base_equality(base, objects_equal, mixin_equal);
 
-		if (answer.tag == EQUALITY_FAILED || answer.from_a_body) {
+		if (answer.tag == EQUALITY_FAILED) {
 			return answer;
+		}
+
+		if (answer.from_a_body) {
+			struct equality_source const paired =
+				supplies_not_equal(bases, first, objects_not_equal, mixin_not_equal);
+
+			return (
+				paired.tag == EQUALITY_FAILED ? paired :
+				(struct equality_source){
+					.tag = EQUALITY_RESOLVED,
+					.from_a_body = true,
+					.needs_derived_not_equal = !paired.needs_derived_not_equal,
+				}
+			);
 		}
 	}
 
 	return (struct equality_source){.tag = EQUALITY_RESOLVED, .from_a_body = false};
+}
+
+/*
+ * Whether any co-base carries a __ne__ of its own, which is then the __ne__
+ * the class resolves and not salix's to replace.
+ *
+ * Asked over the same bases as the equality and separately from it, because
+ * the two can come from different ones: `class B(Equal, WeirdNotEqual, Base)`
+ * takes equality from the first and inequality from the second. Reading the
+ * __ne__ of whichever base supplied __eq__ answered about the wrong class, and
+ * the derived one went in over a pairing that was already there.
+ *
+ * `needs_derived_not_equal` carries "one was found" here; the caller inverts
+ * it, since salix derives one exactly when nobody else supplies it.
+ */
+static struct equality_source supplies_not_equal(
+	PyObject * const bases,
+	Py_ssize_t const first,
+	PyObject * const objects_not_equal,
+	PyObject * const mixin_not_equal
+) {
+	for (Py_ssize_t i = first; i < PyTuple_GET_SIZE(bases); ++i) {
+		PyObject * const base = PyTuple_GET_ITEM(bases, i);
+
+		if (is_struct_class(base)) {
+			break;
+		}
+
+		PY_OWNED(inequality, PyObject_GetAttrString(base, "__ne__"));
+
+		if (inequality == NULL) {
+			return (struct equality_source){.tag = EQUALITY_FAILED};
+		}
+
+		if (inequality != objects_not_equal && inequality != mixin_not_equal) {
+			return (struct equality_source){
+				.tag = EQUALITY_RESOLVED,
+				.needs_derived_not_equal = true,
+			};
+		}
+	}
+
+	return (struct equality_source){.tag = EQUALITY_RESOLVED, .needs_derived_not_equal = false};
 }
 
 /*
@@ -542,45 +597,20 @@ static struct equality_source equality_from_the_co_bases(
 static struct equality_source base_equality(
 	PyObject * const base,
 	PyObject * const objects_equal,
-	PyObject * const mixin_equal,
-	PyObject * const objects_not_equal,
-	PyObject * const mixin_not_equal
+	PyObject * const mixin_equal
 ) {
-	/* Nothing beyond __eq__ is read unless the answer turns on it. A base that
-	 * did not supply the equality has nothing else this decision needs, and
-	 * every attribute read here runs whatever the base put under the name --
-	 * each one being a class that fails to define where it used to build. */
+	/* Only __eq__, and only off a base that might supply it: every attribute
+	 * read here runs whatever the base put under the name, and each one is a
+	 * class that fails to define where it used to build. */
 	PY_OWNED(equality, PyObject_GetAttrString(base, "__eq__"));
 
 	if (equality == NULL) {
 		return (struct equality_source){.tag = EQUALITY_FAILED};
 	}
 
-	if (equality == objects_equal || equality == mixin_equal) {
-		return (struct equality_source){.tag = EQUALITY_RESOLVED, .from_a_body = false};
-	}
-
-	/* Asked only after the one before it was checked: a lookup that raised
-	 * leaves the exception set, and calling back into the API with one pending
-	 * loses it -- which showed up on 3.10 through 3.12 as `_StructMeta returned
-	 * NULL without setting an exception`. */
-	PY_OWNED(inequality, PyObject_GetAttrString(base, "__ne__"));
-
-	if (inequality == NULL) {
-		return (struct equality_source){.tag = EQUALITY_FAILED};
-	}
-
-	/* Neither object's nor the mixin's counts as a pairing the co-base made.
-	 * A co-base may itself derive from the mixin -- it is a permitted base --
-	 * and then its __ne__ resolves to the structural one, which is the thing
-	 * being displaced rather than a pair to leave alone. */
 	return (struct equality_source){
 		.tag = EQUALITY_RESOLVED,
-		.from_a_body = true,
-		.needs_derived_not_equal = (
-			inequality == objects_not_equal ||
-			inequality == mixin_not_equal
-		),
+		.from_a_body = equality != objects_equal && equality != mixin_equal,
 	};
 }
 
--- a/src/meta.c
+++ b/src/meta.c
@@ -28,6 +28,12 @@ struct member_lookup {
 struct equality_source {
 	enum { EQUALITY_RESOLVED, EQUALITY_FAILED } tag;
 	bool from_a_body;
+	/* Whether a non-struct base is what answered. A struct base that resolves a
+	 * body's __eq__ bound object's __ne__ into its own dict when it was built,
+	 * so a subclass inherits the pair and needs nothing; a co-base supplies the
+	 * equality with no __ne__ beside it, and the mixin's structural one is what
+	 * the lookup reaches next. */
+	bool from_a_co_base;
 };
 
 static int StructMeta_traverse(PyObject * self, visitproc visit, void * arg);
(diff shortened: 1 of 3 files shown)


JPHutchins:
Both majors are right. The first is a real gap in this PR; the second is a
false claim in my comment rather than a new defect, and it is the same kind of
overreach the reviewer has caught me in twice today.

## `__ne__` — the half I left behind

Reproduced exactly, in this PR's own headline example:

```
a == b -> True | a != b -> True
```

Deferring the hash to the co-base and leaving `__ne__` to the mixin's
structural one is **#58's shape reached through an inherited `__eq__`** instead
of a body-written one, and `bind_not_equal` exists for precisely that. It now
fires for an equality that came from a co-base as well as one this body wrote.

<details>
<summary>The first version of that fix broke an existing test, and the distinction it forced is the interesting part</summary>

Passing `body_defines_eq || inherits_body_eq` made
`test_a_subclass_of_it_derives_ne_the_same_way` fail: every subclass of a
body-eq struct got a redundant `__ne__` in its own dict, and that test pins the
location deliberately.

It is right to. A **struct** base that resolves a body's `__eq__` bound
`object.__ne__` into its own dict when it was built, so a subclass inherits the
pair and needs nothing. A **co-base** supplies the equality with no `__ne__`
beside it, and the mixin's structural one is what the lookup reaches next.

So the walk now reports *which kind* of base answered, and only a co-base
triggers the bind.

</details>

`a == b` and `a != b` now disagree, for the eq-only co-base and the eq+hash one
both, and there is a test asserting the pair rather than only `==`.

## The walk does not do what I said it does

You are right, and the defect is the sentence, not the code. The comment said
every base is asked "in the order C3 searches them". It ends at the first
struct base — exact for one, an approximation for two, because C3 puts a later
struct base's own `__eq__` ahead of the mixin (the mixin is a shared ancestor,
deferred to the tail). Your trace reproduces:

```
C.__eq__ is BodyEq.__eq__   True
C(1) == C(2)  True | hashes equal  False | set  2
```

**Narrowed rather than chased.** That is the "recorded from one base, answered
by the MRO" shape #76 carries, and attempting the general mechanism inside a
narrow PR is exactly what cost #71 six rounds. #47 is specifically a *non-struct*
base shadowing the struct base's `__eq__`; the two-struct-base version is #76's,
and it is already under a strict xfail in #80's sweep (`BodyEq` in a later
position is in `CONTESTED_HASH`).

The comment now says where the walk is exact and where it is not, and there is
a strict xfail here pinning the limit so the claim cannot drift back.

## The minors

**The fallible probe** is a real behaviour change and it is now pinned rather
than only documented — a co-base with a raising `__eq__` descriptor fails the
class where it used to build. I considered walking `tp_mro`/`tp_dict` instead,
which runs no user code and would remove the regression entirely, and did not:
it needs a `PyType_GetDict` shim for 3.10–3.11 and borrows from another type's
dict under free-threading. Propagating beats demoting — the same class raises
from `==` on first use — and #44's review settled that preference. Say the word
if you would rather have the walk.

**The `__hash__`-only co-base** is deliberate, and I think correct rather than
surprising: `Hashed` supplies no `__eq__`, so the struct base's *structural*
equality is what answers, and a hash that ignores the fields cannot be the
partner of an equality that reads them. Pinned with that reasoning.

## The nits

The "agree" docstring is now true and asserts `!=` — major 1's fix is what
makes it so. The tautological control is a comparison now: `Plain(Equal)` and
`WithAStructBase(Equal, Base)` must have the same `__hash__`, which is the
claim its docstring was making and its body was not. And `object.__eq__` and
the mixin's are fetched once for the walk rather than once per base.

---

926 tests, 3 xfailed. `camas check` 25/25 across 3.10–3.15 and 3.14t;
`camas flake_check` green, which is the only local thing that cross-builds
Windows.

> [!WARNING]
> **LLM Disclosure**
>
> round 1.

JPHutchins:
All three minors are right, and the first is a regression **this PR introduced
last round** — the `__ne__` fix overshot.

## A co-base's own `__ne__` was being taken away

Binding the derived `__ne__` whenever a co-base supplied `__eq__` replaced a
pairing the co-base had already made. Your case:

| | before | after | plain Python |
| --- | --- | --- | --- |
| `W.__ne__ is Weird.__ne__` | False | **True** | True |
| `W(1,0) != W(2,0)` | False | **True** | True |

The flag now says what it is for — `needs_derived_not_equal`, set only where
the co-base supplied `__eq__` and left `__ne__` to the mixin. And you are right
that the old comment's justification was false for an unchanged-eq class; it is
gone.

## `__eq__ = object.__eq__` in a body

Also right, and the fix is smaller than the classification problem it looks
like. **The base's `__hash__` answers it**: Python sets it to `None` for a body
that defines `__eq__` without one, so an explicit `__eq__ = object.__eq__`
declares itself that way even though the value is pointer-identical to
object's.

```
plain  E.__hash__  ->  None
B(E, Base).__hash__ -> None    (was: a working structural hash)
```

That catches every unhashable base with it, which is the right generalisation
rather than a special case: a class deriving from an unhashable one is
unhashable, and leaving the hash alone is what gets that. Binding a structural
one over it was salix making a class hashable that Python says is not.

## The walk ran when its answer was thrown away

Correct, and it also made my pinned test's docstring false in that case — `==`
is object's under `eq=False` and never reaches the descriptor. The walk is now
asked only when `request.options.eq == inherited.eq`, and
`test_a_class_that_throws_that_equality_away_is_not_asked_for_it` pins that
`class B(Hostile, Base, eq=False)` builds.

<details>
<summary>Fixing that surfaced a bug of mine that only 3.10–3.12 could see</summary>

Adding the `__hash__` and `__ne__` lookups, I fetched all three attributes at
once and checked them together. The first raises, and the next two are then
made **with that exception still set**, which loses it — so the class failed
with

```
SystemError: <class 'salix._StructMeta'> returned NULL without setting an exception
```

instead of the descriptor's `RuntimeError`. Green on 3.13, 3.14, 3.15 and
3.14t; red on 3.10, 3.11 and 3.12. Each lookup is checked before the next is
asked for now.

The six-interpreter matrix earned its keep — I would have shipped this
otherwise.

</details>

## The nits

- The sentinels are fetched only once a non-struct base is reached, so a class
  whose first base is a struct — nearly all of them — pays nothing. Done by
  splitting the walk rather than by a lazily-assigned local, since `PY_OWNED`
  is `const` by design.
- "two or more", which is what the MRO trace actually shows.
- The hash-only co-base test asserted things an identity hash would satisfy. It
  now pins the hash as **structural** (`hash(H(1,0)) == hash((1, 0))` and
  `!= hash(H(2,0))`), and the docstring no longer claims a constant hash would
  be an illegal partner — you are right that it would be legal, just not the
  one salix owes a structural equality.

---

926 tests, 3 xfailed. `camas check` 25/25 across 3.10–3.15 and 3.14t;
`camas flake_check` green, run on this branch without switching away from it
mid-build.

> [!WARNING]
> **LLM Disclosure**
>
> round 2.

JPHutchins:
Both minors are right, and both are consequences of the `__hash__` probe I
added last round. **So the probe is reverted rather than patched a third
time**, and the case it was for is filed as #85.

## Why revert rather than fix

Your fix for minor 1 is correct — `(answered || refuses_hashing) && ...` — and
minor 2 would narrow the same way. I did not take either, because of what the
probe's short history says about it:

- it was added for a case **your own round-2 note said has no contract break**;
- I shipped it with **no test**, which is how both of these got through;
- in one round it produced two new defects — `a == b` and `a != b` both False,
  and a widening of exactly the regression that prompted it.

A marginal fix generating a new face every round is the shape that cost #71 six
rounds. Reverting is the narrower answer and it takes both findings with it.

**#85** carries the case, what was tried, why it was backed out, and what a
real fix looks like: classify by *where* `__eq__` is defined — walk the base's
MRO for the first own-dict `__eq__` — rather than by what it resolves to. That
reads no attributes off the base at all, so it would also remove the raising-
descriptor regression the `__eq__` probe still has.

## What is left reads only what the decision turns on

Your minor 2 is fully addressed by that, and then some. A base whose `__eq__`
is object's or the mixin's **has answered** — it adds nothing to the lookup —
and nothing else is asked of it:

```
class RaisingHash:   __hash__ = <descriptor that raises>
class RaisingNe:     __ne__   = <descriptor that raises>

hash(H(1, 0)) == hash((1, 0))    # both build again
hash(N(1, 0)) == hash((1, 0))
```

Only a base that supplies the equality has its `__ne__` read, and only to see
whether it paired the two itself. Pinned as
`test_only_the_equality_is_read_off_a_base_that_did_not_supply_one`.

The `__eq__` probe itself still propagates, still for the reason the round-1
answer gave, and is still the only attribute read.

## The nit

Right, and it is a fair thing to catch me on: last round's commit message
described the sequential-check discipline while the function immediately above
did not follow it. Fixed. They are object's and the mixin's and cannot
realistically raise, so it is consistency rather than a defect — but the
inconsistency was mine and stated in my own words.

---

954 tests, 1 xfailed here. `camas check` 25/25 across 3.10–3.15 and 3.14t.

> [!WARNING]
> **LLM Disclosure**
>
> round 3.

JPHutchins:
**Correction first: my round-3 reply said reverting the `__hash__` probe took
the both-False case with it. That is false and you measured it.**

The `a == b` and `a != b` both `False` shape is caused by the
`equality == objects_equal` short-circuit, which the revert never touched. I
asserted an effect of my own change without running the two lines that would
have shown otherwise — with the extension already built in front of me.

```
class E:  __eq__ = object.__eq__
class B(E, Base): y: object = 0

B(1,0) == B(1,0)   False
B(1,0) != B(1,0)   False
```

It is **#85**, unchanged and correctly still open. My reply overstated what
reverting bought: it removed findings 1 and 2 *of round 3* — the derived-`__ne__`
gap that `refuses_hashing` created, and the `__hash__` read — and left #85's
own symptom where it was.

## Fixed: a co-base derived from the mixin

Right, and it is the more interesting of the two. `_StructMixin` is a permitted
base, so a co-base can derive from it and write its own `__eq__` — and its
`__ne__` then resolves to the **structural** one, which is the answer being
displaced rather than a pairing to leave alone.

```
before:  == True   != True
after:   == True   != False
```

Neither object's `__ne__` nor the mixin's counts as the co-base having paired
them.

## Fixed: the walk ran when the body writes `__eq__`

Also right, and it is the same gate as the `eq` option one round ago — I closed
half of it. The body's `__eq__` sits ahead of every base and `bind_not_equal`
already fires on `body_defines_eq`, so nothing the walk returns is used. Asked
only when neither the body nor the keywords has already settled it.

## Pinned rather than fixed

**The `needs_derived_not_equal` guard had no test** — you are right that
forcing it true left the suite green, which is exactly how I shipped two
untested behaviour changes earlier in this PR. It now asserts `"__ne__" not in
vars(B)` as well as the behaviour, so the *binding* is pinned and not only its
effect.

**A co-base between two struct bases** is #47's own break in a shape that needs
two struct bases to reach. Strict xfail beside the later-struct-base one, and
it belongs to #76 for the same reason: the walk approximates the MRO, and every
place C3 defers the shared mixin past an intermediate base is a place the
approximation differs. I am not widening this PR to chase that — it is the
third distinct shape of it, and #71 is what chasing looks like.

**A raising `__ne__` beside a real `__eq__`** still fails the class. That is the
one name other than `__eq__` the walk reads, and only off a base that answered,
so it is the same trade rather than a new one. Pinned.

**The propagation docstring** rested on a false premise, as you say — a
demoting implementation builds the class with salix's equality rather than
raising. The argument is that demoting is a lie about what was learned, not
that the class is unusable either way. Rewritten to say that.

**The tautological third assert** is gone.

---

954 tests, 2 xfailed here. `camas check` 25/25 across 3.10–3.15 and 3.14t.

> [!WARNING]
> **LLM Disclosure**
>
> round 4 and correcting a false claim in the round-3 reply.

JPHutchins:
One of the three is new and it is a fair hit; the other two are the two already
tracked, and I want to be explicit about the floor that creates.

## The unpinned gate — fixed

Right, and it is the **third fix in this PR I shipped without a test**, which is
the thread running through most of this review's findings. Mutation-confirmed:
removing `!body_defines_eq` leaves every other test green and refuses
`class B(Hostile, Base)` with a body `__eq__` again. Only the new
`test_a_body_that_writes_its_own_equality_is_not_asked_for_a_bases` catches it.

## The two that remain are tracked, not overlooked

**`__eq__ = object.__eq__`** is **#85**, filed in round 3 with the repro, the
two defects the attempted fix produced, and the shape of a real fix (classify
by *where* `__eq__` is defined, walking the base's own dicts, rather than by
what it resolves to). Deferred deliberately, not forgotten.

**A co-base between two struct bases** is a strict xfail in this PR —
`test_a_co_base_between_two_struct_bases_is_not_seen_either` — and belongs to
**#76**. It is the third distinct shape of the same thing: the walk approximates
the MRO, and every place C3 defers the shared mixin past an intermediate base
is a place the approximation differs. The other two shapes are xfailed beside
it.

Both are real, and I am not disputing either. What I am saying is that neither
can be closed inside a PR scoped to #47's report — a non-struct base ahead of
**one** struct base — and that chasing the general mechanism from inside a
narrow PR is exactly what cost #71 six rounds before @JPHutchins ruled to
narrow it.

<details>
<summary>Which means the convergence score has a floor of 2 here, and that is worth naming</summary>

Both are minors, both are correctly re-reported each round, and both are
deliberate deferrals with issues behind them. So the weighted score cannot fall
below 2 while they stand, however many rounds this runs.

The trajectory otherwise: **6 → 3 → 2 → 5 → 3**, and every round's genuinely
new findings have been answered. Rounds 4 and 5 found four real defects between
them and three of the four were "you fixed this and did not pin it" — a
criticism of my discipline rather than of the design, and one I have now
addressed three times over.

</details>

---

959 tests, 2 xfailed here. `camas check` 25/25 across 3.10–3.15 and 3.14t.

> [!WARNING]
> **LLM Disclosure**
>
> round 5.

JPHutchins:
The new minor is right, and it is the **third face of the same mistake in three
rounds** — so this round changes the shape rather than patching it again.

## `__ne__` was answered about the wrong class

`needs_derived_not_equal` read the `__ne__` of whichever base supplied `__eq__`.
That answers about *that base*, not about the class:

```
class B(Equal, WeirdNotEqual, Base)

before:  B.__ne__ is WeirdNotEqual.__ne__   False   != -> False
after:   B.__ne__ is WeirdNotEqual.__ne__   True    != -> 'weird'
plain:   Plain.__ne__ is WeirdNotEqual.__ne__ True  != -> 'weird'
```

It is a **separate question over the same bases** now, asked by the same rule
as the equality: the first co-base carrying a `__ne__` that is neither object's
nor the mixin's supplies the one the class resolves, and salix derives one
exactly when nobody does.

The two earlier faces fall out of that rather than needing their own clauses —
a co-base that paired them itself is found by the same scan, and one derived
from the mixin resolves the *structural* `__ne__`, which is the answer being
displaced rather than a pairing. Both tests still pass, and the new shape has
its own.

Both walks share the approximation the file documents and #76 carries: they
read the bases in order rather than the merged MRO.

## The two nits are the unpinned-fix criticism one level over

An assertion that cannot fail is not coverage, which is the same thing you have
told me three times about fixes without tests. Both are fair:

- **the reverse-order test** asserted three things true of *any* deterministic
  equality and hash. It builds both orders of the same two bases now and
  asserts they differ, which is the claim its name makes;
- **the inert-base test** claimed the search "passes over" the base, which is
  not observable when nothing follows it. It claims the outcome, and points at
  the sibling that does tell stopping from continuing apart.

## The two that remain

Unchanged and unchased: **#85** and the co-base between two struct bases, which
is a strict xfail here and belongs to **#76**. Both correctly re-reported; both
outside what a PR scoped to #47's report can close.

---

960 tests, 2 xfailed here. `camas check` 25/25 across 3.10–3.15 and 3.14t.

> [!WARNING]
> **LLM Disclosure**
>
> round 6.

JPHutchins:
**Finding 4 is the one that mattered, and it names the root the other six
rounds were circling.** The classification fetched attributes and
pointer-compared against cached method-wrappers. Both halves were wrong, and
this round replaces the mechanism rather than patching it again.

## Why the fetch could never stop producing regressions

Your new finding 1 is the **fourth** instance of the same thing, and the list
makes the pattern plain:

| round | the read that was added | the class it started refusing |
| --- | --- | --- |
| 1 | `__eq__` off every co-base | a raising `__eq__` descriptor |
| 3 | `__hash__` and `__ne__` | a raising `__hash__` on an eq-less base |
| 6 | `__ne__` off every co-base | a raising `__ne__` on an eq-less base |

Each fix was correct about *which base answers* and wrong about *what it costs
to find out*. It was never going to stop while the mechanism was a fetch.

And you are right that the pointer comparison rests on CPython handing back the
ancestor's cached wrapper unchanged. That is an implementation detail, and it
is also why the fetch could never tell `__eq__ = object.__eq__` in a body from
not defining one at all.

## Asking the class dicts answers the question that was always the right one

Did a body write this name. `object`'s and the mixin's dicts are skipped,
because those are the two answers that mean nobody did. Nothing runs.

Verified against the built extension:

```
raising __eq__     class builds
raising __ne__     class builds
raising __hash__   class builds
```

The three tests pinning those refusals are gone, replaced by one asserting the
opposite over all three names. There is no attribute-read surface left to
regress.

**#85 closes with it** — `__eq__ = object.__eq__` is a definition now, so the
class is unhashable exactly as plain Python makes it and `!=` derives properly:

```
Plain.__hash__   None      B.__hash__   None
B(1,0) == B(1,0)  False    B(1,0) != B(1,0)  True
```

Both earlier `__ne__` faces still hold, by the same walk rather than by their
own clauses.

<details>
<summary>The shim, and what did not change</summary>

`PyType_GetDict` is 3.12+, so `types.h` carries a two-line shim returning a
strong reference either way. `PyDict_Contains` borrows nothing, which is what
makes this safe under free-threading.

**Findings 2 and 3 are unchanged and still #76's**: the walk reads the bases in
order rather than the merged MRO, so neither a later struct base's `__eq__` nor
a co-base sitting between two struct bases is seen. Two strict xfails. That is
a different problem from the one this round solves — it is about *which order*
to ask in, not about *how* to ask — and it needs the class's real MRO, which
does not exist until after `type.__new__`.

</details>

## The nit

Right again, and I fixed the wrong half last time: `class B(Base, Equal)` never
reaches the co-base walk, so no mutation of it can touch that test. It is
folded into the reverse-order pair, which is the only part that ever
discriminated.

---

962 tests, 2 xfailed here. `camas check` 25/25 across 3.10–3.15 and 3.14t;
`camas flake_check` green.

> [!WARNING]
> **LLM Disclosure**
>
> round 7.

Issue #66, opened 2026-08-05
A mixin subclass over a hashable co-base is equal to a plain value it does not hash like


JPHutchins:
Split out of #48's round-9 review. A subclass of `_StructMixin` over a hashable value type compares equal to a plain value of that type and hashes differently from it, so the two are equal and cannot find each other in a dict or a set.

```python
from salix import Struct

MIXIN = Struct.__mro__[1]
Imp = type("Imp", (MIXIN, tuple), {"__slots__": ()})
imp, plain = Imp((1, 2)), (1, 2)
```
```
>>> imp == plain, plain == imp
(True, True)
>>> hash(imp), hash(plain)
(8796084174068, -3550055125485641917)
>>> {plain: "from the tuple"}.get(imp, "MISSED")
'MISSED'
>>> imp in {plain}
False
```

## Where it comes from

#48 gave every mixin slot an `is_struct` guard, and the two halves chose different fallbacks:

| slot | what a non-struct gets |
| --- | --- |
| `tp_hash` | `object`'s — identity |
| `tp_richcompare` | `NotImplemented` |

`NotImplemented` is the right answer for a comparison salix cannot make, but Python then tries the reflected operation, and `tuple.__eq__(plain, imp)` answers `True` because `Imp` really is a tuple. So equality comes from the co-base and the hash comes from `object`, and the invariant that equal objects hash alike is broken between them.

<details>
<summary><b>Why this is #48's residue rather than #48's regression</b></summary>

Before #48 there was no guard at all: `Struct_hash` read `struct_type_of(self)->struct_field_names` off a type that has no such storage, which is a read past the end of the allocation. The PR turned a memory error into a contract violation, which is the right direction; this is what is left.

The fallback comment in `src/mixin.c` rules out the obvious fix on purpose:

> object's, specifically, and not the co-base's: an impostor over `list` is given object's repr and hashes by pointer, where a plain list subclass is unhashable. Deferring to whatever the co-base defines would mean walking the MRO for something to borrow, on a path reachable only by subclassing a private class.

</details>

<details>
<summary><b>Options</b></summary>

1. **Make a non-struct unhashable.** `tp_hash` raises `TypeError` instead of falling back. An object that cannot be hashed cannot be a dict key, so the invariant cannot be broken — and "unhashable" is an ordinary thing for a Python object to be. Costs the identity hash the current tests pin, and it is a second policy for the same "not a struct" case.
[...]

Issue #57, opened 2026-08-01
The ClassVar/InitVar text path is a heuristic: the escapes and false refusals it is known to have


JPHutchins:
Split out of #50's review. The object path is exact; the text path — which runs only when the annotation reaches salix as source (`from __future__ import annotations`, and see the 3.14 note below) — is a substring match with identifier boundaries, and this is the list of what that costs. Filed so the list lives somewhere a reader can find, instead of accumulating in PR comments.

## False accepts — a class-var-shaped field silently swallows a positional (#14's symptom)

| shape | why it escapes |
| --- | --- |
| `ClassVar as CV` → `CV[int]` | the text says `CV`; nothing connects it to `typing.ClassVar` |
| a factory-local alias on 3.14 | `def make(): CV = typing.ClassVar; class Q(Struct): x: Annotated[CV[int], 'm']` — FORWARDREF resolves against the module's `__globals__` only, so the annotation comes back as its source text and the object path never sees the real form. Refused on 3.13, accepted on 3.14, same source. |
| `Optional[Annotated[ClassVar[int], 'm']]` on the **object** path | `special_form_of` walks `__origin__`, never `__args__`, so a form kept in the arguments is not reached. The *text* path refuses the same annotation, so the two disagree. |

## False refusals — legal code refused, loudly

| shape | why it matches |
| --- | --- |
| a user type actually named `ClassVar` or `InitVar` | the matcher only knows the spelling |
| `Annotated[int, ClassVar]` | the form is metadata, not the type; the text path cannot tell position from spelling. The object path accepts it (`__origin__ is int`), so again the two paths disagree. |

## The shape of a real fix

The two-path split is the cause: an exact check on objects and a spelling check on text will always disagree at the edges. Options, roughly in order of cost:

1. **Walk `__args__` as well as `__origin__`** on the object path. Closes the `Optional[Annotated[...]]` accept; also makes the object path refuse `Annotated[int, ClassVar]`, i.e. it buys agreement by adopting the text path's false refusal.
2. **Parse rather than match** — even a tiny expression tokeniser would know metadata position from type position and would fix both `Annotated` rows.
3. **Resolve the text against the module's namespace** before matching, which recovers the renamed import and the factory alias, at the cost of evaluating annotations salix currently does not evaluate.

[...]

Issue #51, opened 2026-08-01
A default that is only shallowly immutable is still shared: hashability is the test the type list approximates


JPHutchins:
#19 made empty `list`/`dict`/`set`/`bytearray` defaults copy per instance, refused non-empty ones, and shared everything else. That rule tests **shallow** immutability, and the property that actually matters is transitive.

```python
class Inner(Struct):
    v: object

class Holder(Struct):
    pair:   tuple  = ([],)        # shared, and the inner list is mutable
    nested: Inner  = Inner([])    # shared, and Inner's list is mutable
```

`tuple` is shared because tuples are immutable, and a frozen struct is shared because it cannot be rebound — but neither is true of what they *contain*. `a.pair[0].append(1)` is visible through `b.pair[0]`, which is #19's own bug reached one level down. The same applies to `deque()`, `defaultdict()` and any `list` subclass, all of which fall outside the exact-type list and are shared.

<details>
<summary><b>Hashability is exactly the property, and it is free</b></summary>

`hash` on a container hashes its elements, and salix's own structural hash does the same for a struct, so unhashability propagates upward precisely as mutability does:

```
(1, 2)                     hashable   -> transitively immutable, safe to share
([],) tuple w/ list        UNHASHABLE -> something in it can be mutated
Inner(1) frozen struct     hashable   -> safe to share
Inner([]) struct w/ list   UNHASHABLE -> something in it can be mutated
"text"                     hashable   -> safe to share
[] empty list              UNHASHABLE -> something in it can be mutated
frozenset()                hashable   -> safe to share
```

The rule becomes one question instead of a list of four types:

| default | today | with hashability |
|---|---|---|
| `(1, 2)`, `"x"`, `Inner(1)` | share | share |
| empty `list`/`dict`/`set`/`bytearray` | copy | copy |
| non-empty `[1]` | refuse | refuse |
| **`([],)`, `Inner([])`, `deque()`, `MyList()`** | **share — leaks** | **refuse** |

Cost is one `hash()` per defaulted field at class creation, never per construction.

This is what #19 originally proposed (`__hash__ is None`), reached there for the wrong reason — as a way to detect a mutable default at all — and dropped when the implementation switched to matching msgspec.

</details>

<details>
<summary><b>What it would cost, and why it is not obviously right</b></summary>

[...]

PR #86, opened 2026-08-08
style: comments that carry the reason and not the issue number

JPHutchins:
Fourteen GitHub issue references were left behind in `src/` comments by already-merged PRs. CLAUDE.md puts that kind of metadata in the commit message: a reference in a `.c` file points at something that moves — closed, retitled, superseded — and rots there with no one watching. Each reference is deleted and the reason it stood for is kept.

**Comments only. No code is touched**, so there is nothing here that can change behaviour. `check` 25/25 and `flake_check` green.

Four comments were also **false**, and this is where they are paid off rather than in another review round on a PR that had already converged:

| file | said | now |
| --- | --- | --- |
| `meta.c` | `equality_source`'s docstring, stranded above `struct definition`, still described the attribute fetch #83 removed | says which body `from_a_body` means — always a base's, never this class's own |
| `meta.c` | the walk's gate blamed "runs whatever the co-base put under the name" for why it is only asked when usable | the reason is the cost of reading the class dicts along every co-base's MRO |
| `meta.c` | "the two **sentinels** are fetched once for the walk" | the two **names** are interned once |
| `annotations.c` | the `ClassVar` check that reads annotation *values* is "not in this tree yet, so the reason to keep FORWARDREF is a pending one" | #50 merged and `special_form_of` reads them, so the reason is a present one |

<details>
<summary><b>Every reference removed, and what stands in its place</b></summary>

Reasons are preserved verbatim wherever the reference was the only thing deleted.

| site | reference | what changed |
| --- | --- | --- |
| `types.h:132` | `#7 made the write side safe by going through PyMember_SetOne` | "The write side is safe because it goes through PyMember_SetOne" |
| `types.h:139` | `...survived every time. #46.` | trailing reference dropped; the measurement above it is untouched |
| `construct.c:274` | `...that would replace this list; #51 is where that is argued.` | "...that would replace this list." |
| `annotations.c:67` | `#50 adds the #14 ClassVar check` | rewritten — see the table above |
| `fields.c:338` | `A tree since #57's ruling and not a chain` | "A tree and not a chain" |
| `fields.c:417` | `-- #14 again, on the path almost every class takes. Both are walked now, per #57's ruling.` | "on the path almost every class takes. Both are walked now." |
| `fields.c:423` | `#57 owns the disagreement; when this comment claimed...` | "The disagreement is left open here; when this comment claimed..." |
| `fields.c:581` | `#57 keeps the list.` | sentence dropped; the list of escapes above it stays |
| `fields.c:641` | `...a silently unguarded class, which is #14 again with no way to notice.` | "...where a `ClassVar` becomes a field again with no way to notice." |
| `fields.c:677` | `...nothing would ever hand out, which is what #56 was;` | "...nothing would ever hand out;" |
| `fields.c:758` | `That route is out of contract and #51 is where the whole type list is argued.` | "That route is out of contract." |
| `mixin.c:76` | `Hash is the exception, and #66 is why:` | "Hash is the exception:" |
| `mixin.c:88` | `...and #66's fix does not change it.` | "...and refusing to hash does not change it." |
| `hash.c:27` | `...for an object to be. #66.` | trailing reference dropped |

Lines are rewrapped wherever a deletion left a short one, plus one paragraph in `mixin.c` that #81 left wrapped after three words (`they raise. An` / `AttributeError rather than...`).

</details>

<details>
<summary><b>Why the four false ones are in this PR and not their own</b></summary>

The three `meta.c` ones are #83's round-8 nits. That review had come back at a **convergence score of 1** — converged — with nothing but those comments left. Pushing a fix would have started a round 9 and the repo's rule is that every PR ends with a landed review on its current head, so the cheapest correct place for them was the next src-comment-only PR. This is it.

The `annotations.c` one is different in kind: nobody wrote it wrong. It was true when it was written and #50 merging made it false, which is the failure mode a cross-reference has and a stated reason does not. It also could not be *fixed* without deleting a reference, since the sentence's subject was `#50`, so it belongs here regardless.

Verified before rewriting it rather than assumed: `append_declared` holds the annotation value and hands it to `special_form_of`, which reads it — so the check does read the values, and `Format.STRING` would hand every class a string and put it on the text matcher's heuristics instead of only the modules that deferred their annotations.

</details>

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. She asked me to proceed to the next phase of work and drive its PRs to merge; this is the queued comment cleanup that her instruction *"code comments should never point at metadata such as GH issues, refer to my CLAUDE.md"* called for, carried out as its own no-behaviour PR so it cannot hide inside a fix.

### DIFF
--- a/src/annotations.c
+++ b/src/annotations.c
@@ -64,9 +64,9 @@ static PyObject * borrow_annotate(PyObject * const namespace) {
  * the callable and re-runs every annotation in the dict, so a class holding one
  * forward reference evaluates its other annotations again. Plain 3.14 evaluates
  * once and defers. Format.STRING would avoid it, because only the keys are read
- * here -- but not for long: #50 adds the #14 ClassVar check, which reads the
- * values, and STRING hands it strings. Not in this tree yet, so the reason to
- * keep FORWARDREF is a pending one rather than a present one.
+ * here -- but the ClassVar and InitVar check reads the values, and STRING hands
+ * it strings, which would put every class on the text matcher's heuristics
+ * instead of only the ones whose module deferred its annotations.
  *
  * FORWARDREF is the fallback: it leaves an unresolved bare forward reference as
  * a ForwardRef instead of raising, which is all we need since only the field
--- a/src/construct.c
+++ b/src/construct.c
@@ -271,7 +271,7 @@ static enum result fill_defaults(
  * whitelists the same four types.
  *
  * Every one of them is unhashable, which is the single test that would replace
- * this list; #51 is where that is argued. msgspec shares them as well.
+ * this list. msgspec shares them as well.
  *
  * dataclasses refuses the shape outright and needs default_factory to express
  * it at all. This copies, so the common spelling means what it looks like it
--- a/src/fields.c
+++ b/src/fields.c
@@ -335,9 +335,8 @@ static enum inheritance inherits_field(StructType const * const base, PyObject *
  *
  * The annotation is walked rather than probed once, because
  * Annotated[ClassVar[int], ...] reaches ClassVar two hops down and is otherwise
- * the same bug wearing a wrapper. A tree since #57's ruling and not a chain:
- * __origin__, every element of __args__, and a PEP 695 alias's __value__ are
- * all edges.
+ * the same bug wearing a wrapper. A tree and not a chain: __origin__, every
+ * element of __args__, and a PEP 695 alias's __value__ are all edges.
  */
 enum : int {
 	/* A budget on work rather than on depth, because bounding the shape stopped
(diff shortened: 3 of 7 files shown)


JPHutchins:
Both nits are the same observation and it is correct: rewording the gate replaced a false mechanism with a true cost and **dropped a consequence that is still live**. Fixed at `840fe07`.

The old clause *"can refuse a class that would otherwise build"* was true before #83 for the reason #83 removed, and it is still true for a different one. Traced by hand rather than taken from the review:

| step | site |
| --- | --- |
| `struct_type_dict` returns NULL, or `PyDict_Contains` errors | `meta.c:584-594` |
| → `DEFINITION_UNREADABLE` | `meta.c:587`, `:593` |
| → `EQUALITY_FAILED` | `meta.c:499-501`, `:514-516` |
| → `build_struct_class` returns NULL | `meta.c:310-314` |

So a walk run for an answer nobody uses can refuse the class, and that is a reason for the gate that cost alone does not give. The gate now says both.

**Deleting a true clause while fixing a false one is the opposite of what this PR is for**, which is why it is worth a round on a review that had already converged at 0.

<details>
<summary><b>The <code>equality_source</code> nit is answered by that clause rather than by a docstring — declined, with the reason</b></summary>

The recommendation was to add a sentence to `equality_source` covering the tag: *"the walk answers RESOLVED or FAILED, and FAILED is produced when a base's class dict cannot be read, which refuses the class."*

The **consequence** half of that is now in the gate, where a reader can act on it. The **naming** half is what this repo cannot add. CLAUDE.md:

> Docstrings SHALL NEVER restate what the code already clearly states. This is LLM slop, and an unmaintainable violation of SSOT.

A sentence explaining that a tag named `EQUALITY_FAILED` means the question could not be answered is that restatement — the enum name is the documentation, and a second copy is a thing that can drift from it.

Two things make this a smaller gap than it looks:

- `struct definition` directly above still carries the readability half (*"whether the dicts could be read at all"*), and it is the struct that actually produces the failure. `equality_source` only propagates it.
- `base_defines`' own docstring explains why the read is a dict walk at all.

The asymmetry the finding notices — that `struct definition` documents its failure state and `equality_source` does not — is real, and the honest reading is that `struct definition`'s one-liner is itself close to restating its own field names. Removing it is out of scope for this PR; adding a matching one to `equality_source` would double the slop rather than halve it.

</details>

`check` 25/25 and `flake_check` green on `840fe07`.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked me to drive this phase's PRs to merge. It answers round 1 of the automated review on this PR: one nit fixed in the source, one declined on CLAUDE.md's no-restating rule with the reasoning above.

PR #87, opened 2026-08-08
fix: a method named after a field is refused, not turned into its default

### LINKED ISSUE #54: A method or property named after a field becomes that field's default
`drop_class_variables` strips every class-body binding whose name is a field name, so a method that collides with a field disappears without a word and the slot descriptor answers instead:

```python
class Collide(Struct):
    x: int

    def x(self):
        return "method"

Collide(1).x        # 1 — the field, not the method
```

Same for a `property`:

```python
class Collide(Struct):
    y: int

    @property
    def y(self):
        return "property"

Collide(1).y        # 1
```

**Parity, again.** Every alternative does the same thing:

```
salix                       Collide(1).x = 1
dataclass(frozen, slots)    D(1).x = 1
dataclass (plain)           P(1).x = 1
NamedTuple                  N(1).x = 1
```

So this is not salix being unusual, and it is not a correctness bug — the annotation declared `x` a field, and a field it is. Both shapes are pinned in `tests/test_methods.py` so a future change to `drop_class_variables` cannot alter them unnoticed.

<details>
<summary><b>Why it might still be worth refusing</b></summary>

The deletion is *silent*, and it is the same silence #8 was filed for: a class body says one thing and the object does another. #8's case was a class variable shadowing an inherited field, and the fix was to strip it; this is the same strip, hitting something the user plainly meant to keep.

Refusing would be one `PyCallable_Check` (or a check for a descriptor) in `drop_class_variables` before the delete, and a message naming the field. It would make salix stricter than all three alternatives — which is a real argument against, since a user porting from a dataclass would meet an error where they previously met silence.

Against that: nobody writes a method named after a field on purpose, so the error would only ever fire on a mistake.

Related: #51 and #14 are the other two places where "reject or stay level with the alternatives" is the open question. It is worth answering them the same way rather than case by case.

</details>

Not urgent, and it is the *option* being recorded rather than a defect.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked for gaps found while testing methods on structs to be filed rather than fixed.

JPHutchins:
Closes #54.

A `def`, `property`, `classmethod` or `staticmethod` named after a field is now a `TypeError` at class creation instead of being dropped in silence.

```
TypeError: 'x' is a field, and the class body binds a function to that name
which salix would drop for the descriptor that reads the field; rename one of them
```

Dropping it was silent **twice over**, which is the correction you made to this issue's own premise:

| | before | after |
| --- | --- | --- |
| `Collide.x` | the slot descriptor — the method is gone | refused |
| `Collide._struct_defaults_` | `(<function Collide.x>,)` | refused |
| `Collide()` | **built**, `x` holding a function | refused |

So a required field silently turned optional, and anything reading `_struct_defaults_` saw a function where an `int`'s default belonged.

**This makes salix stricter than every alternative** — a plain dataclass, `dataclass(slots=True)` and `NamedTuple` are all silent here. That is the divergence to weigh, and it is the same answer #14 and #51 got: refuse the shape nobody means rather than stay level with it. Nobody writes a method named after a field on purpose, so the error only ever fires on a mistake.

## The rule is the four spellings Python's own syntax produces, not a descriptor test

A descriptor test is shorter and **is not portable**. Measured on this machine, `hasattr(type(v), "__get__")`:

| value | 3.10 | 3.13 | 3.14 | 3.15 |
| `functools.partial(f, 1)` | **False** | True | True | True |
| a bound method | True | True | True | True |

`functools.partial` gained `__get__` in 3.13, so `handler: object = partial(f, 1)` would **build on 3.10–3.12 and be refused from 3.13 on** — class creation that depends on the interpreter. A bound method is a descriptor on every version, and `handler: object = source.handle` is a default someone means.

So: the three decorators answer by type, subclasses included, because none is ever a value anyone means. A **function** is asked for its name, because a function *is* an ordinary value and only a `def` written in this body carries the field's own name.

<details>
<summary><b>What that keeps working, all six asserted in <code>TestDefaultsThatAreCallable</code></b></summary>

```python
class WithHandler(Struct):
    handler: object = module_level_handler   # __name__ != "handler"

class WithLambda(Struct):
    handler: object = identity

class WithPartial(Struct):
    handler: object = functools.partial(module_level_handler, 1)

class WithBound(Struct):
    handler: object = Source().handle

class WithType(Struct):
    kind: object = int

class WithBuiltin(Struct):
    emit: object = print
```

The cost of naming the four rather than testing for a descriptor is `functools.cached_property`: it is its own type rather than a `property` subclass, so it is still dropped as it was before. It has never worked on a struct anyway — it needs a `__dict__` to cache into.

</details>

<details>
<summary><b>Two mutations, because a test that passes on the fix has proved nothing</b></summary>

Each arm was a separate build with `touch src/*.c` in front of it, and the two `.so` files hash differently (`a87f4cf6…` and `fec2c4f4…`).

**1. The predicate off** (`return 0;` at the top) — the six refusal tests fail and every control still passes:

```
FAILED TestNameCollisions::test_a_def_named_after_a_field_is_refused
FAILED TestNameCollisions::test_a_property_named_after_a_field_is_refused
FAILED TestNameCollisions::test_a_classmethod_named_after_a_field_is_refused
FAILED TestNameCollisions::test_a_staticmethod_named_after_a_field_is_refused
FAILED TestNameCollisions::test_the_field_may_be_one_the_base_declared
FAILED TestNameCollisions::test_the_message_names_the_remedy
6 failed, 38 passed
```

**2. The over-broad fix** (`return Py_TYPE(bound)->tp_descr_get != NULL;`) — the negative controls are what catch it, which is the reason they exist:

```
FAILED TestDefaultsThatAreCallable::test_a_module_level_function_defaults_a_field
FAILED TestDefaultsThatAreCallable::test_a_lambda_defaults_a_field
FAILED TestDefaultsThatAreCallable::test_a_partial_defaults_a_field
FAILED TestDefaultsThatAreCallable::test_a_bound_method_defaults_a_field
4 failed, 40 passed
```

`WithType` and `WithBuiltin` correctly survive arm 2 — neither is a descriptor on any version.

</details>

<details>
<summary><b>Where it lives, and why the inherited case is refused too</b></summary>

In `drop_class_variables`, which is the one site that sees **every** field name being taken out of the namespace — this class's own and the ones a base declared. Putting it in `append_declared` instead would have refused a re-annotated field and stayed silent on

```python
class Child(Base):
    def x(self) -> str: ...      # x is Base's field, no annotation here
```

which is the shape the issue was originally filed about. "Refused if you re-annotate, silent if you don't" is not a rule anyone could hold in their head, so it is one rule at one site. `test_the_field_may_be_one_the_base_declared` pins it.

Nothing in the predicate fetches an attribute, so nothing runs the class author's code: `func_name` is a struct field and the rest are type checks. That is deliberate — [#83](https://github.com/JPHutchins/salix/pull/83) spent four of its eight rounds on refusals caused by *reading* a name off a base.

`dict_value_ref` moves from `src/fields.c` to `src/types.h` unchanged, because the check needs the value and that is already the one free-threading-correct dict read in the tree (`PyDict_GetItemRef` on 3.13+, which is every version that can have the race).

</details>

`check` 25/25 across 3.10–3.15 and 3.14t, and `flake_check` green — which matters more than usual here, since the predicate takes the address of `PyProperty_Type`, `PyClassMethod_Type` and `PyStaticMethod_Type`, and on Windows those are `__declspec(dllimport)`. A *static initializer* of such addresses is what broke every `wheel-3xx-windows-x86_64` derivation on 2026-08-05; a runtime comparison is fine, and the cross-build is the proof rather than the assumption.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked me to proceed to the next phase and drive its PRs to merge. I measured the three candidate discriminators for this collision and asked her which answer #54 should get; she chose refusing it, and this is that.

### DIFF
--- a/src/fields.c
+++ b/src/fields.c
@@ -72,7 +72,6 @@ static struct special_form special_form_of(
 	struct form_probes const * probes
 );
 static struct special_form form_within(PyObject * annotation, struct form_probes const * probes);
-static PyObject * dict_value_ref(PyObject * mapping, PyObject * key);
 static struct special_form named_special_form(PyObject * text, struct form_probes const * probes);
 static bool names_form(PyObject * text, PyObject * needle);
 static bool continues_identifier(Py_UCS4 character);
@@ -496,28 +495,6 @@ static struct special_form form_within(
 	}
 
 	return (struct special_form){0};
-}
-
-/*
- * A strong reference to the value, or NULL if the key is gone.
- *
- * PyDict_GetItem hands back a borrowed one, and taking a reference to it is two
- * steps: on a free-threaded build another thread can remove the key and drop
- * the last reference in between. PyDict_GetItemRef takes it under the dict's
- * own lock instead -- 3.13+, which is also every version that can have the
- * race, so the older branch is the plain read it always was.
- *
- * It also reports a failed lookup rather than swallowing it, which is why the
- * caller separates "gone" from "could not tell".
- */
-static PyObject * dict_value_ref(PyObject * const mapping, PyObject * const key) {
-#if PY_VERSION_HEX >= 0x030D0000
-	PyObject * value = NULL;
-
-	return PyDict_GetItemRef(mapping, key, &value) < 0 ? NULL : value;
-#else
-	return Py_XNewRef(PyDict_GetItem(mapping, key));
-#endif
 }
 
 /*
--- a/src/meta.c
+++ b/src/meta.c
@@ -99,7 +99,13 @@ static enum result bind_hash(
 	bool inherits_body_eq
 );
 static enum result drop_class_variables(PyObject * namespace, PyObject * all_names);
-static int defines_a_method(PyObject * bound, PyObject * field_name);
+static enum result refuse_colliding_methods(
+	PyObject * original_namespace,
+	PyObject * all_names,
+	PyObject * class_name
+);
+static int defines_a_method(PyObject * bound, PyObject * field_name, PyObject * class_name);
+static int defined_in_this_body(PyObject * qualname, PyObject * field_name, PyObject * class_name);
 static StructType * create_class(
 	PyTypeObject * metatype,
 	PyObject * name,
@@ -285,6 +291,12 @@ static PyObject * build_struct_class(
 	struct field_plan plan = field_plan_build(base, original_namespace);
 
 	if (field_plan_failed(&plan)) {
+		return NULL;
+	}
+
+	if (refuse_colliding_methods(original_namespace, plan.all_names, name) != RESULT_OK) {
+		field_plan_clear(&plan);
+
 		return NULL;
 	}
 
@@ -910,20 +922,30 @@ static enum result bind_hash(
 	return rebind(namespace, hash_name, options.eq);
 }
 
-/*
- * Any class-body binding of a field name -- an annotated default, or a bare
+/* Any class-body binding of a field name -- an annotated default, or a bare
  * assignment over a name the base already declared -- would sit in this class's
- * dict ahead of the __slots__ descriptor that reads the value.
- *
- * A method is refused rather than dropped. Dropping it was silent twice over:
- * the class lost a definition its body plainly meant to keep, and the value
- * became that field's default, so a required field turned optional and
- * _struct_defaults_ reported a function where an int's default belonged.
- */
+ * dict ahead of the __slots__ descriptor that reads the value. */
 static enum result drop_class_variables(PyObject * const namespace, PyObject * const all_names) {
 	for (Py_ssize_t i = 0; i < PyList_GET_SIZE(all_names); ++i) {
 		PyObject * const field_name = PyList_GET_ITEM(all_names, i);
-		PY_OWNED(bound, dict_value_ref(namespace, field_name));
+		int const present = PyDict_Contains(namespace, field_name);
+
+		if (present < 0 || (present == 1 && PyDict_DelItem(namespace, field_name) < 0)) {
+			return RESULT_ERROR;
+		}
+	}
+
+	return RESULT_OK;
+}
+
+static enum result refuse_colliding_methods(
+	PyObject * const original_namespace,
+	PyObject * const all_names,
+	PyObject * const class_name
+) {
+	for (Py_ssize_t i = 0; i < PyList_GET_SIZE(all_names); ++i) {
+		PyObject * const field_name = PyList_GET_ITEM(all_names, i);
+		PY_OWNED(bound, dict_value_ref(original_namespace, field_name));
 
 		if (bound == NULL) {
 			if (PyErr_Occurred()) {
@@ -933,7 +955,7 @@ static enum result drop_class_variables(PyObject * const namespace, PyObject * c
 			continue;
 		}
 
-		int const method = defines_a_method(bound, field_name);
+		int const method = defines_a_method(bound, field_name, class_name);
 
 		if (method < 0) {
 			return RESULT_ERROR;
@@ -951,37 +973,16 @@ static enum result drop_class_variables(PyObject * const namespace, PyObject * c
 
 			return RESULT_ERROR;
 		}
-
-		if (PyDict_DelItem(namespace, field_name) < 0) {
-			return RESULT_ERROR;
-		}
 	}
 
 	return RESULT_OK;
 }
 
-/*
- * Whether the class body wrote a method under this field's name, rather than a
- * value for the field to default to. 1 yes, 0 no, -1 with an exception set.
- *
- * The four spellings Python's own syntax produces, and no more. A function is
- * asked for its name because a function is also an ordinary value --
- * `handler: object = default_handler` defaults a field to a module-level
- * function and has to keep working -- and only a `def` written in this body
- * carries the field's own name. The three decorators are never a value anyone
- * means, so the type alone answers, subclasses included.
- *
- * A descriptor test would be shorter and is not portable: functools.partial
- * gained __get__ in 3.13, so `handler: object = partial(f, 1)` would build on
- * 3.10 through 3.12 and be refused from there on. A bound method is a descriptor
- * too. Both are defaults someone means. The cost of naming the four is
- * functools.cached_property, which is its own type rather than a property
- * subclass and so is dropped as it was before.
- *
- * No attribute is fetched, so nothing here runs the class author's code:
- * func_name is a struct field and the rest are type checks.
- */
-static int defines_a_method(PyObject * const bound, PyObject * const field_name) {
+static int defines_a_method(
+	PyObject * const bound,
+	PyObject * const field_name,
+	PyObject * const class_name
+) {
 	if (
 		PyObject_TypeCheck(bound, &PyProperty_Type) ||
 		PyObject_TypeCheck(bound, &PyClassMethod_Type) ||
@@ -994,7 +995,43 @@ static int defines_a_method(PyObject * const bound, PyObject * const field_name)
 		return 0;
 	}
 
-	return PyObject_RichCompareBool(((PyFunctionObject *) bound)->func_name, field_name, Py_EQ);
+	return defined_in_this_body(
+		((PyFunctionObject *) bound)->func_qualname,
+		field_name,
+		class_name
+	);
+}
+
+static int defined_in_this_body(
+	PyObject * const qualname,
+	PyObject * const field_name,
+	PyObject * const class_name
+) {
+	if (!PyUnicode_Check(qualname)) {
+		return 0;
+	}
+
+	PY_OWNED(own, PyUnicode_FromFormat("%U.%U", class_name, field_name));
+
+	if (own == NULL) {
+		return -1;
+	}
+
+	int const written_here = PyObject_RichCompareBool(qualname, own, Py_EQ);
+
+	if (written_here != 0) {
+		return written_here;
+	}
+
+	PY_OWNED(nested, PyUnicode_FromFormat(".%U", own));
+
+	if (nested == NULL) {
+		return -1;
+	}
+
+	Py_ssize_t const matched = PyUnicode_Tailmatch(qualname, nested, 0, PY_SSIZE_T_MAX, 1);
+
+	return matched < 0 ? -1 : matched != 0;
 }
 
 static StructType * create_class(
--- a/src/meta.c
+++ b/src/meta.c
@@ -99,6 +99,7 @@ static enum result bind_hash(
 	bool inherits_body_eq
 );
 static enum result drop_class_variables(PyObject * namespace, PyObject * all_names);
+static int defines_a_method(PyObject * bound, PyObject * field_name);
 static StructType * create_class(
 	PyTypeObject * metatype,
 	PyObject * name,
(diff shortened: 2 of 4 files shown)


JPHutchins:
Round 1's first minor is correct and it is a regression this PR introduced. Fixed at `5a323cd`.

I had reasoned about `handler: object = handler` and dismissed it as rare. It is not rare, and the review's second example is worse than the first — **re-defaulting an inherited field**. Both built before this PR and were refused by it.

`func_name` cannot separate them, because the class dict records `handler -> function` identically either way. **`func_qualname` can**: a `def` in a class body is `<class>.<name>`, a module-level function is just `<name>`. The rule is now that the qualname's last two components are this class's name and this field's name — equal to `C.handler`, or ending in `.C.handler` for a class nested in a function or another class.

Measured on the built extension:

| body | qualname | before | now |
| --- | --- | --- | --- |
| `def handler(self)` | `C.handler` | refused | refused |
| `handler: object = handler` (module-level) | `handler` | **refused** | allowed |
| `class Child(Base): handler: object = handler` | `handler` | **refused** | allowed |
| `handler: object = Source.handler` | `Source.handler` | **refused** | allowed |
| `handler: object = lambda v: v` | `C.<lambda>` | allowed | allowed |

The third row is the one the finding named that I had not thought of at all, and the fourth fell out of the same fix.

The class name is a parameter now, so the check moved out of `drop_class_variables` into `build_struct_class` beside the other body-contract refusals. That has a second benefit: it reads `original_namespace`, so the delegating-metaclass re-entry cannot reach it — that namespace has had its field bindings dropped already.

**Mutation:** restoring the `func_name` comparison fails exactly the three new tests and **no refusal test**, so the fix is narrowed rather than weakened.

## The second minor is real and is not fixed here — filed as #89

`@functools.cached_property` and `@functools.cache` under a field's name are still dropped and still become the default. I verified both; `_struct_defaults_` holds the descriptor and `Cached()` builds.

Widening to "any descriptor" is what this PR's own reasoning rejected, and the numbers are why:

| value | 3.10 | 3.13 | 3.14 | 3.15 |
| `functools.partial(f, 1)` has `__get__` | **False** | True | True | True |
| a bound method has `__get__` | True | True | True | True |

That rule builds `handler: object = partial(f, 1)` on 3.10–3.12 and refuses it after — class creation following the interpreter — and refuses `handler: object = source.handle` on every version. Both are asserted working in `TestDefaultsThatAreCallable`.

The `__qualname__` rule *would* generalise, since `functools.wraps` copies it. Two things stop it being a drop-in, and they are why this wants a decision rather than more code:

- `functools.cached_property` has **no** `__qualname__` — its `__set_name__` never runs, because salix drops the binding before `type.__new__`. Only its `.func` has one.
- Reading `__qualname__` off an arbitrary object is an **attribute fetch**. #83 spent four of its eight rounds on refusals caused by fetching, and the predicate here deliberately fetches nothing.

So I took the finding's own fallback: **both spellings are now pinned** by `TestCollisionsStillNotRefused`, asserting what they do today so the silent defaulting cannot regress unnoticed, and the gap is #89 with the measurement on it.

<details>
<summary><b>Both nits fixed</b></summary>

**`test_a_lambda_defaults_a_field` never used a lambda** — correct, it defined `def identity` and tested the same path as the test above it. It uses a real lambda now, whose qualname ends in `<lambda>` and so can never match a field name however the class is spelled.

**The inherited-field collision was only pinned for `def`** — a `@property` colliding with an inherited field is now pinned too, which exercises the type-check branch against `all_names` rather than against this class's own annotations.

</details>

One deliberate limit left standing, stated rather than hidden: `y: object = staticmethod(f)` is refused, because a `staticmethod` object as a *stored value* is indistinguishable from the decorator spelling and is not a thing anyone writes on purpose. If you would rather that were allowed it is the same `__qualname__` hop through `__func__`, and it belongs with #89.

`check` 25/25 and `flake_check` green on `5a323cd`.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked me to drive this phase's PRs to merge. This answers round 1 of the automated review: the false positive fixed in the source, the incompleteness pinned in tests and filed as #89, and both nits fixed.

JPHutchins:
Round 2 answered at `f97ce66`. One source change, three tests.

## The NULL guard is applied and its stated mechanism does not reproduce

> `PyFunction_New` → `PyFunction_NewWithQualname(code, globals, NULL)` leaves `func_qualname == NULL` (the `__qualname__` getter lazily substitutes `func_name`)

Measured, reading the struct field at the offset `meta.c` reads rather than through the getter, after calling `PyFunction_New` via `ctypes.pythonapi`:

```
=== 3.10 ===                                   === 3.14 ===
  +32  name      ptr=0x7ffff7a46230 'sample'     +32  name      ptr=0x7ffff7a71980 'sample'
  +40  qualname  ptr=0x7ffff7a46230 'sample'     +40  qualname  ptr=0x7ffff7a71980 'sample'
```

`func_qualname` is **not NULL** on either — it is the *same pointer* as `func_name`, so CPython substitutes at construction, not lazily in the getter.

So the guard is applied as **hardening, not as a fix for a reachable crash**: I cannot reach a NULL `func_qualname` from Python or through the C API on the floor or the newest, the guard was written to be defensive and silently was not, and `qualname == NULL ||` costs nothing. If there is a construction path I have missed, it is now safe either way.

## The other minor is #89, which the finding agrees is shippable as tracked

> This is a documented, tracked limitation rather than an unaddressed defect

Agreed, and both spellings stay pinned by `TestCollisionsStillNotRefused` so it cannot regress unnoticed.

## Three nits, all real, all about tests that could not see what they claimed

**The direct-equality branch had no coverage at all** — this is the best finding of the round. Every struct in this file is nested inside a test method, so every body method's qualname carries a `<locals>.` prefix and only the tailmatch branch ever fired; the branch that fires for the *common* case, a module-level struct, was untested. Now built at module scope through `exec`, giving `Collide.x` exactly.

**Mutation, because a new test that passes has proved nothing:** deleting the direct-equality branch fails that test and **only** that test — 1 failed, 52 passed. So it is the only thing reaching it, which is exactly what the nit said.

The other two are accepted limits, and I have pinned both rather than argue them:

| shape | why it is refused |
| --- | --- |
| `y: object = staticmethod(f)` | the three decorator types answer by type, and a `staticmethod` object written as a value *is* the same object a decorator makes |
| `handler: object = Colliding.handler`, where that other class is also called `Colliding` | a qualname carries no way to say whose nesting chain it came from |

You were right that my "none is ever a value anyone means" does not hold for the first. It is now a test called `test_a_wrapper_object_used_as_a_default_is_refused` under `TestRefusalsWiderThanTheDefect`, so the over-refusal is visible and allowing it later has to update a test deliberately. Both would be fixed by the same `__qualname__` hop through `__func__`, which belongs with #89.

`check` 25/25 and `flake_check` green on `f97ce66`.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked me to drive this phase's PRs to merge. This answers round 2: the NULL guard applied with a measurement that contradicts its stated mechanism, the untested branch pinned and mutation-checked, and the two over-refusals pinned as accepted limits.

JPHutchins:
Round 3 found a real corruption that this PR's own rule created. Fixed at `e115aea`.

## Name mangling — the pattern could never match

Measured, which is the whole finding in four lines:

```python
class Probe:
    __x: int
    def __x(self): return "method"

vars(Probe)          # key is '_Probe__x'
Probe.__annotations__ # key is '_Probe__x'
fn.__name__          # '__x'
fn.__qualname__      # 'Probe.__x'     <- not mangled
```

The compiler mangles the *storage* and leaves the qualname alone. So the pattern built from the field name as stored — `Mangled._Mangled__x` — could never match `Mangled.__x`, and:

```python
class Mangled(Struct):
    __x: int
    def __x(self): return "method"

Mangled._struct_defaults_   # (<function Mangled.__x>,)   -- BUILT
```

The exact corruption #54 exists to stop, reachable through the private spelling only, and introduced by the round-1 fix. The field name is unmangled before the pattern is built now.

**Why unmangle the field rather than read `func_name`**, which was the finding's first suggestion: `handler: object = lambda v: v` has qualname `C.<lambda>` and `func_name` `<lambda>`, so a pattern built from `func_name` matches and refuses the lambda. A pattern built from the (unmangled) field name does not. `unmangled` applies the language reference's rule in reverse, including the corner where a class name that is all underscores mangles nothing.

**Mutation:** removing the unmangling fails `test_a_private_name_is_refused_although_the_compiler_mangles_it` and nothing else. Its companion — a private field defaulting to a same-named module function — passes either way and is coverage, not a discriminator; a module function's qualname is bare, so unmangling cannot change that answer.

## Two nits folded in

**`PyObject_RichCompareBool` dispatches to a str subclass's `__eq__`** — correct, and it makes my "nothing here runs the class author's code" claim false, since `__qualname__` can be set to a `str` subclass. It is one `PyUnicode_Tailmatch` plus a dot-boundary read now: one allocation instead of two, no dispatch, and the equality branch is gone. That also answers the "two allocations where one suffix match would do" nit.

**The wrapper test claimed all three decorator types and pinned only `staticmethod`** — parametrised over the three.

The `Elsewhere` docstring named a class that does not exist; it names what the test actually builds.

## Two left as they are, with the reason

| nit | why |
| --- | --- |
| the refusal runs after `field_plan_build`, so a default-ordering error pre-empts it | it needs the field names, which is what `field_plan_build` produces. Moving it means computing them twice |
| it re-reads namespace values `append_declared` already saw | a second dict lookup per field, on a path that creates a class. Threading the values out of the plan is a wider change than the defect |

Both are cost or ordering rather than correctness, and neither changes what any class builds.

## The remaining minor is #89

Re-reported, and the finding again says *"Acceptable to ship as tracked (#89)"*. Both spellings stay pinned by `TestCollisionsStillNotRefused`. Nothing has changed there since round 2 — the portability measurement is on the issue.

`check` 25/25 and `flake_check` green on `e115aea`.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked me to drive this phase's PRs to merge. This answers round 3: the mangling corruption fixed and mutation-checked, two nits folded in, two declined with reasons, and #89 unchanged.

JPHutchins:
Round 4 answered at `770379e`, which also merges `main`. The first minor is a false negative that round 3's fix introduced.

## `_C__x` is both a mangled name and a name someone can write

```python
class C(Struct):
    _C__x: int
    def _C__x(self): return "method"

C._struct_defaults_   # (<function C._C__x>,)   -- BUILT
```

`_C__x` is what the compiler stores for `def __x` in class `C`, **and** it is a perfectly ordinary identifier the compiler never touches. Both bind the same key, and the stored name cannot say which happened — so rebuilding the pattern from the unmangled spelling alone made this one match nothing. The corruption #54 exists to stop, reintroduced through the other door.

**Either pattern counts now**: the stored spelling first, then the unmangled one if it differs. A false positive is not reachable that way — a module-level function used as a default has a bare qualname and matches neither.

**Mutation:** removing the stored-spelling attempt fails six tests, the new one among them; removing the *unmangled* attempt was round 3's mutation and fails the private one. Each half is load-bearing.

## The message nit falls out of the same fix

Whichever pattern matched is the spelling the source used, so it is handed back and quoted:

| body | message says |
| --- | --- |
| `def x` | `'x' is a field` |
| `def __x` | `'__x' is a field` |
| `def _C__x` | `'_C__x' is a field` |

Previously the middle row said `'_Mangled__x'`, which never appears in the source — so "rename one of them" named something the author could not find. Correct finding.

## This commit also merges `main`

#88 landed and the two PRs each added a refusal at the same line of `build_struct_class`. Both run now, methods before slots.

Worth recording how the resolution went, because it is a trap rather than a mistake: **the autofix hook reformatted the conflict markers themselves into unparseable C** before I had resolved them, so the region was repaired by hand rather than by taking either side. The 25/25 below is the semantic check on that merge — #88's `__slots__` tests and this PR's method tests pass together, which a clean textual merge would not have guaranteed.

## Unchanged

**#89** — re-reported a third time, and a third time called *"acceptable to ship as tracked"*. Both spellings stay pinned.

**The ordering nit** — declined a third time, same reason: the refusal needs the field names that `field_plan_build` produces, so moving it means computing them twice. It changes which message a class with two independent errors gets, and nothing about which classes build.

`check` 25/25 and `flake_check` green on `770379e`.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked me to drive this phase's PRs to merge. This answers round 4 and folds in the rebase that #88 merging made necessary.

Issue #55, opened 2026-08-01
A metaclass `__new__` that hands back a struct class drops the options and defaults of the call that delegated to it


JPHutchins:
Split out of #48's review round 3. The delegation half of that finding is fixed there; this is the half that remains, and it is narrower than it looks.

## What happens

`create_class` now builds as the winning metatype, so the ordinary handoff never re-enters `StructMeta_new`. What still can is a `StructMeta` subclass with a **Python** `__new__`: `type.__new__` hands off to it, it builds the class (typically via `super().__new__`), and the outer `StructMeta_new` gets back a class whose field table is already installed. It then skips its own install, so the options and defaults it planned are never applied.

```python
class Forwarding(salix.StructMeta):
    def __new__(mcls, *args, **keywords):
        return super().__new__(mcls, *args, **keywords)

class Base(Struct, metaclass=Forwarding):
    x: int

B = salix.StructMeta("B", (Base,), {"__annotations__": {"y": int}}, order=True)
B(1, 2) < B(1, 3)          # TypeError: '<' not supported

D = salix.StructMeta("D", (Base,), {"__annotations__": {"y": int}, "y": 42})
D(1)                       # TypeError: D() missing required argument 'y'
```

## Why the skip is there

The returned class need not be the one this call describes. A `__new__` can hand back any struct class it likes, including one built earlier from unrelated bases — measured, and reachable today:

```python
class Wrong(salix.StructMeta):
    def __new__(mcls, *args, **keywords):
        return super().__new__(mcls, *args, **keywords) if first_call else Existing
```

Installing over that writes `Existing`'s `struct_slot_offsets` from a plan built for a different layout: a field table pointed at memory the instances do not have. The skip is what stops it, and dropping it is not an option.

## What it would take

The install needs to distinguish "the class my delegate just built for me" from "a class my delegate found lying around". Field-name equality is not enough — same names can still mean different offsets when the base chain differs — so it would have to check the base relationship too, or the delegated build would have to be given the keywords and the untransformed namespace it never receives.

## Scope

Reachable only by **both**: a `StructMeta` subclass with a Python `__new__`, **and** an explicit `salix.StructMeta(...)` call over a base of that metatype. Every ordinary route is unaffected — a class statement resolves the metaclass to the subclass itself, so there is no outer call to lose anything:

```python
class Sub(Base):            # metaclass resolves to Forwarding, single build
    y: int = 42
Sub(1).y                    # 42

class Ordered(Base, order=True):
[...]

PR #88, opened 2026-08-08
fix: a body __slots__ entry that is not a field is refused, not dropped

JPHutchins:
Closes the `__slots__` half of #12. The `__new__` half stays open — it is blocked on #13 by measurement.

A body `__slots__` entry that is not a field is now refused instead of thrown away:

```
TypeError: __slots__ names 'extra', which is not a field of this class; a struct's
fields are its slots, so salix would drop it -- declare it as a field, or drop __slots__
```

## The issue asked for a flat refusal and a flat refusal is not implementable

The issue proposed:

```
TypeError: a struct class may not declare __slots__; its fields are its slots
```

I built exactly that first. It **refuses salix's own `__slots__`**, and the suite says so — mutation 2 below. The path:

1. `META("Built", (Base,), namespace)` where `Base`'s metatype is a `_StructMeta` subclass with a Python `__new__`;
2. `build_class_namespace` writes `__slots__` into its copy of the namespace;
3. `create_class` calls `PyType_Type.tp_new(metatype, ...)`, and `type_new` sees `winner != metatype` with a non-`type_new` `tp_new`, so it **hands off** to that metaclass's `__new__`;
4. which calls `super().__new__(...)` — landing back in `StructMeta_new` with the namespace from step 2, `__slots__` and all.

That hand-off is deliberate: it is how a user's metaclass gets to see the class, and `reject_unless_planned` is what guards the result. So the refusal has to tolerate re-entry.

**The rule is therefore: refuse every entry that would be lost.** An entry that names a field, or `__weakref__`, is one salix writes itself — replacing it changes nothing, so there is nothing to refuse. Anything else is discarded today, and that is the whole of the defect.

| `__slots__` | fields | outcome |
| --- | --- | --- |
| `("extra",)` | `x` | **refused** — `extra` would be lost |
| `"extra"` | `x` | **refused** — a bare string is one name to `type.__new__`, so it is read as one here |
| `("x",)` | `x` | accepted; salix writes `("x",)` over it, so nothing is lost |
| `("__weakref__", "x")` | `x`, `weakref=True` | accepted; salix writes both |
| salix's own, on re-entry | — | accepted, which is the point |

It also fixes the weakref corner a value-comparison rule would have had: the re-entrant call arrives with **no keywords** (that is #55), so it recomputes `weakref` from the inherited options and would disagree with the tuple salix wrote one frame up. Membership does not care.

<details>
<summary><b>Two mutations, each a separate build</b></summary>

Different `.so` hashes (`eff8bd6e…`, `3fa329f5…`), `touch src/*.c` in front of each.

**1. The refusal disabled** — only the two refusal tests fail, so nothing else in the suite is standing in for them:

```
FAILED TestBindingsSalixOwns::test_a_body_slot_that_is_not_a_field_is_refused
FAILED TestBindingsSalixOwns::test_a_bare_string_slots_is_refused_too
2 failed, 963 passed
```

**2. The flat refusal the issue proposed** — `present` means refuse:

```
FAILED TestBindingsSalixOwns::test_a_body_slots_naming_only_fields_is_accepted
FAILED TestBindingsSalixOwns::test_weakref_may_be_named_because_salix_writes_it
FAILED TestAMetaclassSubclass::test_a_delegate_that_writes_its_own_new_is_refused_not_mis_planned
```

That third one is the demonstration. Its own message is *"did not plan"*; under the flat rule it becomes *"a struct class may not declare `__slots__`"*, on a namespace no class body wrote.

</details>

<details>
<summary><b>Why it runs after the field plan, and what it does not cover</b></summary>

It needs the field names to know what would be lost, so it runs straight after `field_plan_build` and clears the plan on refusal. A body with both a `ClassVar` and a bad `__slots__` reports the `ClassVar`, since that refusal comes first.

`__slots__` is read from `original_namespace`, so a `NULL` return with an exception set is separated from a plain absence rather than swallowed — otherwise the next C call would run with a pending exception, which is [#83](https://github.com/JPHutchins/salix/pull/83)'s failure mode.

Not covered, and unchanged: a `__slots__` naming a field *twice*, which `type.__new__` would reject on its own; and `__dict__` in `__slots__`, which is refused by this rule since it is not a field — a struct with a `__dict__` is not a shape salix offers.

</details>

`Struct` itself is unaffected — `create_struct_base` builds it from a namespace holding only `__module__`, checked before writing this.

`check` 25/25 across 3.10–3.15 and 3.14t, `flake_check` green.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked me to proceed to the next phase and drive its PRs to merge. She ruled this half of #12 "reject" on 2026-08-01; the note I left saying a naive refusal cannot work turned out to be still true, so this is the version that can.

### DIFF
--- a/src/meta.c
+++ b/src/meta.c
@@ -79,7 +79,11 @@ static PyObject * build_class_namespace(
 	bool inherits_body_eq,
 	bool derive_not_equal
 );
-static enum result refuse_displaced_slots(PyObject * original_namespace, PyObject * new_names);
+static enum result refuse_displaced_slots(
+	PyObject * original_namespace,
+	PyObject * all_names,
+	bool carries_a_weakref_slot
+);
 static PyObject * build_slots(PyObject * new_names, bool weakref);
 static enum result set_match_args(PyObject * namespace, PyObject * all_names, bool wanted);
 static enum result apply_options(
@@ -288,7 +292,14 @@ static PyObject * build_struct_class(
 		return NULL;
 	}
 
-	if (refuse_displaced_slots(original_namespace, plan.new_names) != RESULT_OK) {
+	if (
+		refuse_displaced_slots(
+			original_namespace,
+			plan.all_names,
+			request.options.weakref || has_weakref_slot(base)
+		) !=
+		RESULT_OK
+	) {
 		field_plan_clear(&plan);
 
 		return NULL;
@@ -714,7 +725,8 @@ static PyObject * build_class_namespace(
 
 static enum result refuse_displaced_slots(
 	PyObject * const original_namespace,
-	PyObject * const new_names
+	PyObject * const all_names,
+	bool const carries_a_weakref_slot
 ) {
 	PyObject * const declared = PyDict_GetItemString(original_namespace, "__slots__");
 
@@ -733,17 +745,38 @@ static enum result refuse_displaced_slots(
 
 	for (Py_ssize_t i = 0; i < PyTuple_GET_SIZE(entries); ++i) {
 		PyObject * const entry = PyTuple_GET_ITEM(entries, i);
-		int const named_by_a_field = PySequence_Contains(new_names, entry);
+
+		if (!PyUnicode_Check(entry)) {
+			PyErr_Format(
+				PyExc_TypeError,
+				"__slots__ items must be strings, not '%.200s'",
+				Py_TYPE(entry)->tp_name
+			);
+
+			return RESULT_ERROR;
+		}
+
+		if (PyUnicode_CompareWithASCIIString(entry, "__weakref__") == 0) {
+			if (carries_a_weakref_slot) {
+				continue;
+			}
+
+			PyErr_SetString(
+				PyExc_TypeError,
+				"__slots__ names __weakref__, which salix writes only for a class "
+				"that asks for it; pass weakref=True instead"
+			);
+
+			return RESULT_ERROR;
+		}
+
+		int const named_by_a_field = PySequence_Contains(all_names, entry);
 
 		if (named_by_a_field < 0) {
 			return RESULT_ERROR;
 		}
 
 		if (named_by_a_field == 1) {
-			continue;
-		}
-
-		if (PyUnicode_Check(entry) && PyUnicode_CompareWithASCIIString(entry, "__weakref__") == 0) {
 			continue;
 		}
 
--- a/src/meta.c
+++ b/src/meta.c
@@ -66,6 +66,7 @@ static struct definition any_base_defines(PyObject * bases, Py_ssize_t first, Py
 static struct options inherited_options(PyObject * bases, StructType const * behaviour);
 static bool any_struct_base_is_mutable(PyObject * bases);
 static bool has_weakref_slot(StructType const * base);
+static bool any_base_has_weakref_slot(PyObject * bases);
 static struct options base_options(StructType const * base);
 static PyObject * build_class_namespace(
 	PyObject * original_namespace,
@@ -296,7 +297,7 @@ static PyObject * build_struct_class(
 		refuse_displaced_slots(
 			original_namespace,
 			plan.all_names,
-			request.options.weakref || has_weakref_slot(base)
+			request.options.weakref || any_base_has_weakref_slot(bases)
 		) !=
 		RESULT_OK
 	) {
@@ -681,6 +682,18 @@ static bool has_weakref_slot(StructType const * const base) {
 	return base != NULL && base->heap_type.ht_type.tp_weaklistoffset != 0;
 }
 
+static bool any_base_has_weakref_slot(PyObject * const bases) {
+	for (Py_ssize_t i = 0; i < PyTuple_GET_SIZE(bases); ++i) {
+		PyObject * const base = PyTuple_GET_ITEM(bases, i);
+
+		if (PyType_Check(base) && ((PyTypeObject *) base)->tp_weaklistoffset != 0) {
+			return true;
+		}
+	}
+
+	return false;
+}
+
 /* The namespace handed to type.__new__: a copy of the original with every
  * class-body binding of a field name removed (so none of them clashes with the
  * __slots__ descriptor that reads the value) plus __slots__ / __match_args__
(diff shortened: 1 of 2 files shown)


JPHutchins:
Both majors are real. Fixed at `d7b0183`.

## Major 1 — the `__weakref__` exemption was unconditional, and shouldn't be

Correct, and it is this PR's own defect one door along. `build_slots` writes `__weakref__` only when `options.weakref && !has_weakref_slot(base)`, so:

```python
class W(Struct):
    x: int
    __slots__ = ("__weakref__", "x")

W.__slots__          # ('x',)  -- the entry is gone
weakref.ref(W(1))    # TypeError: cannot create weak reference
```

An entry dropped in silence, which is exactly what this PR exists to refuse, and the message it added *promises* refusal for it. `__weakref__` is now exempt only when the class will carry the slot (`options.weakref || has_weakref_slot(base)`), and refused otherwise — with the remedy that actually works:

```
TypeError: __slots__ names __weakref__, which salix writes only for a class
that asks for it; pass weakref=True instead
```

The finding worried that threading the decision would reopen the corner the unconditional exemption was protecting. It does not: the re-entrant call arrives with **no keywords**, so a `weakref=True` request is already lost on that path and `reject_unless_planned` refuses the class regardless. The existing delegate test passes `order=True`, whose namespace carries no `__weakref__` at all — and it still passes.

## Major 2 — the mechanism is right, the conclusion is not, and the fix is right anyway

`weakref.ref(Weak(1))()` does answer `None`, for exactly the reason given. But the claim that the test "always fails" and that *"the PR's claim of check 25/25 cannot hold"* is not what happens. Measured both ways on `378406f`:

```console
$ pytest tests/test_methods.py -k weakref
1 passed

$ pytest tests/test_methods.py -k weakref --assert=plain
FAILED  tests/test_methods.py::TestBindingsSalixOwns::test_weakref_may_be_named_because_salix_writes_it
```

**pytest's assertion rewriting binds the intermediate to a local**, which keeps the temporary alive for the rest of the expression. So the suite was genuinely green and the 25/25 was a real measurement — but the test was green for a reason that has nothing to do with weakref, which is the finding's actual substance and is worth more than the headline. It holds the instance now and asserts identity, and the whole file passes **with rewriting off as well as on**, so nothing else in it was passing by accident either.

## Both minors, both correct

| finding | fix |
| --- | --- |
| membership was against `new_names`, so a subclass naming an **inherited** field was refused although that slot is already on the base | `all_names`, and the message's *"not a field of this class"* stops being untrue for that case |
| `PySequence_Contains` is value equality, so a non-str entry whose `__eq__` matched a field name passed and was then dropped, and one whose `__eq__` raised replaced the refusal | entries must be strings first, phrased as CPython phrases it: `__slots__ items must be strings, not 'int'` |

<details>
<summary><b>The nit is declined, with the reason</b></summary>

> The `__weakref__` exemption uses `PyUnicode_CompareWithASCIIString`, which can return -1 with an exception set (an allocation failure, e.g. MemoryError, on a non-ASCII string).

It cannot. `PyUnicode_CompareWithASCIIString` asserts its argument is a `str` and then compares the code points against the ASCII bytes directly — it allocates nothing and has no error return, which is why its signature has no way to distinguish -1-as-error from -1-as-less-than. A non-ASCII left-hand side simply compares unequal.

The string check added for minor 2 also means it now only ever runs on a `str`, so the precondition it asserts is established one branch earlier.

</details>

Four tests added: naming `__weakref__` without asking for it, an inherited field's slot accepted, a non-str entry refused, and the fields-only `__slots__` put back into the `match_args=False` test — whose `__slots__` assertion the previous commit had left vacuous, which was the other nit.

`check` 25/25 and `flake_check` green on `d7b0183`.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked me to drive this phase's PRs to merge. This answers round 1 of the automated review on this PR: both majors and both minors fixed, one nit fixed, one declined on the C API's contract.

JPHutchins:
Round 2 answered at `bc35a00`. The first minor is a real refusal of a class that works.

## `has_weakref_slot(base)` asked the layout base; the slot comes from any base

Reproduced exactly as described:

```python
class WR(Struct, weakref=True): pass
class Big(Struct):
    a: int
    b: int

class Both(Big, WR):
    __slots__ = ("__weakref__",)     # TypeError

class Plain(Big, WR): pass           # builds, and is weak-referenceable
```

So the entry was **not** one that would be lost — which is this PR's entire rule — and the refusal contradicted it. `find_struct_base` picks `Big` by field count, `Big` has no weakref slot, and `WR`'s never got asked. `any_base_has_weakref_slot` walks the bases now.

**Mutation:** putting the exemption back on the layout base fails `test_the_base_with_the_slot_need_not_be_the_layout_base` and nothing else.

This is the same *"recorded from one base, answered by another"* shape #78 carries, which is why it wanted the any-base spelling in the first place.

## The two-shapes nit, answered rather than fixed

`build_slots` still asks the layout base, and that is deliberate: it decides whether to **write** the slot, not whether the class **carries** one. Those are different questions, and moving the write side is #78's to do rather than this PR's.

Checked before leaving it, since "they can drift" deserves a measurement rather than a shrug:

```python
class Both(Big, WR, weakref=True): pass
# builds, __slots__ == ('__weakref__',)
```

No duplicate-slot refusal from CPython, because it discounts `__weakref__` when picking the solid base — so the two conditions disagreeing is not currently reachable as a bug.

## The re-entrant call — no functional change, and the message was the defect

Your own verification against `d7b0183^` is the important part and I agree with it: that call was refused before this PR too, by `reject_unless_planned`. The keyword loss on re-entry is #55 and is not something this PR can fix.

What *was* wrong is what you named: the message told a caller who had just passed `weakref=True` to pass `weakref=True`. It now states the condition instead of prescribing a remedy that may already have been followed:

```
TypeError: __slots__ names __weakref__ and this class carries no weakref slot
to name; a struct gets one from weakref=True or from a base that has one
```

True in the ordinary case and true in the re-entrant one, where the class genuinely has no weakref slot because the keyword was dropped a frame up.

## The missing coverage

Correct — every weakref test used the keyword, so the `has_weakref_slot` half of the exemption had none. Two tests now: a subclass of a `weakref=True` base naming `__weakref__` without the keyword, and the base-order case above. To be straight about which is which: the first passes with or without this commit's fix, because there the weakref base *is* the layout base — it pins the branch you asked for, and only the second discriminates.

`check` 25/25 and `flake_check` green on `bc35a00`.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked me to drive this phase's PRs to merge. This answers round 2: the layout-base bug fixed and mutation-checked, the misleading message replaced, the missing branch covered, and the drift nit answered with a measurement.

PR #91, opened 2026-08-08
fix: refuse a default that claims a hash and then refuses to give one

### LINKED ISSUE #51: A default that is only shallowly immutable is still shared: hashability is the test the type list approximates

JPHutchins:
Closes #51.

`xs: object = ([],)` handed every instance the **same inner list**. A tuple is not one of the four types the copy rule covers, so it was shared outright — the aliasing bug that rule exists to stop, one level down and outside its reach.

This is the narrow version you ruled on 2026-08-08: keep the four types exactly as they are, and additionally refuse only the container-of-mutables case.

## The discriminator, re-measured against this `main` before writing anything

The type claims a hash and the instance then declines:

| default | type claims | `hash()` | before | now |
| `([],)` | yes | `TypeError` | shared | **refused** |
| `((1, []),)` | yes | `TypeError` | shared | **refused** |
| `memoryview(bytearray())` | yes | **`ValueError`** | shared | **refused** |
| `(1, 2)`, `'text'`, `frozenset()`, `memoryview(b'')` | yes | ok | shared | shared |
| `deque()`, `defaultdict(list)`, `array('i')` | **no** | raises | shared | shared |
| `[]` | no | raises | copied | copied |

The last two rows are why the **type** is asked first. A deque, an array or a mutable struct sets `__hash__ = None` — it says it does not hash before being asked — and is left shared, which is where msgspec leaves it and what the ruling keeps.

**A writable memoryview raises `ValueError`, not `TypeError`.** Catching only the second lets it through and turns class creation into a `ValueError` from nowhere. Anything that is neither — a `MemoryError`, an interrupt — is not the instance declining, and propagates unchanged.

<details>
<summary><b>Two mutations, each its own build</b></summary>

**Remove the type-claims guard** — 27 failures, the shared-and-not-refused controls among them:

```
FAILED test_a_value_may_be_a_default[dict] [set] [bytearray] [array] [Mutable] ...
FAILED test_refcounts.py::test_a_copied_default_hands_out_no_reference_to_itself
27 failed, 953 passed
```

**Drop `ValueError` from the catch** — three failures, and the raw error escapes into class creation, which is the whole reason the trap is worth a test of its own:

```
E   ValueError: cannot hash writable memoryview object
3 failed, 977 passed
```

</details>

<details>
<summary><b>Cost: two prebuilt <code>.so</code> files swapped between interleaved runs</b></summary>

30 defaulted tuple fields, µs per class:

```
before  13.81   13.63   13.51
after   13.78   14.38   13.97
```

The ranges overlap, so this is at most a few percent and three pairs cannot resolve it further — stated as measured rather than as "no cost". A class declaring **no** defaults pays nothing (5.4 either way). A tuple caches its hash, so the second call on the stored copy is free.

</details>

<details>
<summary><b>What moved in the tests, and one thing I got wrong first</b></summary>

`tests/values.py` gains `REFUSED_AS_DEFAULT`, **selected out of `EVERY`** rather than built fresh so callers can test identity, with the membership written out and asserted:

```python
assert [type(value).__name__ for value in REFUSED_AS_DEFAULT] == ["memoryview"]
```

Getting that wrong was the first thing that happened: the tuple held a *second* memoryview with the same contents, the identity test never matched the one in `EVERY`, and the drift assertion fired at import. It was the assertion doing its job on its first run.

Two existing tests move rather than break. A writable memoryview was listed as shared-and-not-refused and is now refused, so it leaves that parametrize and gets its own. The surrounding docstring said *"every one is unhashable, which is the test #51 argues should replace the list"* — true of the rule as it was, not of the one you ruled. What those values have in common now is that their **types** declare `__hash__ = None`.

</details>

`check` 25/25 across 3.10–3.15 and 3.14t, `flake_check` green.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked me to proceed to the next phase and drive its PRs to merge. This implements the narrower of the two options she approved for #51.

### DIFF
--- a/src/fields.c
+++ b/src/fields.c
@@ -685,7 +685,7 @@ static enum result reject_unsafe_default(PyObject * const field_name, PyObject *
 	Py_ssize_t const filled = struct_copies_default(kind) ? PyObject_Size(value) : 0;
 
 	if (filled <= 0) {
-		return filled < 0 ? RESULT_ERROR : refuse_shared_mutable_contents(field_name, value);
+		return filled < 0 ? RESULT_ERROR : RESULT_OK;
 	}
 
 	PyErr_Format(
@@ -719,13 +719,14 @@ static enum result refuse_shared_mutable_contents(
 		return RESULT_OK;
 	}
 
-	if (PyObject_Hash(value) != -1) {
+	if (PyObject_Hash(value) != -1 || !PyErr_Occurred()) {
 		return RESULT_OK;
 	}
 
 	/* A writable memoryview answers ValueError where a tuple of lists answers
-	 * TypeError, and anything else -- a MemoryError, an interrupt -- is not the
-	 * instance declining and is left to propagate. */
+	 * TypeError, and anything else -- a MemoryError, an interrupt, whatever a
+	 * class author's own __hash__ raises -- is not the instance declining and
+	 * propagates with its own message rather than being read as this. */
 	if (!PyErr_ExceptionMatches(PyExc_TypeError) && !PyErr_ExceptionMatches(PyExc_ValueError)) {
 		return RESULT_ERROR;
 	}
@@ -734,9 +735,10 @@ static enum result refuse_shared_mutable_contents(
 
 	PyErr_Format(
 		PyExc_TypeError,
-		"field '%U' defaults to a %.100s that cannot be hashed, so it holds "
-		"something mutable that every instance would share; default it to one "
-		"that can be, and build the rest with set_field -- from __post_init__, "
+		"field '%U' defaults to a %.100s whose type hashes and whose value will "
+		"not, which is how a container of something mutable answers; salix "
+		"shares such a default across every instance, so give the field one "
+		"that hashes and build the rest with set_field -- from __post_init__, "
 		"or from your own __init__ if the body writes one",
 		field_name,
 		kind->tp_name
@@ -807,7 +809,11 @@ static PyObject * build_defaults(PyObject * const all_names, PyObject * const de
 
 		PyObject * const stored = struct_default_copy(value);
 
-		if (stored == NULL || reject_unsafe_default(field_name, stored) != RESULT_OK) {
+		if (
+			stored == NULL ||
+			reject_unsafe_default(field_name, stored) != RESULT_OK ||
+			refuse_shared_mutable_contents(field_name, stored) != RESULT_OK
+		) {
 			Py_XDECREF(stored);
 			Py_DECREF(defaults);
 
--- a/src/fields.c
+++ b/src/fields.c
@@ -77,6 +77,7 @@ static struct special_form named_special_form(PyObject * text, struct form_probe
 static bool names_form(PyObject * text, PyObject * needle);
 static bool continues_identifier(Py_UCS4 character);
 static PyObject * module_attribute(char const * module_name, char const * attribute);
+static enum result refuse_shared_mutable_contents(PyObject * field_name, PyObject * value);
 
 /* The plan only takes references once every step has succeeded; the working
  * collections belong to this scope either way. */
@@ -684,7 +685,7 @@ static enum result reject_unsafe_default(PyObject * const field_name, PyObject *
 	Py_ssize_t const filled = struct_copies_default(kind) ? PyObject_Size(value) : 0;
 
 	if (filled <= 0) {
-		return filled < 0 ? RESULT_ERROR : RESULT_OK;
+		return filled < 0 ? RESULT_ERROR : refuse_shared_mutable_contents(field_name, value);
 	}
 
 	PyErr_Format(
@@ -694,6 +695,49 @@ static enum result reject_unsafe_default(PyObject * const field_name, PyObject *
 		"one and fill it with set_field -- from __post_init__, or from your own "
 		"__init__ if the body writes one, which displaces the constructor "
 		"__post_init__ runs from",
+		field_name,
+		kind->tp_name
+	);
+
+	return RESULT_ERROR;
+}
+
+/*
+ * A type that says it hashes and an instance that then refuses is a container
+ * whose contents are mutable, and salix shares such a default across every
+ * instance. The four copied types answer above; a deque or an array declares
+ * __hash__ = None, says so before being asked, and is left shared as msgspec
+ * leaves it.
+ */
+static enum result refuse_shared_mutable_contents(
+	PyObject * const field_name,
+	PyObject * const value
+) {
+	PyTypeObject * const kind = Py_TYPE(value);
+
+	if (kind->tp_hash == NULL || kind->tp_hash == PyObject_HashNotImplemented) {
+		return RESULT_OK;
+	}
+
+	if (PyObject_Hash(value) != -1) {
+		return RESULT_OK;
+	}
+
+	/* A writable memoryview answers ValueError where a tuple of lists answers
+	 * TypeError, and anything else -- a MemoryError, an interrupt -- is not the
+	 * instance declining and is left to propagate. */
+	if (!PyErr_ExceptionMatches(PyExc_TypeError) && !PyErr_ExceptionMatches(PyExc_ValueError)) {
+		return RESULT_ERROR;
+	}
+
+	PyErr_Clear();
+
+	PyErr_Format(
+		PyExc_TypeError,
+		"field '%U' defaults to a %.100s that cannot be hashed, so it holds "
+		"something mutable that every instance would share; default it to one "
+		"that can be, and build the rest with set_field -- from __post_init__, "
+		"or from your own __init__ if the body writes one",
 		field_name,
 		kind->tp_name
 	);
--- a/src/fields.c
+++ b/src/fields.c
@@ -256,9 +256,15 @@ static enum result append_declared(
 		/* A default is the class-body value bound to the field name. */
 		PyObject * const declared_default = PyDict_GetItem(namespace, field_name);
 
+		/* Probed here rather than where the defaults tuple is built, because that
+		 * walks the inherited ones too and would re-ask every subclass creation.
+		 * A base's defaults answered when the base was built. */
 		if (
 			declared_default != NULL &&
-			PyDict_SetItem(default_by_name, field_name, declared_default) < 0
+			(
+				PyDict_SetItem(default_by_name, field_name, declared_default) < 0 ||
+				refuse_shared_mutable_contents(field_name, declared_default) != RESULT_OK
+			)
 		) {
 			return RESULT_ERROR;
 		}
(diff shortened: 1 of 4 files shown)


JPHutchins:
Round 1 answered at `36a7d2c`. Two of the three minors are fixed; the third is a comment, and those are out of scope — see the end.

## The probe hashed every non-copied default twice

Correct. `build_defaults` checks the declared value and then the stored copy, and for anything outside the four copied types those are **the same object**, so a class author's `__hash__` ran twice per field. It runs on the stored copy only now — the object the class actually keeps, which is where the existing reasoning already put "the second read is the one that counts".

Measured with a counting `__hash__`:

```
hash calls during class creation: 1     (was 2)
```

## The message asserted a cause it cannot know

Also correct, and this is the better half of the finding. *"so it holds something mutable"* is an **inference** from a hash that failed, and the failure could be the author's own `__hash__` raising `TypeError` for unrelated reasons. It now states what was seen and what salix concluded from it:

```
field 'v' defaults to a tuple whose type hashes and whose value will not, which
is how a container of something mutable answers; salix shares such a default
across every instance, so give the field one that hashes and build the rest with
set_field -- from __post_init__, or from your own __init__ if the body writes one
```

That also answers the nit that *"default it to one that can be [hashed]"* read as "make this same value hashable".

## The third part is a deliberate trade, and now says so in a test

> lets any exception other than TypeError/ValueError propagate and abort class creation where the class previously built

True, and I am keeping it. The alternatives are worse: swallowing turns a `MemoryError` or an interrupt into "holds something mutable", and reporting it as this rule's refusal is a false claim about someone else's exception. Verified:

```python
class Angry:
    def __hash__(self): raise RuntimeError("boom")

class Bad(Struct):
    v: object = Angry()      # RuntimeError: boom
```

Pinned by `test_a_hash_that_fails_for_its_own_reasons_propagates_unchanged`, whose docstring says outright that such a class used to build and share the value. If you would rather it built, that is a one-line change and belongs to you rather than to me.

I also took the `-1` nit: `PyObject_Hash` returning -1 with no exception set would have refused with nothing to report. CPython normalises a real hash of -1 to -2 so it should not happen, and the guard costs one call.

<details>
<summary><b>The Python mirror in <code>values.py</code> — answered, not changed</b></summary>

> `_shares_mutable_contents` duplicates the C discriminator, guarded by a one-value assert

The duplication is real and deliberate, and the mitigation is the assertion that names the membership rather than deriving it:

```python
assert [type(value).__name__ for value in REFUSED_AS_DEFAULT] == ["memoryview"]
```

So the mirror selects, and the *list* is what the suite states. If the rule widens, that assertion fails and someone has to write the new member down. It earned its keep on its first run — I had built a second memoryview with the same contents, the identity test never matched the one in `EVERY`, and it fired at import.

On the third path: yes, a future value whose `__hash__` raises something other than `TypeError`/`ValueError` would raise at import of every module that imports `values`. Every value in that set is a literal in the same file with a C-level hash, so reaching it means someone adding such a value — and an import-time failure naming the file is not a bad way to find out. Left as is rather than adding a layer that makes the mirror diverge further from the C.

</details>

## The comment findings

One minor and three nits are comments and docstrings made stale by this change — `construct.c`'s sharing block, and the two `TestMutableDefaults` docstrings that still mention the writable memoryview in the shared group.

They are correct, and I am deliberately not fixing them. JPH's instruction today: *"dont worry about the comments just dont add more. i will do a deslop phase in the future to remove all comments."* Spending rounds on comment accuracy is spending them on something being deleted. Recording them here so the next round has them pre-answered rather than re-finding them.

`check` 25/25 and `flake_check` green on `36a7d2c`.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked me to drive this phase's PRs to merge and separately ruled that comments are not to be maintained ahead of a deslop pass. This answers round 1 of the automated review on this PR.

JPHutchins:
Round 2 answered at `4e65c91`. Both new minors are places the probe ran where it should not, and both are better findings than round 1's.

## The probe re-asked every inherited default, on every subclass

`build_defaults` walks the **full** default set, and `append_inherited` has already put the base's defaults into it. Reproduced exactly as described:

```
hash calls across Base + 2 empty subclasses: [1, 1, 1]
```

Cost aside, the correctness half is the part worth having: a `__hash__` whose answer depends on state could flip `class Child(Base): pass` to refused long after the base built. My previous commit's "hash a default once" held **within one class creation and not across a hierarchy** — and `test_the_probe_hashes_a_default_once` could not see the difference, which is the more useful half of your finding.

It runs in `append_declared` now, over what this body declares. A base's defaults answered when the base was built. `[1]` for the whole hierarchy.

Probing the *declared* value rather than the stored copy is the same question, which is what makes the move safe: outside the four copied types the copy **is** the declared object, and the four declare `__hash__ = None` so the probe returns before looking at anything.

## A value that holds itself is shared again

```python
loop = Loop(None)
set_field(loop, "v", loop)

class Holder(Struct):
    w: object = loop          # was: RecursionError, ~8 MB in
```

Correct, and I have treated it as a defect rather than a trade — which is the difference from round 1's `RuntimeError` case. A `RecursionError` is not the instance *declining*; it is the probe never reaching an answer. And a frozen struct is safe to share whatever it points at, itself included, so sharing is both the pre-PR behaviour and the right one. Cleared rather than propagated.

That leaves the two failure modes distinguished by what they mean:

| the hash raises | means | salix |
| --- | --- | --- |
| `TypeError` / `ValueError` | the instance declines | refuse |
| `RecursionError` | the probe cannot finish asking | share |
| anything else | not this rule's business | propagate unchanged |

Two tests added, for the two things nothing existing could observe: the per-hierarchy call count, and the self-referential default.

## The comment findings, third time

One minor (`construct.c`'s sharing block) and three docstring nits are stale because of this PR. They are correct. JPH ruled today: *"dont worry about the comments just dont add more. i will do a deslop phase in the future to remove all comments."*

So they stay, and this is the round to say it plainly rather than repeat it: **the `construct.c` minor will not be fixed on this PR**, which means the score floor here is 1 rather than 0. That is converged by the threshold, and it is the deliberate consequence of the deslop ruling rather than an oversight.

`check` 25/25 and `flake_check` green on `4e65c91`.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked me to drive this phase's PRs to merge and separately ruled that comments are not to be maintained ahead of a deslop pass. This answers round 2 of the automated review on this PR.

JPHutchins:
Round 3 answered at `01e1b89`. The new minor is a fair consequence of round 2's fix, and it is **half** right — which half matters.

## The carve-out is broader than its justification, and that is deliberate

Reproduced exactly as described:

```python
class RecursiveHash:
    def __init__(self): self.items = []
    def __hash__(self): return hash(self)

value = RecursiveHash()

class Holder(Struct):
    v: object = value          # shared

value.items.append("seen by every instance")
Holder().v.items               # ['seen by every instance']
```

Real aliasing. But it is **not a regression, and not an oversight**: it is where the rule already stood before this PR, and where it still stands for `deque`, `array` and `defaultdict` — all shared while holding mutables, because they say up front that they do not hash. The rule JPH ruled refuses what it can positively identify and shares the rest; sharing what the probe cannot *classify* is the same policy, not a hole in it.

**What was genuinely wrong is the sentence I put next to it.** The comment promised that whatever a class author's `__hash__` raises propagates, and the branch directly above it makes that untrue for exactly one exception. You caught a contradiction I introduced two commits ago. The two agree now, and the carve-out says outright that it is the incomplete half of the rule rather than a guarantee.

Both of those are comments **this PR added**, so correcting them is not maintaining the layer that is out of scope — see below.

**The price is pinned** by `test_a_mutable_whose_own_hash_recurses_is_shared_too`, which asserts the aliasing rather than describing it. A later change that narrows the carve-out has to update a test that explains why it existed.

## The special-form ordering nit — declined, with the reason

> The probe runs before the special-form check, so a ClassVar/InitVar field whose default holds a mutable is refused with the probe's message instead of the form's

Correct about the ordering. Moving the probe below the form check also moves it below the inheritance `continue`, which would stop it seeing a **re-annotated inherited field's new default** — and that is precisely the case round 2 moved it there to cover. So the choice is which of two doubly-wrong declarations gets the less apt message, and I would rather keep the coverage. Nothing about which classes build changes either way.

## The comment findings

`construct.c`'s sharing block and three test docstrings remain stale, per JPH's ruling today. As said last round, that puts the score floor on this PR at **1 rather than 0** — converged by the threshold, and a deliberate consequence rather than something left undone.

`check` 25/25 and `flake_check` green on `01e1b89`.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked me to drive this phase's PRs to merge and separately ruled that comments are not to be maintained ahead of a deslop pass. This answers round 3.

PR #94, opened 2026-08-09
feat: pickle, copy and deepcopy for structs (#13)

JPHutchins:
## Summary

Implements **#13** — `pickle`, `copy` and `deepcopy` for structs — per the ruling: `__getstate__`/`__setstate__` plus a `tp_new` on struct classes; the mixin stays uninstantiable.

**Four failure modes fixed:**
1. **No-body-`__init__` classes failed at dump** — `object.__reduce_ex__` requires `tp_new != NULL`; the vectorcall was the only allocator.
2. **Frozen classes with a body `__init__` failed at loads** — the default reducer's state restore hit the frozen `tp_setattro`.
3. **copy/deepcopy of any `tp_new` class failed** (`object.__new__(WithInit) is not safe`) — `install_fields` sets `tp_new` after the type exists, so CPython's own `__new__` wrapper never lands in the class dict; an `install_fields`-added wrapper fixes the dispatch.
4. **Protocols 0/1** — deliberately still unsupported (only a `__reduce__` could fix them, and that shape is rejected for re-running `__post_init__`); pinned in a test so the limitation is deliberate.

**The mechanism:**
- `install_fields` now always sets `tp_new = Struct_new` (vectorcall stays for body-less classes; `Struct_new` already handles the zero-arg `cls.__new__(cls)` case).
- `__getstate__`/`__setstate__` as `tp_methods` on the mixin (shared, zero per-class cost). **State shape `(values_by_name, unset_names)`** — a dict keyed by field name plus the unset names, because `None` is a value, not an absence, and a field rename in an old pickle errors loudly instead of silently shifting a positional tuple.
- **Restore writes slots directly via `write_slot`/`PyMember_SetOne`** — the constructor/`set_field` path: bypasses `tp_setattro` (frozen restores cleanly), takes its critical section internally (3.14t-safe), version-portable 3.10–3.15. `__post_init__` and `__init__` are never called on the restore path, so a custom `__init__` with a non-field signature is preserved.
- 3.10 `_PyCFunction_CAST` fallback; the `__new__` wrapper skips a body `__new__`.

**Validation:** full suite green on all six versions (1059 passed / 11 skipped / 10 xfailed — 48 new tests, the 11th xfail flipped to a pass), free-threaded green (1045 passed, new concurrent pickle-while-write test), lint/analyze/c_test/format/type_check green, gate green.

**Accepted consequences:** protocols 0/1 unsupported (pinned); `inspect.signature` on body-less struct classes becomes `ValueError` (standard for C-constructed classes); a field named `__getstate__`/`__setstate__`/`__new__` shadows the machinery (pre-existing dunder-field foot-gun, not extended).

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. JPH directed the issue queue (#13 next: pickle/copy/deepcopy, ruled `__getstate__`/`__setstate__` + `tp_new`); a max-effort design agent produced a validated plan, which I approved, and a max-effort implementation agent built it, which I gated and reviewed. This PR aggregates that work.


tests/test_pickle.py
```diff
@@ -0,0 +1,299 @@
+import copy
+import pickle
+
+import pytest
+
+from salix import Struct, set_field
+
+# The classes live at module level because pickle resolves a class by qualified
+# name, and an instance of a test-local class cannot be unpickled.
```
JPHutchins, line 11: doc slop

tests/test_free_threading.py
```diff
@@ -188,6 +199,47 @@ def work():
...
+    """__getstate__ reads each slot under the same per-slot critical section
+    repr and == use, and __setstate__ writes through PyMember_SetOne like
+    set_field does; this is the check that pickle composes the two against
+    concurrent writers, which is the crash the reader tests above pinned.
+
+    Survival is the assertion, because a segfault takes pytest with it. The
+    value assertions tolerate any state the struct ever held: the writers only
+    ever store two-element lists and two-tuples, so a restored value of any
+    other length means a reader saw a torn slot.
+    """
```
JPHutchins, line 221: doc slop

tests/test_pickle.py
```diff
@@ -0,0 +1,299 @@
...
+
+init_calls: list[int] = []
+post_init_calls: list[int] = []
```
JPHutchins, line 14: Final?

tests/test_pickle.py
```diff
@@ -0,0 +1,299 @@
...
+FACTORIES = {
+    "plain": lambda: Plain(1, 2),
+    "mutable": lambda: Mutable(1, 2),
+    "with_init": lambda: WithInit(7),
+    "frozen_with_init": FrozenWithInit,
+    "child": lambda: Child(1, 2),
+    "empty": Empty,
+    "with_post_init": lambda: WithPostInit(1),
+    "mutable_default": MutableDefault,
+}
```
JPHutchins: named tuple? Final? why are we writing non type safe slop?

JPHutchins (commented):
some slop

JPHutchins:
also not sure that thr test coverage is really adequate - are we covering nested recursive types etc, arbitrary members etc?

JPHutchins:
Marking this PR as the **reference implementation** for issue #13. Per @JPHutchins's direction, it is canceled (left open, not merged) and the work will be split post-compact into sequential, independently-mergeable slices — the fundamental `__getstate__`/`__setstate__`+`tp_new`, then reduce robustness, then getnewargs/body-init reconstruction — each a separate branch/PR, with #13 closing only when all are merged and fully tested.

This branch (`fix/pickle-copy-deepcopy`, 22 review rounds) holds the full accumulated design + tests as the reference for those slices.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. JPH directed canceling #94 as a reference and splitting #13's resolution post-compact.

Issue #84, opened 2026-08-08
test_base_combinations: the newest xfail bucket has no count bound, and three harness nits


JPHutchins:
#80's round-5 review, which converged at score 1, left four items. They are
recorded here rather than iterated, since the signal said stop.

## The one minor: the newest bucket has no bound

`tests/test_base_combinations.py` gives every strict-xfail bucket a companion
that asserts it is **wholly** broken, because an aggregate `== []` over a fully
failing bucket emits one bit and flips only on a complete fix. Four buckets have
it. `test_hashability_follows_equality_when_a_later_base_answers_it` — added in
that PR's round 4 — does not, so a partial #76 fix that settles some of its 24
identity-unhashable combinations leaves the xfail failing exactly as expected.

One assertion, matching the four beside it:

```python
assert len(hashability_disagrees_with_equality(CONTESTED_EQ)) == 24
```

except stated against the bucket rather than the literal, as the others are.

## Three nits

- `test_a_single_base_asking_to_drop_an_inherited_weakref_slot_is_the_same_bug`
  asserts the current #78 behaviour **unconditionally** rather than as an
  xfail, so fixing #78 turns it red as a plain failure rather than as an
  XPASS. It should be marked, or inverted, when #78 lands.
- The frozen write-probe is written twice — in `observe()` and again in
  `writes()` inside
  `test_a_write_and_a_delete_agree_about_whether_the_class_is_frozen`.
- `observe()`'s frozen probe catches `TypeError` but not the `IndexError` that
  `cls.__struct_fields__[0]` raises for a zero-field class. No class in the
  sweep has zero fields — every one gets `z` — so it cannot fire today, but it
  is a trap for anyone widening the catalogue.

None of these changes what the suite asserts about salix; all four are about
the harness holding its own line.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. #80 merged
> on the reviewer's convergence signal (score 1) with these outstanding, per
> her standing instruction to merge as reviews converge rather than to chase
> zero findings.

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
[...]

PR #96, opened 2026-08-09
test: pin the harness for the next #76/#78 fixes (#84)

JPHutchins:
## Summary

Test-only follow-up to #80's review (issue #84) — no `src/` changes. It pins the base-combination sweep's harness so the next two fixes (#76 options-vs-MRO, #78 weakref) turn their xfails red *correctly*:

- **Bucket bound (the one minor):** added `test_every_contested_hashability_combination_really_is_broken`, asserting `len(hashability_disagrees_with_equality(CONTESTED_EQ)) == 24` — the exact pin from the issue. CONTESTED_EQ is 272 wide but the hashability predicate fires on only 24 (all "identity-equality, unhashable": mutable-first combos made unhashable from the record while the MRO answers identity). Verified 24 on both 3.13 and 3.14.
- **Nit 1:** `test_a_single_base_asking_to_drop_an_inherited_weakref_slot_is_the_same_bug` is now `xfail(strict=True)` (same reason string as the sweep's #78 xfail) with inverted assertions — today it XFAILs; when #78 lands it XPASSes red.
- **Nit 2:** removed the duplicated `writes()` helper — the write/delete test now uses `observe(cls).frozen != (not deletes(cls))`, verified exactly equivalent by a four-row truth table.
- **Nit 3:** `observe()`'s frozen probe now catches `(TypeError, IndexError)` (zero-field classes raise IndexError at `__struct_fields__[0]`).

I simulated a partial #76 fix (settling 2 of the 24) and confirmed the new companion turns red — it catches partial fixes.

**Validation:** pytest on 3.14 (1009 passed / 10 skipped / 12 xfailed), base-combinations on 3.13 + 3.14 (27 passed / 8 xfailed), ruff clean. Test-only change — static type checks scope to `tests/typing/`, untouched.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins. JPH directed the issue queue (parallelizable groups) and the keep-subagents-alive discipline; this is the #84 test-harness PR from that work.

### DIFF
--- a/tests/test_base_combinations.py
+++ b/tests/test_base_combinations.py
@@ -160,7 +160,7 @@ def observe(cls: type) -> Behaviour:
     try:
         setattr(instance(cls), cls.__struct_fields__[0], 99)
         frozen = False
-    except TypeError:
+    except (TypeError, IndexError):
         frozen = True
 
     return Behaviour(
@@ -307,13 +307,6 @@ def test_a_write_and_a_delete_agree_about_whether_the_class_is_frozen():
     other is frozen against an assignment and not against a `del`.
     """
 
-    def writes(cls: type) -> bool:
-        try:
-            setattr(instance(cls), cls.__struct_fields__[0], 99)
-            return True
-        except TypeError:
-            return False
-
     def deletes(cls: type) -> bool:
         try:
             delattr(instance(cls), cls.__struct_fields__[0])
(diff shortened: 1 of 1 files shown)


tests/test_base_combinations.py
```diff
@@ -589,6 +586,20 @@ def test_hashability_follows_equality_when_a_later_base_answers_it():
...
+    """The bound for the xfail above, and the one the bucket cannot anchor:
+    the hashability predicate is narrower than the contested-equality bucket
+    it runs over, so only the 24 mutable-first combinations, made unhashable
+    from the record while the MRO answers identity, violate, and the pin is
+    that count rather than the bucket's 272.
+
+    Delete this with the xfail above when #76 lands; they describe the same
+    defect from opposite sides.
+    """
+
```
JPHutchins, line 599: doc slop

JPHutchins (commented):
doc slop

PR #98, opened 2026-08-10
feat: the reconstruction and reduce layer — copy/deepcopy, getnewargs, completeness (rest of #13, slice B)

JPHutchins:
**Part 2 of 2 resolving #13** (pickle, copy and deepcopy). Stacked on #97 (Slice A, base = `feat/pickle-slice-a`); the diff is the remaining feature on top of A's three-part state + tp_new. Once #97 merges, this retargets to `main` (the commits become identical).

## What this adds

- **Reconstruction handling.** A per-type thread-local marker (`struct_reconstruction_type`) is raised only for the type a reduce reconstruction is building, so `Struct_new` can tell a reconstruction (whose arguments come from `__getnewargs__` and must bind into the fields) from a normal construction, and a construction nested inside a reconstruction's `__new__` body of a different type reads as construction. A body `__new__` is routed through the wrapper on every construction path; the `__new__` fallback to `Struct_new` is reconstruction-only.
- **The reduce layer.** `__reduce_ex__` defers to object's when a custom `__reduce__`/`__reduce_ex__` is in play, and otherwise walks the MRO to find a `__getnewargs_ex__`/`__getnewargs__` declaration. The `__reduce_newobj__` / `__reduce_newobj_ex__` reconstruction helpers (with `__reduce_newobj_check__` variants for tests) validate their tuple arguments and raise `TypeError` rather than segfaulting on a malformed or hostile reduce — the round-21 crash, shipped fixed.
- **The completeness gate.** A getnewargs reconstruction is complete only when every required field is set; an under-providing `__getnewargs__` raises `missing required argument` instead of round-tripping a half-built struct. The None-state skip honors the stdlib contract (no `__setstate__` for a None state) only when the class overrides `__setstate__`; a struct's own `__setstate__` still normalizes None to the empty state so the gate runs.
- **copy/deepcopy.** `__deepcopy__` honors the dispatch_table/reduce path (deepcopying through copyreg rather than re-implementing it), a string reduce (copy's `_reconstruct`) round-trips, `__getstate__` is read-only, and the instance-dict restore is atomic on a failed merge (build-then-swap).

This is the reference branch's rounds 4-22 shipped whole: the getnewargs feature (declaration probe → argument binding → completeness gate → the reduce helpers) is one indivisible unit, and every intermediate round state either accepted a getnewargs struct's arguments and silently dropped them, or accepted a half-built reconstruction with no error. No safe merge point exists inside it.

Gated on the full matrix (3.10-3.15 + 3.14t): 120 pickle tests, 1097 passed / 35 skipped / 10 xfailed, `c_test`, `c_analyzer`, `format_check`, `flake_check` all green.

### DIFF
--- a/src/construct.c
+++ b/src/construct.c
@@ -5,6 +5,22 @@
 #include "result.h"
 #include "types.h"
 
+/* Py_TPFLAGS_INLINE_VALUES and Py_TPFLAGS_MANAGED_DICT joined the public
+ * object.h in 3.13 and 3.11 respectively; the flag values have been stable, so
+ * they are spelled here rather than gated on the public header. */
+enum {
+	STRUCT_TPFLAGS_INLINE_VALUES = 1 << 2,
+	STRUCT_TPFLAGS_MANAGED_DICT = 1 << 4,
+};
+
+/* A heap type's instance dict lives just before the object header on a
+ * managed-dict build (3.11+). Reading it directly, rather than through
+ * _PyObject_GetDictPtr, avoids materializing an inline-values dict: that
+ * helper creates and stores one for a null slot, which would make __getstate__
+ * mutate the instance. The offset is CPython's MANAGED_DICT_OFFSET. */
+
+_Thread_local PyTypeObject * struct_reconstruction_type = NULL;
+
 struct field_lookup {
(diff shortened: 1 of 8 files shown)


JPHutchins:
Reopening to trigger the pull_request CI dispatch (stacked-base PRs don't fire it on open).

JPHutchins (commented):
lots of doc slop, comments should be audited and removed.

Issue #99, opened 2026-08-10
pickle for structs — deferred; approaches learned from the canceled attempts


JPHutchins:
**Scope:** pickle support for salix structs — deferred by ruling (2026-08-10): copy and deepcopy land first as two PRs, pickle follows once they establish the state machinery. This issue records what the canceled attempts (#94, the slice PRs #97/#98) learned, so the future pickle work starts from knowledge rather than scratch.

<details>
<summary>The foundation copy/deepcopy will establish (pickle builds on it)</summary>

The approved mechanism, proven across ~30 review rounds and now shipping via the copy/deepcopy PRs:

- **Three-part state** `(values_by_field_name, unset_names, instance_dict_or_none)` — a positional 2-tuple is lossy because `None` is a value, not an absence. A dict key shadowing a field name lives in the third element, never merged into values.
- **`tp_new = Struct_new` on every struct class** (pickle's reducer requires it), plus a per-class `__new__` wrapper validating `PyType_IsSubtype`. A body `__new__` is honored on direct `X.__new__(X)`; routing it through every construction path is the reconstruction-layer work (below).
- **`__setstate__` validates the whole state before writing** (atomic), writes slots via the member descriptor (bypasses `tp_setattro`, so frozen structs restore), restores the instance dict by building a replacement up front and swapping under the instance critical section (managed-dict/inline-values safe; the replaced dict's finalizers must run outside the lock).
- **`__getstate__` is read-only** — never materializes an absent dict (managed-dict builds always have one; pre-3.13 builds report None for an untouched slot).
- Non-struct mixin subclasses ("impostors") delegate to `object.__getstate__`/pickle's `load_build` shapes.
</details>

<details>
<summary>The pickle-specific machinery (the deferred work)</summary>

Everything beyond copy/deepcopy, which the reference branch built in rounds 4-22 and the slice split then re-derived:

1. **The reconstruction layer.** A per-type thread-local marker (`struct_reconstruction_type`) is raised only for the type a reduce reconstruction is building, so `Struct_new` can tell a reconstruction (whose args come from `__getnewargs__` and must bind into the fields) from a normal construction; a construction nested inside a reconstruction's `__new__` body of a *different* type reads as construction. A body `__new__` is routed through the wrapper on every construction path via `dispatch_new`/`Struct_new_thunk`/`struct_body_new`.
[...]

PR #101, opened 2026-08-11
feat: __deepcopy__ for structs (deepcopy support)

JPHutchins:
## Summary

Implements `copy.deepcopy` for structs — the second slice of #13 (copy was #100, pickle deferred to #99). `Struct_deepcopy` (METH_O) mirrors `Struct_copy` and shares its dispatch machinery through a name-parameterized prologue and delegate.

<details>
<summary>Semantics</summary>

- **Dispatch order** (reproducing copy.py): the mixin's `__deepcopy__` is found by `getattr(x, '__deepcopy__')`; inside it, a co-base `__deepcopy__` is deferred to first, then `copy.dispatch_table`, then the raw path.
- **The co-base deferral resolves `__deepcopy__` on the instance**, because copy.py's deepcopy looks it up with `getattr(x, '__deepcopy__')` while its copy uses `getattr(cls, '__copy__')`. Instance resolution applies the descriptor to the instance: a plain function becomes a bound method, a classmethod binds to the concrete class, and an unset member-descriptor slot raises the AttributeError that `getattr(x, name, None)` reads as "no method" — falling through to the raw path exactly as copy.py would. (The `__copy__` side keeps the merged class-level resolution, unchanged.)
- **The raw path memoizes the shell before copying any state**, so a field (or instance-dict value) that references the source resolves to the shell instead of re-entering deepcopy — `copied.self_field is copied`. The slot values are snapshotted under the same single-section read the copy path uses, then each shallow copy is replaced by its deep copy outside the section (`copy.deepcopy(value, memo)`), sharing the memo so a value reached from two slots or the dict is copied once.
- **Delegation**: impostors go through the same reduce chain copy.py would, with the memo passed to `copy._reconstruct` (copy passes `None`), and `"un(deep)copyable object of type ..."` copy.Error.
- **Documented trade-offs**: the mixin's `__deepcopy__` shadows a body/co-base `__reduce_ex__`/`__reduce__` for deepcopy (as the mixin's `__copy__` already does for copy) — copy.py's precedence is `__deepcopy__` first, and the struct's extension point is a body `__deepcopy__`, which the deferral honors; the reduce machinery is pickle's protocol, deferred to #99. The dict branch snapshots before deep-copying (the free-threading fix), so a reference to the source's own `__dict__` aliases to a different object than the copy's installed `__dict__`.
- A non-dict memo raises `TypeError("__deepcopy__() argument must be a dict")` (the reference's check); `refuse_mixin_method_fields` now refuses a field named `__deepcopy__` alongside `__copy__`; the stub declares `def __deepcopy__(self, memo: dict[Any, Any]) -> Self`.

</details>

<details>
<summary>Verification</summary>

- Full `check` gate green: pytest 3.10–3.15 + 3.14t free-threaded (incl. two new free-threading tests racing a deepcopy against a writer), c_format, c_tidy, c_analyzer, c_test, ruff, mypy/pyright/ty, tooling.
- Discriminating tests verified against broken variants: the self-reference/cycle tests fail with RecursionError on a late-memoize build; the new refcount test fails on a store-without-decref build (transient snapshot refs leaked).
- 31 new tests in `tests/test_deepcopy.py` mirroring `test_copy.py`; refcount and free-threading coverage added for the new path.

</details>

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-flash-0731 on behalf of @JPHutchins. The prompt: implement deepcopy support for salix structs (PR2 of issue #13) per the recorded design — memoize-shell-first, copy.py-faithful dispatch — then gate, and iterate with the automated reviewer until converged.


JPHutchins:
**Round 1 — 6 minor + 2 nit. 4 fixed with discriminating tests, 4 refuted with measurement.** New commits: `79e09bb`.

<details>
<summary>Fixed (4)</summary>

**F1 (minor) — presupplied memo entry now wins.** `Struct_deepcopy` reproduces copy.py's entry check (`y = memo.get(d, _nil); if y is not _nil: return y` — [copy.py](https://github.com/python/cpython/blob/main/Lib/copy.py)) at the top of the method, so a direct protocol call honors a seeded entry instead of overwriting it with the fresh shell. The check runs before the co-base deferral, mirroring copy.py where the memo wins over everything. Test extended: `instance.__deepcopy__(memo) is stand_in`.

**F2 (minor) — the non-dict-memo TypeError is now unreachable-inconsistent no more.** The `PyDict_Check` guard is hoisted above the `is_struct` branch; both paths refuse a non-dict memo with the same TypeError. New test: `Impostor([1]).__deepcopy__(None)` raises.

**F3 (minor) — the dispatch_table gate now matches copy.py per operation.** Confirmed against the installed 3.13 copy.py: `copy` gates the branch on identity (`if reductor is not None:`), `deepcopy` on truthiness (`if reductor:`) — the one gate that differs between the operations. The prologue takes a `dispatch_truthy` flag; the deepcopy path skips falsy entries. Measured before fixing: `copy.dispatch_table[S] = 0; copy.deepcopy(S())` returns `999` (entry skipped) while `copy.copy(S())` raises TypeError (entry called). New test: `copy.dispatch_table[Point] = 0; copy.deepcopy(Point(1, "two"))` returns the raw copy. (This fix caught my own NULL-deref: the first version ran `PyObject_IsTrue(NULL)` when no entry existed — segfault on every struct deepcopy — fixed and re-gated.)

**F5 (minor) — the dict branch deep-copies a snapshot.** `PyDict_Copy(dict)` runs in C under the dict's own lock, so the snapshot is a consistent moment; `copy.deepcopy` then walks the snapshot, which no writer can change size under. The free-threading test's writer now changes dict size every round (insert + pop); verified the pre-fix build fails that test on 3.14t with exactly `RuntimeError('dictionary changed size during iteration')`.

</details>

<details>
<summary>Refuted with measurement (4)</summary>

**F0 (minor) — body `__reduce_ex__` bypassed: this is copy.py's own precedence.** `deepcopy` dispatches on `getattr(x, '__deepcopy__', None)` and only reaches the reduce chain when no `__deepcopy__` exists. A plain class with a body `__deepcopy__` has its `__reduce_ex__` bypassed identically — measured: `class R: def __deepcopy__(self, m): return 'deepcopy-won'; def __reduce_ex__(self, p): return (int, (999,))` → `copy.deepcopy(R())` is `'deepcopy-won'`. The struct's mixin `__deepcopy__` is the class's `__deepcopy__`; a body that wants different deepcopy semantics defines its own `__deepcopy__`, which the deferral honors. The reduce chain is pickle's protocol, deferred to #99.

**F4 (minor) — the shell stays in the memo on failure: copy.py does the same.** Measured: `copy.deepcopy([1, 2, Boom()], memo)` where Boom's `__deepcopy__` raises leaves `memo[id]` → the partial list `[1]`. A retry with the same memo returns the partial container. Our shell is the container equivalent; the reference implementation (commit `5099d0f^`) likewise registers without cleanup.

**F6 (nit) — fixtures stay local.** Every test module in this repo defines its own module-level structs (`test_refcounts.py`'s `Sentinel`/`Pair`, `test_free_threading.py`'s `Point`, the typing tests); `tests/conftest.py` is 7 lines of path setup and hosts no fixtures. The mirrored `test_deepcopy.py`/`test_copy.py` modules follow that convention.

**F7 (nit) — the dispatch tail stays explicit.** The two struct-path tails genuinely differ (the `dispatch_truthy` gate and the reconstruct memo), so a shared helper would thread those very differences through parameters — more indirection, not less. The shared surface is ~6 lines at 2 sites, and the merged `__copy__` shape went through 14 review rounds.

</details>

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-flash-0731 on behalf of @JPHutchins. The prompt: respond to the automated review of PR #101 (deepcopy support) — fix the accepted findings with discriminating tests, refute the rest with measurements of copy.py behavior, and reply per round.

JPHutchins:
It feel like doc slop is proliferating again. Comments should only be used as a last resort. Remove the doc slop.

## How to Avoid Code Comments

Excess prose existing within the code base is a critical failure and classified as LLM Doc-Slop. Here are some ways to avoid this dangerous SSOT violation:

IMPORTANT: do not remove existing comments unless explicitly asked to do so.

1. Self-documenting types with verbose names and fields
2. Verbose variable names
3. Helper functions that clarify intent (careful — it is possible to make the code less readable by having too many single-use helpers!)
4. Unit tests that demonstrate why something was done with positive and negative cases — unit tests are allowed to have some (not much) extra prose as long as it is tightly coupled to the code
5. Doctests (testable docstrings, e.g. in python and rust)
6. IMPORTANT: And finally, commit messages — commit messages are where the verbose prose ends up when all else fails. Verbose explanations of what was done and why is tied to a concrete instance of the code base. Drift is impossible -- well, squashes and other git history rewrites need to be careful! The commit message can even contain links and citations to external resources - an opt-in context goldmine for future agents and developers.

JPHutchins:
Removed — `ef7fcf3`, gate green before pushing.

<details>
<summary>What went, what stayed</summary>

Removed (restatement or wrong):
- The descriptor-protocol enumeration in `deferred_co_base_copy` (plain function → bound method, classmethod → leaf, unset member slot → AttributeError) — visible in the `tp_descr_get` branch itself.
- "None reads as no method under both gates" — restated the `resolved == Py_None` test, and its "`if copier:` (copy)" citation was factually wrong (copy.py gates both method lookups on `is not None`).
- The prologue block's narration of the `argument` call, and the reduce-branch comment's non-tuple/TypeError clause (PySequence_Tuple's own error).

Kept (mysterious why, invisible from the code):
- memoize-shell-first — why the shell enters the memo before the state is copied.
- `PyDict_Copy` snapshot — why the dict branch copies before deepcopying.
- the per-operation dispatch gate — copy.py's identity-vs-truthiness asymmetry.

The removed prose's still-valuable content is recorded in the commit message, per "commit messages are where the verbose prose ends up".

</details>

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-flash-0731 on behalf of @JPHutchins. The prompt: remove the doc slop from PR #101's new comments per the How-to-Avoid-Code-Comments rule, keep only the mysterious-why comments, and record the removed prose in the commit message.

JPHutchins:
**Round 2 (head `dcb3f30`) — 3 minor + 6 nit. 6 fixed with discriminating tests, 3 refuted with measurement; PR body updated with the documented trade-offs.**

<details>
<summary>Fixed (6)</summary>

**F4 (minor) — the deferral can no longer re-find the mixin's own method.** The MRO scan now starts after `_StructMixin`: a method before it was already found by `getattr` and called by copy.py directly, and a body rebinding the mixin's own `__deepcopy__` (`__deepcopy__ = Struct.__deepcopy__`) used to be re-found by the scan and re-enter `Struct_deepcopy` forever. Measured pre-fix: `RecursionError` on both a plain rebind and one over a co-base. Post-fix the rebind degrades to the raw copy — new test `test_a_rebound_mixin_deepcopy_does_not_reenter_forever`. (The `__copy__` counterpart is fixed by the same change.)

**F2 (nit) — direct protocol calls now register the result.** The deferred, dispatch_table and reduce branches of `Struct_deepcopy` register their result in the memo before returning, matching the raw path's shell and copy.py's own `memo[d] = y` on success. New test: `test_a_direct_protocol_call_registers_the_deferred_result`.

**F5 (nit) — a failed deepcopy no longer poisons the memo.** `memo_failure` deletes the shell's memo entry on every error return after registration, so a retry with the same memo rebuilds fresh instead of returning the partial shell. Verified discriminating: the new test fails on a build without the deletion.

**F3 (nit) — the free-threading writer now changes the dict's size every round.** The writer inserts a fresh key (`i % 65536`) and pops the key from 128 rounds earlier (`(i - 128) % 65536`) each round — the key space far exceeds the trailing window, so every round grows and shrinks the dict instead of converging to a constant size after warm-up (the old pattern, confirmed by simulation: 128 size changes in 600 rounds).

**F7 (nit) — the stub memo is `dict[int, Any]`**, the key type copy.deepcopy actually passes, so an accurate subclass override is LSP-legal while the common `{}` spelling still type-checks.

**F8 (nit) — the refusal derives from the mixin's type dict** instead of a hand-maintained names array: the `tp_methods` entries arrive in `StructMixin_Type`'s dict as method descriptors, so a third mixin method is refused automatically and the two lists cannot drift.

</details>

<details>
<summary>Refuted with measurement (3)</summary>

**F0 (minor, re-raised) — body `__reduce_ex__` bypass.** The round-1 refutation stands: a plain class with a body `__deepcopy__` has its `__reduce_ex__` bypassed by copy.py identically (measured), so the struct's behavior is exactly "a class with this `__deepcopy__`". Consulting reduce would pre-build pickle's reconstruction machinery (the #99 deferral) and diverge from the `__deepcopy__`-wins contract — and the mixin's `__copy__` has shadowed reduce for copy since before #100, blessed across 14 rounds. The change from pre-PR2 is real and now **documented in the PR body**: the extension point for custom deepcopy semantics is a body `__deepcopy__`, honored by the deferral; reduce stays pickle's protocol.

**F1 (nit) — the `_nil` sentinel edge.** The divergence requires the caller to fish `copy.deepcopy.__defaults__[1]` and plant it as a memo value — copy.py itself never writes the sentinel as a value, only as the `get` default. Noted in the reply rather than coded.

**F6 (nit) — dict aliasing divergence.** The snapshot is the deliberate fix for the size-changing-writer race (verified: the live-dict build raises `RuntimeError('dictionary changed size during iteration')` on 3.14t). A field referencing the source's own `__dict__` therefore aliases to a different object than the copy's installed `__dict__`; **documented in the PR body** as the trade-off, per the reviewer's own "accept the snapshot trade-off" option.

</details>

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt: answer the automated round-2 review of PR #101 (deepcopy) — fix the accepted findings with discriminating tests, refute the rest with measurements, document the trade-offs in the PR body, and reply per round.

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. I decoded round 4's findings from the review comment, implemented the three fixes with discriminating tests, measured the fourth against copy.py's `_keep_alive`, and ran the camas gate before pushing.

Round 4 answered in `8f53851`: three nits fixed (each with a discriminating test), one refuted by measurement.

<details>
<summary>F0 — identity results memoized (fixed)</summary>

All three registration sites now skip a result that is the source itself, matching copy.py's `if y is not x: memo[d] = y`:

```c
if (deferred != self && PyDict_SetItem(memo, key, deferred) < 0) {
```

and the same guard on the impostor and dispatch/reduce paths. Three tests, one per site: an identity co-base `__deepcopy__`, a string-returning `dispatch_table` copier, and an impostor string `__reduce__` — each asserts `copied is instance` and `id(instance) not in memo`.
</details>

<details>
<summary>F1 — unbounded mixin walk (fixed)</summary>

The walk is bounded by `PyTuple_GET_SIZE(mro)`; a mixin-less MRO returns NULL (the function's "no method" result), so the caller falls through to the dispatch path instead of reading past the tuple.
</details>

<details>
<summary>F3 — rollback replacing the real error (fixed)</summary>

`memo_failure` now wraps the delete in `PyErr_Fetch`/`PyErr_Restore`, so a field whose `__deepcopy__` removes the source's entry before raising keeps its own exception. Test: `test_a_field_that_clears_the_memo_still_raises_its_own_exception`. The dict-subclass half of the finding does not apply — `PyDict_DelItem` operates on the dict internals and bypasses `__delitem__` — but the fetch/restore fix covers the failure mode either way.
</details>

<details>
<summary>F2 — dead snapshot entry (refuted by measurement)</summary>

The snapshot never dies. The dict branch calls `copy.deepcopy(snapshot, memo)`, and copy.py's own tail runs for that call — `if y is not x: memo[d] = y; _keep_alive(x, memo)` — which appends the snapshot to the memo's keep-alive list, pinning it for the memo's lifetime. Measured on 3.14.6 with a dict-bearing struct and a direct `__deepcopy__(memo)` call:

```pycon
>>> memo[id(memo)]
[['world'], {'extra': ['world']}]
```

the keep-alive list holding both the deep-copied list and the snapshot dict itself. `_keep_alive` exists and is called by `copy.deepcopy` on every supported version (checked 3.10, 3.11, 3.12, 3.13, 3.14). The `memo[id(snapshot)] = deep_dict` entry is the standard memo entry copy.py makes for every intermediate object it deep-copies; the keep-alive list is exactly the mechanism that makes id reuse safe, and the snapshot is on it.
</details>

Commit `8f53851`; the full `check` gate is green (6-version pytest matrix + 3.14t + analyzers + lint + type check).

Issue #53, opened 2026-08-01
functools.cached_property cannot be used on a struct, and nothing says so


JPHutchins:
A struct's fields are its slots and there is no instance `__dict__`, so `cached_property` has nowhere to put the value:

```python
class Cached(Struct):
    x: int

    @functools.cached_property
    def expensive(self) -> int:
        return self.x * 100

Cached(2).expensive
# TypeError: No '__dict__' attribute on 'Cached' instance to cache 'expensive' property.
```

**This is parity, not a shortfall.** Measured:

```
salix                        TypeError: No '__dict__' attribute ...
dataclass(frozen, slots)     TypeError: No '__dict__' attribute ...
NamedTuple                   TypeError: No '__dict__' attribute ...
dataclass(frozen) no slots   200
```

Only a dataclass carrying a `__dict__` succeeds, and carrying one is precisely what salix exists not to do. So this is not a bug to fix — it is a thing a user will try, hit, and have to work out for themselves.

<details>
<summary><b>Two things that do work, both pinned in tests/test_methods.py</b></summary>

**`functools.cache` on a method.** Caches on the function rather than the instance, so the absent `__dict__` is not in its way. The cost is that the cache holds the struct alive for the life of the class, which is why `ruff` flags `B019`.

```python
class Computed(Struct):
    x: int

    @functools.cache
    def slow(self) -> int:
        return self.x * 100
```

**A declared field, filled by `__post_init__`.** This is the one a slotted dataclass cannot copy, because it has no way through the frozen wall:

```python
class Cached(Struct):
    x: int
    expensive: int = 0

    def __post_init__(self) -> None:
        set_field(self, "expensive", self.x * 100)
```

Eager rather than lazy, and it costs a slot — but it is a genuine answer, and it is salix-shaped.

</details>

**What this issue is asking for:** a line in the README, once #18 is written. Not a `__dict__`, and not a capability.

The alternative — a `dict=True` class keyword that adds an instance dict — would make `cached_property` work and would cost every struct that opted in the memory and the per-instance allocation the design is built to avoid. Recording it as the option that exists and was not taken.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by claude-opus-5 on behalf of @JPHutchins, who asked for gaps found while testing methods on structs to be filed rather than fixed.

JPHutchins:
Is this still true, actually?

PR #102, opened 2026-08-13
fix: settle a delegate-built struct class with the call's options and defaults

### LINKED ISSUE #55: A metaclass `__new__` that hands back a struct class drops the options and defaults of the call that delegated to it

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to keep working the pre-publish batches — after #101 merged, the residue (#55, #57, #53), one PR each, iterated with the automated reviewer.

## Semantics

`META("Built", (Base,), namespace, order=True)` over a `StructMeta` subclass with a Python `__new__` re-enters the build with no keywords and a namespace `drop_class_variables` has already stripped, so the class the delegate hands back has its fields installed but none of the call's options or defaults — and the outer call used to refuse it.

The outer call now **settles** that class when it is the one the call describes, and keeps the refusal otherwise:

<details>
<summary>What the settle applies, and what is still refused</summary>

- **Settled** — the plan's defaults (`struct_defaults`/`struct_default_count`), the call's options (`struct_options`, so `order` takes effect at compare time), `struct_resolves_body_eq`, and the dunder rebindings for every option that changed across the re-entry: eq (the five comparison names, plus `__ne__` when the body did not answer equality), repr, frozen (the mutability pair and the hash, following `bind_hash`'s body-eq rule), and match_args. Each arm reproduces what a fresh-namespace `apply_options` would have bound, skipping names the body defined.
- **Refused** (the same TypeError) — a class whose field names, name, or layout base differ from the plan, so a class the delegate found lying around is never overwritten with this call's plan; a renamed class, refused by the `ht_qualname`/`ht_name` check; a weakref-slot difference, because a slot layout is baked at creation and cannot be settled afterwards.
</details>

## Tests

The strict `xfail` on `test_both_survive_a_delegate_that_writes_its_own_new` flips: the forwarding delegate now carries the defaults and `order=True` through. New pins: the foreign-class refusal and the rename refusal, one test per settled option (`frozen`, `eq`, `repr`, `match_args`), and the weakref refusal.

Closes #55.

### DIFF
--- a/src/meta.c
+++ b/src/meta.c
@@ -126,11 +126,26 @@ static enum result install_fields(
 	struct options options,
 	bool resolves_body_eq
 );
-static enum result reject_unless_planned(
-	StructType const * struct_class,
+static enum result settle_planned(
+	StructType * struct_class,
+	StructType const * base,
+	PyObject * name,
 	struct field_plan const * plan,
-	struct options options
+	PyObject * original_namespace,
+	struct options options,
+	struct options inherited,
+	bool frozen_across_bases,
+	bool body_defines_eq,
+	bool inherits_body_eq,
+	bool derive_not_equal
 );
+static enum result settle_rebind(
+	StructType * struct_class,
+	PyObject * original_namespace,
+	char const * const * names,
+	bool from_mixin
+);
+static enum result refuse_unplanned(StructType const * struct_class);
 static enum result install_post_init(StructType * struct_class);
 static bool defines_own_init(StructType const * struct_class);
 static PyObject * StructMeta_call(PyObject * self, PyObject * args, PyObject * keywords);
@@ -307,6 +322,7 @@ static PyObject * build_struct_class(
 	}
 
 	bool const inherits_body_eq = inherited_equality.from_a_body;
+	bool const frozen_across_bases = request.options.frozen && any_struct_base_is_mutable(bases);
 
 	PY_OWNED(
 		namespace,
@@ -317,7 +333,7 @@ static PyObject * build_struct_class(
 			request.options,
 			base,
 			inherited,
-			request.options.frozen && any_struct_base_is_mutable(bases),
+			frozen_across_bases,
 			body_defines_eq,
 			inherits_body_eq,
 			inherited_equality.needs_derived_not_equal
@@ -337,7 +353,19 @@ static PyObject * build_struct_class(
 				request.options,
 				body_defines_eq || inherits_body_eq
 			) :
-			reject_unless_planned(struct_class, &plan, request.options)
+			settle_planned(
+				struct_class,
+				base,
+				name,
+				&plan,
+				original_namespace,
+				request.options,
+				inherited,
+				frozen_across_bases,
+				body_defines_eq,
+				inherits_body_eq,
+				inherited_equality.needs_derived_not_equal
+			)
 		);
 
 		if (settled != RESULT_OK) {
@@ -1088,11 +1116,74 @@ static PyTypeObject * winning_metatype(PyTypeObject * const requested, PyObject 
 	return winner;
 }
 
-static enum result reject_unless_planned(
-	StructType const * const struct_class,
-	struct field_plan const * const plan,
-	struct options const options
+static enum result settle_rebind(
+	StructType * const struct_class,
+	PyObject * const original_namespace,
+	char const * const * const names,
+	bool const from_mixin
 ) {
+	PyObject * const source = (
+		from_mixin ? (PyObject *) &StructMixin_Type :
+		(PyObject *) &PyBaseObject_Type
+	);
+	PY_OWNED(class_dict, struct_type_dict(&struct_class->heap_type.ht_type));
+
+	if (class_dict == NULL) {
+		return RESULT_ERROR;
+	}
+
+	for (char const * const * name = names; *name != NULL; name += 1) {
+		if (PyDict_GetItemString(original_namespace, *name) != NULL) {
+			continue;
+		}
+
+		PY_OWNED(bound, PyObject_GetAttrString(source, *name));
+
+		if (bound == NULL || PyDict_SetItemString(class_dict, *name, bound) < 0) {
+			return RESULT_ERROR;
+		}
+	}
+
+	return RESULT_OK;
+}
+
+static enum result refuse_unplanned(StructType const * const struct_class) {
+	PyErr_Format(
+		PyExc_TypeError,
+		"%.200s.__new__ returned a struct class this call did not plan: its "
+		"metaclass is re-entered without the keywords or the class body's "
+		"defaults, so what it built is not what was asked for",
+		Py_TYPE(struct_class)->tp_name
+	);
+
+	return RESULT_ERROR;
+}
+
+static enum result settle_planned(
+	StructType * const struct_class,
+	StructType const * const base,
+	PyObject * const name,
+	struct field_plan const * const plan,
+	PyObject * const original_namespace,
+	struct options const options,
+	struct options const inherited,
+	bool const frozen_across_bases,
+	bool const body_defines_eq,
+	bool const inherits_body_eq,
+	bool const derive_not_equal
+) {
+	static char const * const comparison[] = {
+		"__eq__",
+		"__lt__",
+		"__le__",
+		"__gt__",
+		"__ge__",
+		NULL,
+	};
+	static char const * const not_equal[] = {"__ne__", NULL};
+	static char const * const representation[] = {"__repr__", NULL};
+	static char const * const mutability[] = {"__setattr__", "__delattr__", NULL};
+	static char const * const hash_name[] = {"__hash__", NULL};
 	PY_OWNED(planned, PyList_AsTuple(plan->all_names));
 
 	if (planned == NULL) {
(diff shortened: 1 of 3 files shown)


JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to iterate this PR with the automated reviewer.

Round 1 answered — all six findings fixed in `93d4304`, each with a discriminating test. Gate green across the matrix before the push.

**Executive summary:** the rebinds now go through the type (so the slots follow), the match_args-off arm restores a body-defined `__match_args__`, the weakref gate checks the class's actual slot instead of the option pair, and the name check reads `ht_name`. The six pass-through tests share one `Forwarding`.

<details>
<summary><b>F1 + F5 — slot-following rebinds, pinned by a stripping delegate</b></summary>

The pass-through delegate masked the arms because the outer call's own namespace injection survives into the re-entered build — the class already carried the outer bindings. A delegate that *strips* the dunders is the case that matters, and there the built class inherits the mixin's C slots (`tp_hash`, `tp_richcompare`, `tp_setattro`, `tp_repr`), which never consult the dict — so dict writes were inert. `settle_rebind` and the unhashable arm now write through `PyType_Type.tp_setattro`, so `update_slot` recomputes the slots exactly as `type_new` would have from a fresh namespace. The new `test_a_delegate_that_strips_the_dunders_is_settled` strips all ten names and asserts mutation, identity equality, object repr, and `hash(instance) == object.__hash__(instance)` — each assertion fails against the dict-write version.
</details>

<details>
<summary><b>F0 — body `__match_args__` restored</b></summary>

The fresh path only ever *adds* `__match_args__` when wanted, so a body-defined one survives `match_args=False`; the settle's delete diverged. It now restores the body's value (dict write — `__match_args__` has a get-only type getset, so it cannot go through `setattro`), and deletes only when the body defined none. Pinned by `test_a_body_match_args_survives_the_delegate_when_turned_off`.
</details>

<details>
<summary><b>F2 + F3 — weakref per direction</b></summary>

- The settle gate now checks the built class's `tp_weaklistoffset` against the plan's `weakref=True`, so `weakref=False` over a weakref-carrying base settles exactly as the fresh path accepts it (the slot is inherited). Pinned by `test_a_weakref_base_keeps_its_slot_when_the_delegate_turns_weakref_off`.
- The True-over-False direction is refused **before** `create_class` — when the call is a direct `StructMeta` call, the plan wants weakref, no base carries the slot, and the winning metatype's `tp_new` will hand the build off — with a message naming the hand-off (`"weakref=True cannot cross a metaclass __new__ that hands the build off..."`). Previously it died in the inner call's displaced-slot check with an error about `__slots__` the caller never wrote. Pinned by `test_a_delegate_cannot_add_a_weakref_slot_the_call_planned` (matches `"cannot cross"`).
</details>

<details>
<summary><b>F4 — `ht_name` for the name check</b></summary>

A body-set `__qualname__` is legal and the fresh path accepts it, so only `ht_name` (a genuine rename of the call's name) refuses. Pinned by `test_a_body_qualname_does_not_upset_the_settle`, which also asserts the qualname survives.
</details>

<details>
<summary><b>F6 — one shared `Forwarding`</b></summary>

Module-level `Forwarding(META)` replaces the six inline copies; `Wrong` and `Renaming` stay inline because they behave differently.
</details>

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to iterate this PR with the automated reviewer.

Round 2 answered in `2f7d58e` — both findings and the systemic problem, fixed by construction rather than by patching the two instances.

**Executive summary:** the settle no longer has its own copy of the option conditions. `binding_plan()` computes every binding decision once (comparison, `__ne__`, representation, mutability, hash action, match_args) and both paths execute it — `apply_options` into the namespace, `settle_planned` onto the live class. The two drift instances die with the copy.

<details>
<summary><b>F0 — body dunders restored, even against the inner's own re-injection</b></summary>

`restore_stripped` runs before the plan's arms: for every body-defined name, if the class dict's entry is not the body's own object (by identity), the body's value is rebinding through `setattro`. The identity check matters beyond plain stripping — the mutated namespace changes the *inner* build's body flags, so the inner's `bind_hash` injects the mixin's hash over a stripped body hash; the restore catches that too. Pinned by `test_a_delegate_that_strips_a_body_dunder_restores_it` (strips body `__eq__` and `__hash__`, asserts both behaviors; fails against the skip-if-present version).
</details>

<details>
<summary><b>F1 — the early accept is gone</b></summary>

The count-only accept path is deleted; the settle always installs the plan's defaults. A lying-around class matching the plan's name, fields, and default count with different values is corrected to the call's plan. Pinned by `test_a_delegate_returning_same_count_defaults_is_corrected_not_accepted` (a statement-built `Built` carrying `y: 99` handed back by the delegate; `built(1).y == 42`).
</details>

<details>
<summary><b>The systemic problem — answered by the refactor</b></summary>

`binding_plan` + `enum hash_binding` now own the conditions; `bind_not_equal` and `bind_hash` folded in. The remaining per-executor difference is the materialization: the namespace path lets `type_new` compute the slots, the settle path must update a live type's slots through `setattro` — that is not a second copy of a decision, it is the transport. The fresh path's output is unchanged (the full matrix and the in-file C tests stay green).
</details>

Also filed upstream: the systemic problems were invisible in the findings JSON (`systemic_problems: []` while the rendered comment carried the section) — [JPHutchins/code-review#156](https://github.com/JPHutchins/code-review/issues/156).

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to iterate this PR with the automated reviewer.

Round 3 answered in `c4a8819` — five fixed, one refuted by measurement.

<details>
<summary><b>F0 — one comparison table</b></summary>

Both paths now bind the same five names and handle `__ne__` through the plan's own arms (`answered_by_body`, `rebind_not_equal`); the six-name table and its two compensating settle arms are gone. The fresh path's output is unchanged (the whole matrix stays green).
</details>

<details>
<summary><b>F1 — injected dunders refused</b></summary>

`settled_by_the_plan` says which names the plan owns; a class-dict dunder that is neither the body's own definition nor a planned binding is refused before any settle step runs. Pinned by `test_a_delegate_that_injects_a_dunder_is_refused` (a delegate injecting `__eq__` with eq unchanged — the case where the option did not flip, exactly the finding's repro).
</details>

<details>
<summary><b>F2 — two-sided weakref, one invariant</b></summary>

The settle now checks `carries_slot == weakref_expected(options, bases)` — both directions. Note on the repro mechanics, measured: a bare `__weakref__` namespace key does **not** create a slot when `__slots__` is present (the key lands in the class dict, `tp_weaklistoffset` stays 0, `weakref.ref` raises), and an `__slots__`-based injection is rebuilt away by the re-entered build. The reachable case is a delegate returning a class built elsewhere with `weakref=True` — pinned by `test_a_delegate_returning_a_weakref_slotted_class_is_refused`.
</details>

<details>
<summary><b>F3 — state writes last</b></summary>

`struct_defaults`/`struct_options`/`struct_resolves_body_eq` are now written after every step that can fail, so an error leaves the class untouched.
</details>

<details>
<summary><b>F4 + F5</b></summary>

`options_agree` deleted with its last caller. The up-front gate and the settle share the `weakref_expected`/`weakref_slot_is_new` predicates; the gate only pre-empts the direction the re-entered build genuinely cannot carry, and keeps its message naming the hand-off (the settle's generic refusal covers everything else).
</details>

<details>
<summary><b>F6 — refuted</b></summary>

The settle runs once per metaclass call, on a class that is seconds old and private to the delegate's `__new__` — nothing has cached a lookup on it yet, so `PyType_Modified` invalidates nothing. The setattro is the mechanism that updates the slots (that was F1's fix); skipping "identical" wrappers would compare freshly-created slot wrappers, which are never pointer-identical, so the skip could only fire by running user code on the dict entries. The cost is one version-tag bump per exotic class creation.
</details>

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to iterate this PR with the automated reviewer.

Round 4 answered in `37f9e26` — both minors fixed, both nits answered.

<details>
<summary><b>F0 — the install_fields-time decisions re-applied</b></summary>

`__init__` and `__post_init__` joined the refusal/restore lists (a delegate-injected one refuses, a stripped body one restores), and the settle's tail now mirrors install_fields: the `tp_new`/`tp_vectorcall` branch plus `install_post_init`. The first gate run caught exactly the trap the branch must avoid — dispatch prefers `tp_vectorcall`, so the init case also clears it (`tp_vectorcall = NULL`), matching the fresh build. Pinned by the two strip tests: a stripped `__init__` routes through the body constructor (and a body `__init__` displaces `__post_init__` exactly as a fresh build behaves — the test asserts the displacement, not a run), and a stripped `__post_init__` runs.
</details>

<details>
<summary><b>F1 — the injected-`__ne__` hole closed</b></summary>

The answered arm now binds object's `__ne__` whenever the body did not define one — unconditionally, so a delegate-injected `__ne__` is overwritten rather than skipped. The refusal loop admits `__ne__` in the answered case for the same reason: the arm is what fixes it. Pinned by `test_a_delegate_that_injects_ne_when_the_body_answers_eq_is_overwritten` (an injected always-True `__ne__`; `!=` then answers object's, False).
</details>

<details>
<summary><b>F2 — refuted, with the mechanism</b></summary>

`create_class` calls `type_new` with no keywords, so the re-entered build's `options_read` always resolves to the inherited options — the new slot cannot be added by the re-entered build, which is exactly what the message says. A delegate that builds the class itself via its own full `META` call is the only cooperative route, and that requires the outer call's refusal to have already fired. The gate trades that exotic pattern for the clear early error; everything that reaches the settle is judged by the two-sided check. Keeping it.
</details>

<details>
<summary><b>F3 — done</b></summary>

The displaced-slots call site now uses `weakref_expected(request.options, bases)`; the predicate has no inline twin left.
</details>

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to iterate this PR with the automated reviewer.

Round 5 answered in `ab021b9` — all four findings fixed, and the gate caught a segfault of ours along the way.

<details>
<summary><b>F0 — body `__new__` reconciled</b></summary>

Measured first: the fresh build keeps a body `__new__` in the class dict but dispatch never consults it (a body `__new__` records zero calls — `tp_vectorcall` wins). The settle matches that exactly: `__new__` joined the restore/refuse lists, so a stripped body `__new__` is restored into the dict and an injected one refuses. No install-time decision to re-apply — the fresh path makes none for `__new__`.
</details>

<details>
<summary><b>F1 — the post_init over-hold closed</b></summary>

`install_post_init` now `Py_XSETREF`s the hook. The first gate run caught a real one here: the 3.14 `Py_SETREF` macro decrefs the old value with `Py_DECREF` (not X), and the module-init install's `struct_post_init` is still NULL — `Py_DECREF(NULL)` segfaulted on every version under the pytest environments (the import-only smoke test masked it; the matrix gate caught it). `Py_XSETREF` is the NULL-tolerant variant; the other two `Py_SETREF` sites are guarded non-NULL.
</details>

<details>
<summary><b>F2 — hashability pinned</b></summary>

`test_a_delegate_can_turn_frozen_off` now also asserts `hash(instance)` raises `TypeError: unhashable` — the settled class is unhashable by behavior, not just by the dict entry, on every version in the matrix (3.10–3.15 + 3.14t all pass, which is the empirical answer to the per-version claim).
</details>

<details>
<summary><b>F3 — the bases gate</b></summary>

The identity gate compares the whole bases tuple against the call's. Pinned by `test_a_delegate_that_adds_a_non_struct_base_is_refused` (a statement-built `Built(Base, InitBase)` handed back).
</details>

PR #103, opened 2026-08-13
docs: the ClassVar/InitVar escapes and the cached_property answer

### LINKED ISSUE #53: functools.cached_property cannot be used on a struct, and nothing says so

### LINKED ISSUE #57: The ClassVar/InitVar text path is a heuristic: the escapes and false refusals it is known to have

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to work the residue phase — this is the documentation half of it.

Two README sections, each the measured claim from its issue, neither depending on the #18 landing-page rewrite:

- **ClassVar and InitVar** — what the text path's spelling match costs: the table from the issue, as the terse version a reader can reach.
- **Caching a computed value** — `cached_property` refuses (no instance `__dict__`; a slotted frozen dataclass refuses identically), with the two answers that work, both running as doctests so they cannot rot.

The README stays executed: the new doctests pass under the same collection as the rest of the file.

Closes #53. Closes #57.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,10 +31,10 @@ the check is a spelling match, and the two paths differ at the edges:
 | `Optional[Annotated[ClassVar[int], "m"]]` | refused | refused |
 
 3.14 evaluates resolvable annotations by default, so there the alias reaches
-the object path and is refused; a name nothing can resolve — bare or
-compound — arrives as an object too, and is accepted, while an annotation
-whose evaluation raises AttributeError still fails the class. Every row is
-pinned in `tests/test_classvar_paths.py`.
+the object path and is refused; a name whose root cannot be resolved
+arrives as an object too, and is accepted, while any other evaluation
+error still fails the class. Every row is pinned in
+`tests/test_classvar_paths.py`.
 
 ## Caching a computed value
 
@@ -72,8 +72,9 @@ method, or a field that `__post_init__` fills:
 Under the default eq=True the method cache keys on the value, not the
 instance: value-equal structs share one entry, and a `set_field` that
 changes the value misses — the cache then holds the old entry and the new
-one. With eq=False the struct hashes by identity, and `set_field` after a
-hit returns the stale cached value while the hash is unchanged; a body
+one. With eq=False the struct hashes by identity — unless the body defines
+its own `__eq__`, which still makes it unhashable — and `set_field` after
+a hit returns the stale cached value while the hash is unchanged; a body
 `__hash__` over field values changes the hash and the lookup misses. The
 cache refuses an unhashable struct — anything that makes `hash()` raise:
 under the default options that is `frozen=False`, a body-defined `__eq__`,
--- a/README.md
+++ b/README.md
@@ -18,28 +18,31 @@ Point(x=1.0, y=2.0)
 ## ClassVar and InitVar
 
 A field annotated `ClassVar[...]` or `InitVar[...]` is refused at class
-creation, exactly when the annotation is an object. When the annotation
-reaches salix as source text (under `from __future__ import annotations`)
-the check is a spelling match, and the spellings it cannot see are:
+creation. The check is exact when the annotation reaches salix as an object;
+when it arrives as source text -- a quoted string, or an annotation that
+nothing evaluated -- the check is a spelling match, and the two paths differ
+at the edges:
 
-| annotation | what the spelling match does |
-| --- | --- |
-| an alias of `ClassVar`, `CV[int]` | accepted; the field swallows a positional |
-| a factory-local alias, 3.14 | accepted; the same symptom |
-| `Optional[Annotated[ClassVar[int], "m"]]` | the object path accepts, the text path refuses |
-| a type of yours actually named `ClassVar` | refused |
-| `Annotated[int, ClassVar]` | refused, though the form is metadata |
+| annotation | object path | text path |
+| --- | --- | --- |
+| an alias of `ClassVar`, `CV[int]` | refused | accepted; the field swallows a positional |
+| a type of yours actually named `ClassVar` | accepted | refused |
+| `Annotated[int, ClassVar]` | accepted | refused, though the form is metadata |
+| `Optional[Annotated[ClassVar[int], "m"]]` | refused | refused |
+
+3.14 evaluates resolvable annotations by default, so there the alias reaches
+the object path and is refused; the text path sees only what nothing can
+resolve.
 
 ## Caching a computed value
 
 `functools.cached_property` caches into an instance `__dict__`, and a struct
-has none; a slotted frozen dataclass refuses it the same way. Cache on the
-method instead -- the cache then holds every instance it has seen -- or fill
-a field from `__post_init__`:
+has none; a slotted frozen dataclass refuses it the same way. Two answers:
+`functools.cache` on the method, or a field that `__post_init__` fills:
 
 ```python
 >>> import functools
->>> from salix import set_field
+>>> from salix import Struct, set_field
 
 >>> class Computed(Struct):
 ...     x: int
@@ -62,6 +65,11 @@ a field from `__post_init__`:
 8
 
 ```
+
+The method cache holds every instance it has seen -- one entry per mutation
+state, since structs hash by value -- and refuses when the instance is
+unhashable. A body-defined `__init__` replaces the constructor that runs
+`__post_init__`, so the field answer only works without one.
 
 `salix/__init__.pyi` is the API. `src/salix.c` says what this is and is
 not. `tasks.py` is what runs. `uv run camas benchmark` says what it costs.
--- a/README.md
+++ b/README.md
@@ -73,15 +73,16 @@ Under the default eq=True the method cache keys on the value, not the
 instance: value-equal structs share one entry, and a `set_field` that
 changes the value misses — the cache then holds the old entry and the new
 one. With eq=False the struct hashes by identity — unless the body defines
-its own `__eq__`, which still makes it unhashable — and `set_field` after
+its own `__eq__` without a `__hash__`, which makes it unhashable, or its
+own `__hash__`, which replaces the identity hash — and `set_field` after
 a hit returns the stale cached value while the hash is unchanged; a body
 `__hash__` over field values changes the hash and the lookup misses. The
 cache refuses an unhashable struct — anything that makes `hash()` raise:
-under the default options that is `frozen=False`, a body-defined `__eq__`,
-a body `__hash__` that refuses, an unhashable field value, or equality
-inherited from a struct base that defines `__eq__`. An `__init__` that is
-not `object.__init__` — a body-written one, inherited or not — replaces
-the constructor that runs `__post_init__`, so the field answer only works
+a body `__eq__` without a body `__hash__`, a body `__hash__` that
+refuses, `frozen=False`, an unhashable field value, or equality inherited
+from a struct base that defines `__eq__`. An `__init__` that is not
+`object.__init__` — a body-written one, inherited or not — replaces the
+constructor that runs `__post_init__`, so the field answer only works
 without one. The table and the caching behaviors above are pinned in
 `tests/test_classvar_paths.py`.
 
--- a/README.md
+++ b/README.md
@@ -27,12 +27,13 @@ the check is a spelling match, and the two paths differ at the edges:
 | --- | --- | --- |
 | an alias of `ClassVar`, `CV[int]` | refused | accepted; the field swallows a positional |
 | a type of yours actually named `ClassVar` | accepted | refused |
-| `Annotated[int, ClassVar]` | accepted, 3.11+ | refused, though the `ClassVar` there is metadata |
+| `Annotated[int, ClassVar]` | accepted | refused, though the `ClassVar` there is metadata |
 | `Optional[Annotated[ClassVar[int], "m"]]` | refused | refused |
 
 3.14 evaluates resolvable annotations by default, so there the alias reaches
-the object path and is refused; an annotation nothing can resolve arrives as
-an object too, and is accepted. Every row is pinned in
+the object path and is refused; a bare name nothing can resolve arrives as an
+object too, and is accepted, while an annotation whose evaluation fails any
+other way still fails the class. Every row is pinned in
 `tests/test_classvar_paths.py`.
 
 ## Caching a computed value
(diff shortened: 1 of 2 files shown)


JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to iterate this PR with the automated reviewer.

Round 1 answered in `bbbf45a` — the table was wrong, and it is now re-measured rather than edited.

**Executive summary:** the first draft transplanted the issue's table, which recorded the object path from before the `__args__` walk shipped. Every row has now been measured on the built extension (3.14.6 object path with evaluated annotations; 3.13.7 with string annotations standing in for what the compiler stores under future-annotations) and the table rewritten from those results.

<details>
<summary><b>Measured matrix</b></summary>

| annotation | object path | text path |
| --- | --- | --- |
| an alias of `ClassVar`, `CV[int]` | refused | accepted; the field swallows a positional |
| a type of yours actually named `ClassVar` | accepted | refused |
| `Annotated[int, ClassVar]` | accepted | refused, though the form is metadata |
| `Optional[Annotated[ClassVar[int], "m"]]` | refused | refused |

F0 confirmed: the `Optional[Annotated[...]]` row was pre-fix; both paths refuse it now. F3 confirmed: rows 4-5 are text-path-only over-refusals, and the intro now says the two paths differ at the edges instead of mislabeling them. F1: the 3.14 sentence states the measured default-mode behavior (a resolvable alias reaches the object path there and is refused — measured with a real class statement on 3.14.6).
</details>

<details>
<summary><b>F2 + F4 + F5 — the caching caveats, all measured</b></summary>

- Body `__init__` displaces the hook: measured, `double` stays `0` with a body `__init__` present — the section now says so.
- `functools.cache` on a struct holding a list: `TypeError: unhashable type: 'list'` — the section says the cache refuses when the instance is unhashable, and that value hashing accumulates one entry per mutation state.
- The snippet now imports `Struct` alongside `set_field`, so it copies standalone.
</details>

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to iterate this PR with the automated reviewer.

Round 2 answered in `9725215` — all ten findings fixed by measurement, and both tables are now pinned in `tests/test_classvar_paths.py` so the prose is run rather than proof-read, per the repo's own doctrine.

<details>
<summary><b>What changed, and what was measured</b></summary>

- **F0**: measured on 3.14 — an unresolvable annotation arrives as a ForwardRef *object* and is accepted on the object path; the sentence says so (pinned with a `skipif < 3.14`).
- **F1**: em dashes throughout.
- **F2**: every row of both tables is now a test in `tests/test_classvar_paths.py` — object-path rows via evaluated annotations, text-path rows via strings, so each runs on every interpreter in the matrix. The README points at the file.
- **F3**: "refused, though the `ClassVar` there is metadata".
- **F4**: measured — an *inherited* `__init__` displaces the hook too (`defines_own_init` keys on `tp_init != object's`); the sentence says "defined or inherited" and both cases are pinned.
- **F5**: the unhashable triggers are named: `frozen=False` or a body-defined `__eq__` (measured — a mutable struct raises `TypeError: unhashable` under `functools.cache` outright).
- **F6+F8**: re-measured with a frozen struct mutated via `set_field`: value-equal instances share one entry (one call recorded), a post-hit mutation misses and the cache holds both entries (`currsize == 2`). The sentence now states keying-on-value, sharing, and the leak — not "holds every instance".
- **F7**: measured — `cached_property` works on a struct with a dict-bearing co-base (caches into the dict); the sentence qualifies "unless a non-struct base carries one".
- **F9**: "walks the annotation's forms" instead of "exact".
</details>

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to iterate this PR with the automated reviewer.

Round 3 answered in `91459eb` — all six findings fixed.

<details>
<summary><b>What changed</b></summary>

- **F0**: the object-path `Annotated[int, ClassVar]` test carries the twin's `< 3.11` skipif (typing refuses a bare special form as Annotated metadata before 3.11).
- **F1**: the caching caveats are scoped to eq=True, and the eq=False behavior is stated and pinned: identity hashing (measured — `set_field` after a hit returns the stale cached value, a cache *hit*), and eq=False mutable structs are hashable.
- **F2**: the 3.14 test now asserts the routing itself — the annotation comes back from `annotationlib.get_annotations(..., Format.FORWARDREF)` as a non-str ForwardRef object (measured).
- **F3**: the refusal rows are parametrized and pin the message with `pytest.raises(match="ClassVar")` (measured: all three refusal shapes emit the same "is annotated ClassVar, which salix does not support" message).
- **F4**: the unpinned cells are now pinned — the alias field swallowing a positional (the constructed field holds the positional), and a body-defined `__eq__` making the cache refuse (hash becomes None under bind_hash's body-eq rule).
- **F5**: parametrize idiom throughout the table rows.
- **F6**: the opening is scoped — refused when the annotation reaches the check as one of those forms, so the 3.14 ForwardRef sentence no longer contradicts it.
</details>

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to iterate this PR with the automated reviewer.

Round 4 answered in `e696953` — F0, F1, F3, F5, F6 fixed; F2 and F4 refuted.

<details>
<summary><b>Fixed</b></summary>

- **F0**: the eq=False sentence now carries the body-`__eq__` exception (measured: `bind_hash` answers a body-defined `__eq__` with `__hash__ = None` regardless of eq), and the unhashable triggers name the unhashable-field-value case; all three pinned.
- **F1**: the dataclass clause drops "frozen" — the refusal is slots alone.
- **F3**: the identity pins assert distinct hashes for distinct instances — `object.__hash__` scrambles the pointer, so `hash != id` numerically (measured; the first pin attempt asserted the numeric equality and the gate caught it).
- **F5**: the Annotated row says "accepted, 3.11+", matching its pin's guard.
- **F6**: the `__init__` sentence is scoped to "an `__init__` that is not `object.__init__`", so a constructor inherited from another struct no longer contradicts it.
</details>

<details>
<summary><b>Refuted</b></summary>

- **F2**: the file deliberately re-pins the README's tables in one place — the README points at it, so a reader follows one pointer; the near-copies exist so each cell fails independently with its own assertion.
- **F4**: salix's annotation reader uses the same `annotationlib` FORWARDREF access the test calls (the #44 machinery), and the acceptance is asserted alongside — a regression to the text path changes the walk's input and the field outcome the test asserts.
</details>

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to iterate this PR with the automated reviewer.

Round 5 answered in `a69affb` — both minors and five of six nits fixed; F6 answered honestly.

<details>
<summary><b>Fixed</b></summary>

- **F0**: the identity-hash pins hold the instances; no more freed temporaries.
- **F1**: the trigger list became the rule — anything that makes `hash()` raise — with the default-option causes named; the omitted triggers are pinned: body `__hash__ = None` under both eq options, a body `__hash__` winning over the identity hash under eq=False (measured: `hash() == 7`), and equality inherited from a struct base. The eq=False sentence is scoped against a body `__eq__` *or* `__hash__` of any kind.
- **F2**: measured on 3.10.20 — `Annotated[int, ClassVar]` constructs and is accepted, so the skipif and the 3.11+ qualifier are gone; the row is unconditional and the pin runs everywhere.
- **F3**: scoped to "a `set_field` that changes the value misses".
- **F4**: scoped to a bare unresolvable name; other evaluation failures still fail the class.
- **F5**: "inherited from any base" — a struct base's body `__init__` displaces the hook too (measured).
- **F7**: the closing sentence now names exactly what is pinned.
</details>

<details>
<summary><b>F6 — answered</b></summary>

The routing pin asserts the annotation salix's own accessor returns is a non-str ForwardRef, plus the acceptance. A salix regression that switches to reading strings would still *accept* the same annotation, so no black-box test can distinguish the two paths when their outcomes coincide — the pin covers what is observable, and the sentence no longer claims more. The distinct-outcome rows (alias, metadata) do discriminate, and those are pinned per-row.
</details>

Also on the loop: #102 converged at round 6 (score 0) and merged as `8106b09`.

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to iterate this PR with the automated reviewer.

Round 6 answered in `1bac7ae` — three wording fixes, one refutation.

<details>
<summary><b>Fixed</b></summary>

- **F0**: the trigger list says "a body `__hash__` that refuses" — a value-returning body `__hash__` is respected and hashable under both eq options (pinned: `hash(ValueHash(3)) == 3` under the defaults, and the existing eq=False `SevenHash` pin).
- **F1**: the eq=False clause now ties the stale hit to the hash being unchanged: a body `__hash__` over field values changes the hash and the lookup misses.
- **F3**: scoped to "a body-written one, inherited or not" — a struct base's *generated* constructor is `object.__init__` and does not displace (the very case the earlier scoping had covered; the round-5 wording re-broadened it).
- **F4**: the 3.14 sentence covers bare and compound unresolvable names — both measured accepted (`x: Nope` and `x: Nope.attr`) — and names AttributeError as the failing case, measured (`AttributeError: type object 'Nope' has no attribute 'missing'`).
</details>

<details>
<summary><b>F2 — refuted by measurement</b></summary>

In fresh processes on both the main venv and the pytest environment, `class WithNope(Struct): x: Nope` is accepted with the field installed. No order dependence reproduced; the pin runs first-or-last identically.
</details>

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to iterate this PR with the automated reviewer.

Round 7 answered in `dca546f` — all four findings fixed.

<details>
<summary><b>What changed</b></summary>

- **F0**: the eq=False sentence restores the body-`__eq__` exception the round-6 edit had dropped.
- **F1**: both claims now have pins — `test_a_field_hash_misses_after_set_field_under_eq_false` (a body `__hash__` over field values: `set_field` changes the hash and the cache recomputes, 10) and `test_a_struct_base_body_init_displaces_post_init` (a body-written `__init__` inherited from a struct base leaves `double == 0`).
- **F2**: the alias's object-path cell is pinned with its own parametrize id ("an-alias-evaluates-to-the-form" — the alias spelling evaluates to the form the object path refuses); the compound-name and AttributeError halves of the 3.14 sentence are pinned on the 3.14 gate (`x: Nope.attr` accepted; `x: Nope.missing` fails with AttributeError, measured).
- **F3**: the sentence now partitions on the root: a name whose *root* cannot be resolved arrives as an object and is accepted; any other evaluation error still fails the class.
</details>

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to iterate this PR with the automated reviewer.

Round 8 answered in `00233c1` — F0 refuted, all six nits fixed.

<details>
<summary><b>F0 — refuted by measurement</b></summary>

On 3.10.20 and 3.11.14, `typing.Annotated[int, typing.ClassVar]` keeps the form in `__metadata__`, not `__args__` (measured: `__args__ == (int,)`, `__metadata__ == (typing.ClassVar,)`), and the object path accepts it — identical to 3.12+. The pin runs without a skipif on the whole matrix and is green on 3.10 and 3.11. No version-dependent walk behavior reproduced.
</details>

<details>
<summary><b>Nits fixed</b></summary>

- **F1**: the eq=False labels are qualified — a body `__eq__` *without* a body `__hash__` makes the struct unhashable; a body `__hash__` replaces the identity hash.
- **F2**: the alias row now constructs a real alias inside the test and pins the refusal.
- **F3**: the no-instance-`__dict__` negative is pinned (`Plain(1).__dict__` raises AttributeError).
- **F4**: the trigger list is illustrative over the rule ("anything that makes `hash()` raise") — no exhaustiveness claim.
- **F5**: the "under the default options" framing is gone; the triggers are bind_hash's absolute rule (`frozen=False` among them, as a requested option).
- **F6**: the 3.14 composition is pinned with a real class statement — a resolvable `typing.ClassVar[int]` annotation is evaluated and refused on the 3.14 gate.
</details>

PR #109, opened 2026-08-14
fix: bind what the MRO resolves — multi-base structs behave as their record

### LINKED ISSUE #76: Options are recorded from one base and answered by the MRO, so a multi-base struct can record one thing and behave as another

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was to resolve issue #76 per the design recorded in the batch notes: bind what the MRO resolves, gate on more than one struct base, flip the xfails.

Closes #76.

**What:** a multi-base struct records its options from the first struct base while the dunders the MRO answers can come from any base. The fix asks the real MRO after the class exists and binds what the option asks for.

<details>

**The mechanism, in two parts.** A struct base binds a dunder into its own dict only where its creation transitioned an option, so a later base's binding sits ahead of the shared mixin in the MRO. And CPython copies `tp_richcompare`/`tp_hash` from the first base only when the new class's namespace overrides nothing — a multi-base class whose namespace carries salix's own bindings falls back to the default slots: identity equality, no ordering, id hash, whatever the record says.

**The repair** — `settle_mro_bindings`, run after either build path finishes, gated on more than one struct base so the single-base path is untouched:

- `tp_richcompare` becomes the structural slot when eq is recorded and neither this class's body nor its first base's answers equality itself; `tp_hash` becomes the structural hash where the plan binds it.
- `__eq__`/`__ne__`/`__repr__` are rebound into the class dict when the real MRO hands out a different binding than the option asks for.
- A class whose own body answers equality, or whose first base's does, is left to answer it — the record *is* the body equality there.

**The pins.** Six xfails flip: the four behaviour buckets (`repr`, `equality`, `ordering`, `hash` — the sweep asserts the class behaves as the first base's record), and the two walk-limit shapes in `test_non_struct_bases.py`, rewritten to pin the record winning over a later base's body `__eq__` and over a co-base between two struct bases (structural pair, hash included). The two broken-state bounds the fix empties are deleted with them, per their own docstrings. The frozen shape the issue lists is already held by `frozen_across_bases` — measured on 3.14, writes are refused, no pin needed.

**Verification** — the suite went from red on exactly the six flipped pins to green: 1148 passing on 3.14, camas gate green on the three changed paths.

</details>

### DIFF
--- a/src/meta/class.c
+++ b/src/meta/class.c
@@ -648,23 +648,31 @@ static enum result settle_mro_bindings(
 	}
 
 	PyTypeObject * const type = (PyTypeObject *) struct_class;
-	bool const body_defines_eq = PyDict_GetItemString(original_namespace, "__eq__") != NULL;
-	bool const answers_itself = body_defines_eq || inherits_body_eq;
+	bool const answers_itself = bindings.answered_by_body || inherits_body_eq;
 
-	if (options.eq && !answers_itself && type->tp_richcompare != Struct_rich_compare) {
+	if (!answers_itself && options.eq && type->tp_richcompare != Struct_rich_compare) {
 		type->tp_richcompare = Struct_rich_compare;
 	}
 
-	if (bindings.hash == HASH_BIND && type->tp_hash != Struct_hash) {
+	if (options.eq && bindings.hash == HASH_BIND && type->tp_hash != Struct_hash) {
 		type->tp_hash = Struct_hash;
 	}
 
 	PY_OWNED(mixin_eq, PyObject_GetAttrString((PyObject *) &StructMixin_Type, "__eq__"));
 	PY_OWNED(object_eq, PyObject_GetAttrString((PyObject *) &PyBaseObject_Type, "__eq__"));
+	PY_OWNED(mixin_ne, PyObject_GetAttrString((PyObject *) &StructMixin_Type, "__ne__"));
+	PY_OWNED(object_ne, PyObject_GetAttrString((PyObject *) &PyBaseObject_Type, "__ne__"));
 	PY_OWNED(mixin_repr, PyObject_GetAttrString((PyObject *) &StructMixin_Type, "__repr__"));
 	PY_OWNED(object_repr, PyObject_GetAttrString((PyObject *) &PyBaseObject_Type, "__repr__"));
 
-	if (mixin_eq == NULL || object_eq == NULL || mixin_repr == NULL || object_repr == NULL) {
+	if (
+		mixin_eq == NULL ||
+		object_eq == NULL ||
+		mixin_ne == NULL ||
+		object_ne == NULL ||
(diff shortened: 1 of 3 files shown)


JPHutchins (commented):
minor proliferation of doc slop in the source code

Issue #60, opened 2026-08-01
A zero-field frozen struct has one possible value but allocates a new object per call: 93 ns where an interned singleton is 6 ns


JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by `claude-opus-5[1m]` on behalf of @JPHutchins. She asked me to look at the scaffold of [argtree](https://github.com/JPHutchins/argtree) — a type-tree CLI argument parser — and work out what salix would have to implement in C for that parser to approach zero startup cost. This is one of five findings.

## Summary

A frozen struct with zero fields has exactly one possible value. Salix allocates a fresh object for it on every call anyway: **93.1 ns**, against **6.2 ns** to load an already-bound reference. `Empty()` is tested (`tests/test_struct.py`) but not special-cased.

**Ask:** when a struct type is created with `frozen=True` and zero fields, build the single instance once and have `tp_new`/vectorcall return `Py_NewRef` of it.

Two things fall out, and the second matters more than the speed:

1. Construction becomes an incref — a ~15× improvement on a call that is in a parser's per-token path.
2. `ToInt() is ToInt()` becomes **true**, so a consumer can dispatch a closed sum by identity instead of by `isinstance`, and can put variants in a `set` or use them as dict keys with meaningful semantics.

## Why a decoder hits this constantly

A closed sum type in the functional style is mostly nullary variants. argtree's grammar has **11 zero-field variant classes** out of 41 struct types:

```python
# src/argtree/_native/grammar.py
class ToInt(NamedTuple): ...      # Convert = ToInt | ToFloat | ToPath | ToStr | ToEnum | ToLiteral | Custom
class ToFloat(NamedTuple): ...
class ToPath(NamedTuple): ...
class ToStr(NamedTuple): ...
class One(NamedTuple): ...        # Arity = One | Exactly | Star | Plus
class Star(NamedTuple): ...
class Plus(NamedTuple): ...
class Required(NamedTuple): ...   # Default = Required | HasValue | NoneDefault
class NoneDefault(NamedTuple): ...
class Count(NamedTuple): ...      # Slot = Flag | Count | Opt | Many | Pos | PosMany
# src/argtree/_native/tokens.py
class EndOpts(NamedTuple): ...    # Token = Long | Short | Value | EndOpts
```

[...]

PR #137, opened 2026-08-21
feat: __replace__ and salix.replace, the copy-with-override protocol

### LINKED ISSUE #61: There is no copy-with-override, so an immutable cursor must restate every field it is not changing
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by `claude-opus-5[1m]` on behalf of @JPHutchins. She asked me to look at the scaffold of [argtree](https://github.com/JPHutchins/argtree) — a type-tree CLI argument parser — and work out what salix would have to implement in C for that parser to approach zero startup cost. This is one of five findings.

## Summary

There is no copy-with-override on a struct. `NamedTuple` has `_replace`, `dataclasses` has `replace`, and since 3.13 the standard spelling is `copy.replace()` dispatching to `__replace__`. Salix has none of them, so the only way to advance an immutable cursor by one field is to name every other field at the call site — which is exactly the code that a frozen record type exists to stop you writing.

**Ask:** implement `__replace__` in C, so `copy.replace(cursor, pos=n)` works, plus a bound method so a consumer does not have to `import copy` to use it.

The C implementation is a straight win over `_replace`, which is not a small overhead: **321.8 ns against 116.0 ns** to build the same object positionally. `_replace` routes through `_make(map(kwds.pop, self._fields, self))` and a leftover-keys check; a C version is an allocation plus N increfs with the overridden slots swapped, which should land at or below the plain constructor.

## Measurements

CPython 3.14.0 (Clang 20.1.4), Linux x86_64 (WSL2). `timeit`, min of 7 autoranged repeats, two-field type.

| | ns/call |
| --- | --- |
| `t._replace(a=2)` | **321.8** |
| `Two(2, t.b)` — the hand-written equivalent | 126.9 |
| `Two(1, 'y')` — plain positional construction, the floor | 116.0 |

In situ, argtree's cursor:

| | µs/call |
| --- | --- |
| `Cursor.bump()` → `self._replace(pos=self.pos + 1)` | **0.37** |
| `Cursor(toks, 1, False)` — direct construction | 0.12 |

`bump()` is called once per consumed token, and once more per lookahead in `gather`. For a parser this is the single hottest constructor.

## The idiom this is for

An immutable cursor over the token stream is the core of a total, pure decoder — no mutable index, no backtracking state, every position a value:

```python
# src/argtree/_native/tokens.py
class Cursor(NamedTuple):
	"""Immutable index-cursor over the token stream; ``literal`` is set once past ``--``."""

	toks: tuple[Token, ...]
	pos: int
	literal: bool

	def peek(self) -> Token | None:
		return self.toks[self.pos] if self.pos < len(self.toks) else None

	def bump(self) -> Cursor:
		return self._replace(pos=self.pos + 1)

	def go_literal(self) -> Cursor:
		return self._replace(pos=self.pos + 1, literal=True)
```

Written against salix as it stands today, `bump` becomes `Cursor(self.toks, self.pos + 1, self.literal)` and `go_literal` becomes `Cursor(self.toks, self.pos + 1, True)`. That works, and it is even faster than `_replace`. But it re-states the field list at every call site, so adding a fourth field to `Cursor` silently breaks nothing and loudly breaks everything, and the whole point of a struct is that it owns its own field list.

## Notes on scope

- `copy.replace()` (3.13+) is the standard protocol: `object.__replace__(self, /, **changes)`. Implementing that dunder is the portable half.
- A bound method is the other half. `copy` is not a free import (it pulls in `types`, `weakref`, and its dispatch tables), and a library whose pitch is *"for programs that pay for their imports"* should not make its users import `copy` to advance a cursor. `msgspec` ships `msgspec.structs.replace`; `attrs` ships `evolve`.
- Naming: `NamedTuple` uses the underscore prefix (`_replace`) to keep the field namespace clear, and salix has the same collision problem — a field named `replace` would shadow the method. The existing `__struct_*` convention and the module-level `salix.set_field(instance, name, value)` both suggest salix's answer is a module-level function rather than a method, i.e. `salix.replace(instance, **changes)`. That sidesteps the collision entirely and matches the shape already established by `set_field`.
- Interaction with #24 (frozen writes raise `TypeError`, not `AttributeError`) and #2 (`object.__setattr__` on a frozen struct): a proper `__replace__` is the sanctioned way to derive a value from a frozen struct, and would take pressure off both of those escape hatches.
- Interaction with #56 (a body `__init__` disables defaults) and #12 (a body `__new__` is discarded): `__replace__` must construct through whatever the class's real construction path is, or it will diverge from `__init__` in exactly the cases those issues are about. Worth deciding deliberately whether `replace` re-runs `__post_init__`; `dataclasses.replace` does, `msgspec.structs.replace` does not.

---

Part of #64, which has the full startup budget and the method behind these numbers.

JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked me to work through issue #61 (copy-with-override) and drive the PR through the review loop.

## Summary

Closes #61. `__replace__(self, /, **changes)` on the mixin (the 3.13+ `copy.replace` protocol dunder) plus `salix.replace(instance, **changes)` as the module-level spelling, per the issue's collision reasoning — no bare `replace` method, since a field could shadow it. The C implementation is ~6x faster than `NamedTuple._replace` on the same machine.

## Design decisions

<details>
<summary>Two construction paths, dispatched exactly as the constructor dispatches</summary>

- **No body `__init__`** — the fast path: allocate, bind the changes through the constructor's own `bind_keywords` (an unknown change raises the constructor's exact `TypeError`), copy the untouched slots from the source under one critical section (`struct_slots_copy_into` gained a destination-skip so the changes survive the copy), then run `__post_init__` the way the vectorcall would.
- **Body `__init__`** — the real construction path: field values + changes become one kwargs dict, and the class is called with it. Replace cannot diverge from `__init__` (the #56/#12 interaction the issue called out).
- Zero changes returns the instance itself: replace is "the value with these changes"; no changes, the value is the instance. `copy.replace` refuses empty changes before the dunder is ever called, so this only affects direct dunder calls. Consistent with the singleton interning from #60.
</details>

<details>
<summary>__post_init__ re-runs</summary>

dataclasses semantics, and forced by construction: the fast path inlines exactly the vectorcall's sequence (bind, fill, post_init), so skipping it would be the divergent choice. Pinned by `test_post_init_runs_again_with_the_changed_value` (`calls == [1, 2]`).
</details>

<details>
<summary>The METH_FASTCALL convention cost a segfault round</summary>

For tp_methods descriptors, CPython passes self as the first parameter and nargs excludes it — `Objects/descrobject.c: meth(args[0], args+1, nargs-1, kwnames)`, verified in the 3.10/3.11/3.12/3.14 sources, all identical. `object.__replace__` does not exist in CPython; the dunder is a pure convention, so the error messages are salix's own. The first local build read change values at `arguments[1 + i]` — one past the keyword array — storing a garbage pointer into the kwargs dict and corrupting the heap; caught by the refcount probes as failures + segfaults across 3.10–3.15.
</details>

<details>
<summary>Collision and override</summary>

A field named `__replace__` is refused by the existing mixin-method scan (`refuse_mixin_method_fields` walks the mixin dict dynamically — no new machinery). A body `__replace__` overrides the mixin's, and `salix.replace` routes through `getattr(instance, "__replace__")`, so the override is honored there too (pinned).
</details>

## Measurements

Same machine, CPython 3.14.6, min of timeit repeats, the issue's two-field shape:

| | ns/call |
| --- | --- |
| `t._replace(a=2)` (NamedTuple) | 329.9 |
| `Two(2, t.b)` — hand-written | 40.4 |
| `t.__replace__(a=2)` | **54.8** |
| `salix.replace(t, a=2)` | 201.1 |

The dunder is 1.36x the hand-written equivalent (the issue's floor was 116 ns on an earlier build); the module function pays a `getattr` + call for the ergonomic spelling.

## Tests

`tests/test_replace.py` (15 behavioral pins: fast path, module function, unknown-change TypeError, zero-change identity, positional refusal, post_init re-run, body-init routing, override honored, mutable, singleton, dict copied shallowly, non-struct refusal, impostor refusal, `copy.replace` on 3.13+), two refcount probes, one free-threading pin (kept slots are a one-section snapshot), typing pins in `accepted.py`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

## Round 2 (after the reviewer's round 1, score 2.04)

<details>
<summary>The systemic answer: one state-carry primitive, one deliberate divergence</summary>

The four unintended divergences collapsed structurally. The slow path now snapshots the kept fields under **one critical section** for the kwargs (same guarantee the fast path gets), calls the constructor, then runs the **same `struct_copy_slots_and_dict`** the fast path uses — so whatever the constructor left unwritten (fields, co-base members, the instance dict) takes the source's values. The only remaining divergence between the paths is the deliberate one: construction routing. The shallow trio (`copy`, fast replace, slow replace) now shares the one helper; the member loop gained the destination-held skip the field loop already had (the nit's half of it).
</details>

<details>
<summary>Body `__new__`: answered by measurement, not code</summary>

Salix discards a body `__new__` at class creation on **both** construction paths — that is #12's resolution (`install_constructor` leaves `tp_new` NULL on the vectorcall path and sets `Struct_new` on the own-init path, so the body's `__new__` is never in the slot). Neither replace path can therefore invoke one, because replace shares the constructor's slots exactly. Pinned: `test_a_body_new_is_discarded_by_replace_as_by_construction` asserts `calls == []` across construction *and* replace. (The reviewer's claim that the slow path invokes it would require `tp_new` to hold the body's `__new__`, which it measurably does not.)
</details>

<details>
<summary>Keyword-only slow path: declared constraint</summary>

The slow path passes every field as a keyword, so a positional-only `__init__` raises its own `TypeError` — pinned (`test_a_positional_only_init_propagates_its_own_type_error`). `dataclasses.replace` has the same constraint; it is inherent to "route through `__init__`".
</details>

<details>
<summary>The stub's return type: declared</summary>

The pyi's `_StructT` promise describes the mixin's implementation. An overriding `__replace__` is user code, and its annotation is the user's — the module function honors the override (`test_a_body_replace_override_is_honored_by_the_module_function`), which is the only case that can return anything else.
</details>

<details>
<summary>Shadowing: now copy.replace's own lookup</summary>

`salix.replace` resolves `__replace__` on the **type**, exactly as `copy.py` does (`getattr(cls, '__replace__', None)`), so an instance-dict entry or `__getattr__` hook cannot shadow the dunder — pinned (`test_an_instance_dict_entry_does_not_shadow_the_dunder`).
</details>

---

## Round 3 (after the reviewer's round 2, score 1.84)

<details>
<summary>The systemic answer: one snapshot serves both phases</summary>

The slow path now assembles the kwargs from one section-spanning snapshot — the snapshot helper gained a `missing` filler parameter (Py_None for hash, NULL for replace), so the post-construction back-fill reuses the **same acquisition** instead of re-reading the source. The back-fill covers the constructor-left slots from that snapshot; members and the dict come through the shared shallow primitive. The instance dict is now **merged**, not replaced: entries the constructor itself wrote win over the source's copy — a `__init__` that computes fresh dict extras keeps them, and the source's extras still carry. The fast path runs `__post_init__` before the dict install (construction's own ordering), with the merge preserving whatever `__post_init__` wrote — pinned by `test_post_init_runs_before_the_source_dict_is_installed`.
</details>

<details>
<summary>Unknown changes are refused on both paths</summary>

The slow path validates every change name through the same `named_field` lookup the vectorcall uses (split out of `bind_keywords`, one copy of the error string), so a swallowing `**kwargs`-`__init__` can no longer silently absorb an unknown change — pinned.
</details>

<details>
<summary>The constructor call's result is type-checked</summary>

`PyObject_Call`'s result is checked against the class, raising the same `SystemError` shape `ensure_singleton` uses, so a rogue metaclass `__call__` returning a foreign struct cannot receive slot writes — pinned with a mutual-swap rogue metaclass.
</details>

<details>
<summary>Zero-change replace on a mutable struct is a real copy</summary>

Identity remains for frozen structs (indistinguishable values); a mutable zero-change replace now goes through the normal paths — fresh instance, `__post_init__` re-runs — pinned. (dataclasses parity.)
</details>

<details>
<summary>Claim checked by measurement, not taken on faith: `__init__ = None`</summary>

The round-2 finding said a body `__init__ = None` misroutes replace and silently drops changes. Measured: the class fails **construction** first — the `tp_init` slot wraps the `None`, so `NoneInit(3)` raises `'NoneType' object is not callable` and replace can never receive an instance. Both paths raise identically; pinned. (The premise of NULL `tp_init` does not hold; `defines_own_init` is untouched.)
</details>

<details>
<summary>Declared constraints, widened</summary>

- The slow path requires `__init__` to accept the **full field list as keywords** — widened from round 2's positional-only wording; a subset-signature `__init__` raises its own TypeError, pinned. `dataclasses.replace` shares the constraint.
- Constructor-ignored **defaulted** fields take their defaults (the constructor's own semantic; the pre-fill is indistinguishable from an `__init__` write). dataclasses parity.
- The module function's one-element tuple beats a `PyObject_Vectorcall` conversion (the kwargs dict would need a kwnames tuple built either way; the tuple-pack is the cheaper half — measured).
</details>

---

## Merged (round 4 converged at score 0.47) — recorded dispositions

The three carried findings and the nits below are recorded, not chased — the head was converged and certified, and each is a boundary the design already declares or a micro-optimization better folded into a follow-up:

- `replace-slow-path-null-slot-drop`: NULL source slots are skipped in the kwargs, so a `__init__` that declares a required parameter the instance never holds raises its own TypeError. Passing `Py_None` would be the alternative; the current behavior keeps NULL slots "unwritten stays unwritten" on both paths.
- `replace-co-base-shadowing`: a non-struct co-base's `__replace__` is shadowed by the mixin's, unlike `__copy__`/`__deepcopy__` which defer. The copy dunders are copy.py protocol entry points that copy.py itself resolves; `__replace__` is salix's own mixin contract. Fold the deferral into a follow-up if a co-base scenario demands it.
- `replace-slow-path-two-phase-read`: the T1 snapshot covers the field slots (constructor-honored fields, the advertised guarantee, FT-pinned); co-base members and the dict pointer are read at T2 under the shared primitive's own section. Boundary accepted as documented.
- `replace-slow-path-dict-assembly-redundant` (streak 3): the Contains check is live (it skips the kept-insert for changed names; the change loop then inserts each once) — the cost is one hash lookup per field on the slow path, which is a constructor call. Accepted.
- `replace-module-tuple-pack` (streak 3): `PyObject_Call` builds the kwnames tuple regardless, so the one-element args tuple is not the whole story; the tuple-pack remains the cheaper half of the conversion measured. Accepted.
- `empty-tuple-helper-unused`, `replace-slow-path-dict-presize`, `test-fixture-duplication-drift`, `replace-metaclass-subclass-result-rejected`: recorded for a follow-up pass (the exact-type guard rejects a metaclass `__call__` returning a subclass instance; the foreign-sibling case the guard exists for is pinned).

### DIFF
--- a/salix/__init__.pyi
+++ b/salix/__init__.pyi
@@ -1,4 +1,4 @@
-from typing import Any, Final
+from typing import Any, Final, TypeVar
 
 from typing_extensions import Self, dataclass_transform
 
@@ -14,6 +14,7 @@ class Struct:
     __struct_metadata__: Final[tuple[tuple[Any, ...], ...]]
     def __copy__(self) -> Self: ...
     def __deepcopy__(self, memo: dict[int, Any]) -> Self: ...
+    def __replace__(self, /, **changes: object) -> Self: ...
     def __init_subclass__(
         cls,
         *,
@@ -25,4 +26,8 @@ class Struct:
         weakref: bool = False,
     ) -> None: ...
 
+_StructT = TypeVar("_StructT", bound=Struct)
+
+
 def set_field(instance: Struct, name: str, value: object, /) -> None: ...
+def replace(instance: _StructT, /, **changes: object) -> _StructT: ...
--- a/src/construct.h
+++ b/src/construct.h
@@ -12,6 +12,13 @@ PyObject * Struct_vectorcall(
 
 PyObject * Struct_new(PyTypeObject * struct_class, PyObject * arguments, PyObject * keywords);
 
+PyObject * Struct_replace(
+	PyObject * self,
+	PyObject * const * arguments,
+	Py_ssize_t nargs,
+	PyObject * keyword_names
+);
+
 PyObject * Struct_set_field(PyObject * module, PyObject * arguments);
 
 bool struct_copies_default(PyTypeObject const * kind);
--- a/src/construct/construct.c
+++ b/src/construct/construct.c
@@ -24,6 +24,7 @@ static enum result fill_defaults(
 	PyObject * self,
 	Py_ssize_t positional_count
 );
+static struct field_lookup named_field(StructType const * type, PyObject * name);
 static enum result run_post_init(StructType const * type, PyObject * self);
 
 static PyObject * interned_value(StructType const * const type, bool const no_arguments) {
@@ -132,7 +133,7 @@ PyObject * Struct_replace(
 	StructType * const type = struct_type_of(self);
 	Py_ssize_t const change_count = keyword_names != NULL ? PyTuple_GET_SIZE(keyword_names) : 0;
 
-	if (change_count == 0) {
+	if (change_count == 0 && type->struct_options.frozen) {
 		return Py_NewRef(self);
 	}
 
@@ -145,27 +146,35 @@ PyObject * Struct_replace(
 			return NULL;
 		}
 
+		for (Py_ssize_t i = 0; i < change_count; ++i) {
+			if (named_field(type, PyTuple_GET_ITEM(keyword_names, i)).tag != FIELD_LOOKUP_FOUND) {
+				return NULL;
+			}
+		}
+
 		PY_OWNED(values, PyTuple_New(type->struct_field_count));
 
 		if (values == NULL) {
 			return NULL;
 		}
 
-		STRUCT_BEGIN_CRITICAL_SECTION(self);
-
-		for (Py_ssize_t i = 0; i < type->struct_field_count; ++i) {
-			PyTuple_SET_ITEM(values, i, Py_XNewRef(*struct_slot(type, self, i)));
-		}
-
-		STRUCT_END_CRITICAL_SECTION();
+		struct_slots_ref_into(type, self, values, NULL);
 
 		for (Py_ssize_t i = 0; i < type->struct_field_count; ++i) {
 			PyObject * const value = PyTuple_GET_ITEM(values, i);
 
-			if (
-				value != NULL &&
-				PyDict_SetItem(changed, PyTuple_GET_ITEM(type->struct_field_names, i), value) < 0
-			) {
+			if (value == NULL) {
+				continue;
+			}
+
+			PyObject * const name = PyTuple_GET_ITEM(type->struct_field_names, i);
+			int const present = PyDict_Contains(changed, name);
+
+			if (present < 0) {
+				return NULL;
+			}
+
+			if (present == 0 && PyDict_SetItem(changed, name, value) < 0) {
 				return NULL;
 			}
 		}
@@ -195,7 +204,28 @@ PyObject * Struct_replace(
 			return NULL;
 		}
 
-		if (struct_copy_slots_and_dict(type, self, replaced) < 0) {
+		if (Py_TYPE(replaced) != cls) {
+			PyErr_SetString(
+				PyExc_SystemError,
+				"salix internal error: the replace construction returned a different type"
+			);
+
+			return NULL;
+		}
+
+		for (Py_ssize_t i = 0; i < type->struct_field_count; ++i) {
+			PyObject * const value = PyTuple_GET_ITEM(values, i);
+			PyObject * * const slot = struct_slot(type, replaced, i);
+
+			if (value != NULL && *slot == NULL) {
+				*slot = Py_NewRef(value);
+			}
+		}
+
+		PY_MOVABLE(source_dict, NULL);
+		struct_slots_copy_into(type, self, replaced, &source_dict);
+
+		if (source_dict != NULL && struct_dict_copy_merged(source_dict, replaced) < 0) {
 			return NULL;
 		}
 
@@ -208,14 +238,22 @@ PyObject * Struct_replace(
 		return NULL;
 	}
 
-	if (
-		bind_keywords(type, copy, arguments, 0, keyword_names) != RESULT_OK ||
-		struct_copy_slots_and_dict(type, self, copy) < 0
-	) {
+	if (bind_keywords(type, copy, arguments, 0, keyword_names) != RESULT_OK) {
 		return NULL;
 	}
 
-	return run_post_init(type, copy) == RESULT_OK ? py_move(&copy) : NULL;
+	PY_MOVABLE(source_dict, NULL);
+	struct_slots_copy_into(type, self, copy, &source_dict);
+
+	if (run_post_init(type, copy) != RESULT_OK) {
+		return NULL;
+	}
+
+	if (source_dict != NULL && struct_dict_copy_merged(source_dict, copy) < 0) {
+		return NULL;
+	}
+
+	return py_move(&copy);
 }
 
 static enum result run_post_init(StructType const * const type, PyObject * const self) {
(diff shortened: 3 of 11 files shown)


JPHutchins:
Is it possible for replace to be type safe? I assume not. It would need major type checker support e.g. mapping field names to valid Map keys -> field types. It's an OK trade off for now, I just want to make sure that we aren't leaving anything on the table here.

PR #144, opened 2026-08-23
chore: fold the recorded nits from the review convergence rounds

JPHutchins:
The converged rounds of #137/#138/#139/#140 closed with recorded nits, folded here rather than reopened:

- **tasks.py** spelled the venv-creation command twice: the matrix build and the free-threaded leg each wrote it out. `MAKE_ENV` is now the one definition, and the free-threaded bootstrap derives its `when=` guard from the same template: recreate the venv only when it is missing.
- The pep695 lint leaf checks py312-only syntax; it now declares that with a `when=` guard on the running interpreter instead of running the leaf everywhere.
- `clean` destroyed the environments its own matrix reuses. It now exempts `.venvs` and the free-threaded root.
- The comment block above `make_env` existed twice; the duplicate is removed and the surviving copy keeps the build-leaf wording.
- `tests/test_mutability.py` pins the `__orig_class__` accept-and-discard that shipped in #139 without a pin: the write is accepted and nothing is left behind.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt asked to fold the review-round nits recorded at each PR's convergence into one residue PR, gate it, and open it for review.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -1,3 +1,4 @@
+import sys
 from pathlib import Path
 
 from camas import Claude, Config, Parallel, Project, Sequential, Task, by_suffix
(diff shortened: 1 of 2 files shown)


JPHutchins:
Convergence disposition (round 1, score 0.91, converged):

**Systemic `when-guard-scoped-only` — decided, recorded.** The guard only ever skips on scoped runs; full runs (CI `camas ci`, local `camas check`) still re-clear the venv. In CI the checkout is fresh so `--clear` is correct anyway; a local full run paying the clear is the accepted cost. The stated skip delivers exactly where it matters (the scoped agent loop).

**Nits recorded, not chased** (folded into the next residue PR):
- `lint-coverage-hole-scoped-run` — pep695 lint leaf's `when` fires only on 3.12+ interpreters, so a 3.10 gate skips it; low-stakes (CI covers it).
- `venv-rebuild-guard-inert-on-full-runs` — same as the systemic above.
- `stale-venv-guard-no-invalidation` — no invalidation path for a stale 3.14t patch venv now that clean exempts `.venvs`; a stamp file or restoring the clean lever is future work.
- `windows-venv-path-assumption` — `bin/python` is POSIX-only; on Windows the venv leaf always re-runs (correct, just no skip).

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was the code-review convergence protocol: record the round-1 nits at convergence and merge without chasing them.

Issue #119, opened 2026-08-15
Consolidate the body-binding classification into one shared helper


JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was the code-review bot's round-1 systemic on PR #118: the body-binding classification is answered a different way at each refusal site.

**The observation.** "Does this binding belong to this class's own body" is answered four ways: `refuse_mixin_method_fields` type-checks `PyMethodDescr_Type` (namespace.c:460-493), `refuse_colliding_methods` qualname-tail-matches (495-534), the reserved-metadata refusal matches exact names regardless of binding, and #89's carve-out adds a file-static `qualname_of` probe with exactly one caller. The mixin site's type check means a functools-wrapped method placed on the mixin's dict would still silently default — not user-reachable today (`dir(salix)` exports only `Struct` and `set_field`), but each new wrapper family currently requires touching a new per-site special case.

**What a fix looks like.** A shared `binding_belongs_to_body(namespace, name, class_name)` helper — qualname resolution with the same AttributeError discipline, shared across the sites — gives the mixin site wrapper-awareness and keeps the carve-out's boundary expressed once. Consolidation advice; not a user-facing defect.

JPHutchins:
Merged via PR #145 (round 1, score 0.94, converged). The round's minor finding — the refusal set coupling to interpreter descriptor `__qualname__` — was disposed in the PR: the qualname classification is the issue's design, the reviewer's probes confirm it benign on 3.12–3.14, and the parametrised test pin is the loud-failure guard.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was the review-loop merge protocol: close the issue behind the converged PR with the round's disposition.

PR #145, opened 2026-08-23
refactor: one body-binding classifier for every refusal site (#119)

JPHutchins:
**#119** — the body-binding classification, answered once instead of four ways.

## What changed

`binding_belongs_to_body(namespace, name, class_name, spelling)` in `namespace.c` replaces two of the four answers:

- `defines_a_method` (the colliding-methods site's qualname probe) becomes the shared helper — same tri-state, same `spelling` out-param, same AttributeError discipline, now reading its binding from the namespace itself.
- The mixin site's `PyMethodDescr_Type` check becomes a call to the same helper, with the mixin's own `__qualname__` as the class name (a static `PyTypeObject` has no C field for it on any supported interpreter; the runtime attribute is what the interpreter builds the mixin's method-descriptor qualnames from).
- The reserved-metadata refusal is untouched: it refuses *names*, not bindings, which is a different question than the one the helper answers.

The payoff is the issue's: a new wrapper family added to the mixin is recognised at every site by adding one check in one place.

## Measured behavior delta

The classifier also answers for the mixin's wrapper descriptors, whose qualnames name the mixin exactly as the method descriptors' do. Probe against today's build:

<details><summary>accept → refuse (the same "field would shadow the mixin's method" refusal <code>__copy__</code> already gets)</summary>

| field name | before | after |
|---|---|---|
| `__repr__`, `__setattr__`, `__delattr__`, `__lt__`, `__le__`, `__eq__`, `__ne__`, `__gt__`, `__ge__` | accepted | refused |
| `__hash__` | CPython slots `ValueError` | salix's mixin refusal (earlier, clearer) |
| `__copy__`, `__deepcopy__`, `__replace__` | refused | refused (unchanged) |
| `__doc__`, the eight metadata names | as before | unchanged |

</details>

The shadowing the refusal describes is the same hazard under those names — a field named `__eq__` silently replaces the mixin's structural equality for attribute lookup — so the refusal now covers the whole mixin-carried set. `test_methods.py` pins the full set parametrically.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was issue #119's brief: consolidate the body-binding classification into one shared helper, measure the mixin-site delta before committing to it, pin it, and open the PR for review.

### DIFF
--- a/src/meta/namespace.c
+++ b/src/meta/namespace.c
@@ -46,9 +46,9 @@ static enum result apply_options(
 );
 static enum result rebind(PyObject * namespace, char const * const * names, bool from_mixin);
 static enum result drop_class_variables(PyObject * namespace, PyObject * all_names);
-static int defines_a_method(
-	PyObject * bound,
-	PyObject * field_name,
+static int binding_belongs_to_body(
+	PyObject * namespace,
+	PyObject * name,
 	PyObject * class_name,
 	PyObject * * spelling
 );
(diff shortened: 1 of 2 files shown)


JPHutchins:
Convergence disposition (round 1, score 0.94, converged):

**Minor `mixin-refusal-set-version-coupling` — decided by the issue's brief, recorded.** #119 directs the qualname classification; the refusal set deriving from the interpreter's descriptor `__qualname__` contents is the design, not an accident. The reviewer's own probes confirm it is benign on 3.12–3.14 (descriptor qualnames present and consistent via PyType_GetQualName on both sides). The parametrised pin in test_methods.py is the guard that makes any future interpreter change fail loudly, which is the structural answer the finding's C-predicate alternative would only relocate.

**Hidden nits recorded, not chased** (folded into the next residue PR):
- `double-dict-fetch-dead-branch` — the refusal-path refetch's NULL-without-error `continue` is unreachable now that the helper owns the fetch.
- `unused-spelling-out-param` — the mixin site passes `&spelling` it never reads; the out-param serves the colliding site's unmangle logic.
- `mixin-gate-widening-descriptor-branch` — the property/classmethod/staticmethod branch now also gates the mixin site; harmless today, part of "one classifier".

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was the code-review convergence protocol: record the round-1 findings at convergence and merge without chasing them.

PR #146, opened 2026-08-23
chore(tasks): structured gate diagnostics for the fixer ladder

JPHutchins:
The checking leaves now declare `agent_format`, so a failing gate hands the camas fixer subagents machine-readable diagnostics instead of tailed text:

- **lint** emits ruff rdjson — the gate's diagnostic payload carries each finding's rule, exact range, and suggestion, which is what the lint-fixer ladder reads.
- **pytest** (the six matrix legs and the free-threaded leg) write junit through `{report}` path mode — the test-fixer tier gets per-test results instead of a traceback tail.

The flags are appended only on gate runs; human and CI runs keep the commands as they were. Verified live before committing: a scoped gate on this very diff failed the lint leaf and returned `output_kind: rdjson` with the full ruff payload including the import-fix suggestion; after applying it the same gate is green with all junit leaves running clean.

## Rounds 1–3 disposition

**Streak-3 findings, deferred to camas (JPH's ruling: no camas changes):** `structured-payload-stderr-merged`, `path-mode-report-empty-on-crash`, `structured-kind-mislabeled-errored-leaf`, systemic `structured-diagnostics-camas-side-deferred`. These describe camas gate behavior (stderr mixing, the path-mode fallback on a crashed leg, the output-kind label on the fallback), not salix configuration. The salix-side mitigations are in: the raised limits mean the pointer fallback fires only past a full suite's junit or a hundred-plus-diagnostic rdjson, which is exactly the case where the payload must survive anyway.

**Fixed in the rounds:** the limits (round 1/2: junit 1_000_000, rdjson 64_000, named constants — `FULL_SUITE_JUNIT`, `HUNDREDS_OF_DIAGNOSTICS` — rather than comments, per the repo's no-comments rule); the hoisted `RUFF_CHECK` fragment and the pep695 leaf's file-keyed `when=` guard with the whole-repo leaf's exclude (round 3), so a scoped gate lints the file once and a full run targets it once at py312.

**Declined with reason:** pinning the ruff version (round 1's stderr-mitigation suggestion) — pinning a linter version is a project policy change, out of scope for this PR; recorded here instead.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was JPH's correction that the camas agent-facing options (agent_format and Config.agent) are capabilities to USE in the salix workflow; this PR wires the two diagnostic-rich leaves into them, learned from camas_docs and the installed fixer-agent definitions, and the rounds-1–3 disposition follows the review protocol.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -56,10 +56,11 @@ STRICT_BUILD = {"SALIX_STRICT": "1"}
 NIX_INPUTS = ("src/", "nix/", "tools/", "tests/", "flake.nix", "flake.lock", "pyproject.toml")
 
 build = Sequential(make_env, Task(BUILD, mutates=True, env=STRICT_BUILD))
+JUNIT_FORMAT = AgentFormat("--junitxml {report}", "junit", limit=1_000_000)
 pytest = Task(
     PYTEST,
     env=ENVIRONMENT_PER_INTERPRETER,
-    agent_format=AgentFormat("--junitxml {report}", "junit"),
+    agent_format=JUNIT_FORMAT,
 )
 # --no-sync: analyze reaches this from ci, and CI never installs the project.
 compile_flags = Task("uv run --no-sync python tools/compile_flags.py", mutates=True)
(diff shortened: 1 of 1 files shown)


JPHutchins:
Convergence disposition (round 5, score 0.24, converged):

- `junit-limit-sized-for-passing-suite` (minor) — recorded: the 1_000_000 limit was sized from a passing full-suite junit; a failing leg's tracebacks ride in the same document. The pointer fallback past the limit is the documented safety, and a >1MB failing junit is unlikely (a handful of ~2KB tracebacks); noted for the next residue fold rather than chased.
- `exclude-flag-redundant-with-pyproject` (nit) — recorded: the CLI extend-exclude could live in pyproject's ruff exclude instead; kept on the leaf so the exclusion is visible beside the leaf that needs it.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was the code-review convergence protocol: record the round-5 findings at convergence and merge.

PR #147, opened 2026-08-23
feat: ClassVar-annotated class variables are kept, not refused (#134)

### LINKED ISSUE #134: Support ClassVar-annotated class variables
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. She asked whether salix supports ClassVar, learned it is deliberately refused (#50) with the unannotated-binding workaround, and asked for this long-term goal to be tracked: ClassVar as the typed way to put scoped constants in an immutable static location, with full type checking and no performance penalty.

## Executive summary

`x: ClassVar[int]` in a struct body is refused today (`'x' is annotated ClassVar, which salix does not support; write it below the fields, without an annotation` — the #50 ruling). The workaround is an unannotated class-body assignment, which stays in the class dict untouched. The long-term goal is to accept the annotated spelling and make it the typed equivalent of that workaround.

## What support means

- A `ClassVar`-annotated name is **not a field**: excluded from `__struct_fields__`, `__struct_defaults__`, `__struct_annotations__`/`__struct_metadata__`, the slots, and the constructor. Exactly what the unannotated path does today, but recognized from the annotation.
- The value stays an **ordinary class attribute** — the immutable static location, readable on the class and every instance, no per-instance storage and no per-instance cost (the unannotated path today already costs nothing; the ClassVar check runs once at class creation).
- **Full type checking**: the stub declares the behavior so mypy, pyright, and ty accept `x: ClassVar[T]` (and reject its use as a field) — `tests/typing/accepted.py`/`rejected.py` entries to match, alongside the `dataclass_transform` surface.
- Defaults: a `ClassVar` with an assignment keeps it in the class dict (the dataclass rule — a ClassVar without a default is also legal in dataclasses; decide whether salix requires one, since the class dict is where the value must live).

## Constraints the implementation must preserve

- The reserved-metadata names refuse even as `ClassVar` — the existing pin says the reservation outranks the ClassVar advice.
- `InitVar`'s refusal is separate and stays.
- The "a field is what the annotations say" rule stays coherent: `special_form_of` already probes ClassVar for the refusal; support means the probe's answer becomes "exclude and keep" instead of "refuse".

## Related

- #50 (the refusal ruling), #59 (the annotation retention this builds on), the ClassVar pins in `tests/test_special_forms.py` and `tests/test_metadata_names.py`.

JPHutchins:
**#134** — `x: ClassVar[int] = 5` becomes the typed equivalent of the unannotated class-body assignment, per JPH's ruling on the issue (no default → refused: a class variable is a constant).

## What changed

The probe's answer flips: wherever `special_form_of` answers ClassVar (object path or text path), the name is **excluded from the field plan instead of refused** — the binding stays in the class dict, and the constructor, slots, `__match_args__`, and every metadata table exclude it. The shared-mutable refusal no longer applies to it: that check is a field rule, and the ClassVar annotation declares the sharing.

## Decisions, and where they come from

<details><summary>Rulings and semantics table</summary>

| shape | before | after |
| --- | --- | --- |
| `x: ClassVar[T] = value` | refused | kept, not a field |
| `x: ClassVar` / no value | refused (unsupported) | refused: "without an assigned value; a class variable is a constant, so assign the value in the class body" |
| `x: ClassVar[list] = [1]` (shared mutable) | refused (shared-mutable) | kept — the unannotated path's semantics |
| reserved metadata name as ClassVar | reserved refusal | reserved refusal (outranks) |
| subclass re-annotating an inherited field | no-op | no-op (inheritance check runs first) |
| `InitVar` | refused | refused (unchanged) |
| ClassVar nested in another annotation | refused | **excluded-and-kept** — the issue's "the probe's answer becomes exclude and keep" applied literally, not top-level-only. The with-default nested shape is unpinned either way. |

</details>

## Verified

- Live probe: fields/defaults/annotations/`__match_args__` all exclude the name; class and instance attribute reads agree; the no-default and reserved-outrank refusals fire with the expected messages.
- Full gate green: pytest 3.10–3.15 + free-threaded, c_test, analyzers, ruff, and mypy/pyright/ty — no stub change needed: PEP 681's `dataclass_transform` already teaches all three checkers that a ClassVar-annotated name is not a field; `accepted.py` pins it.
- All existing refusal pins survive the flip: every nested/bare shape in `test_special_forms.py` and `test_classvar_paths.py` still refuses, via the no-default rule.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was JPH's "what are the chances of taking this on" for #134; the scope assessment (high — the machinery already implements the target semantics) led to this implementation.

closes #134

### DIFF
--- a/src/fields.h
+++ b/src/fields.h
@@ -6,14 +6,15 @@
 #include "types.h"
 
 struct field_plan {
-	PyObject * all_names;     /* list[str] */
-	PyObject * new_names;     /* list[str] */
-	PyObject * defaults;      /* tuple */
-	PyObject * annotations;   /* tuple[Any], aligned with all_names */
-	PyObject * metadata;      /* tuple[tuple[Any, ...]], aligned with all_names */
+	PyObject * all_names;        /* list[str] */
+	PyObject * new_names;        /* list[str] */
+	PyObject * class_var_names;  /* list[str] */
+	PyObject * defaults;         /* tuple */
+	PyObject * annotations;      /* tuple[Any], aligned with all_names */
+	PyObject * metadata;         /* tuple[tuple[Any, ...]], aligned with all_names */
 };
 
-/* Owns its five references.  On failure every member is NULL and an
+/* Owns its six references.  On failure every member is NULL and an
  * exception is set. */
 struct field_plan field_plan_build(StructType const * base, PyObject * namespace);
 
--- a/README.md
+++ b/README.md
@@ -17,23 +17,27 @@ Point(x=1.0, y=2.0)
 
 ## ClassVar and InitVar
 
-A field annotated `ClassVar[...]` or `InitVar[...]` is refused at class
-creation when the annotation reaches the check as one of those forms. The
-check walks the annotation's forms when it arrives as an object; when it
-arrives as source text — a quoted string, or a future-annotations string —
-the check is a spelling match, and the two paths differ at the edges:
+A name annotated `ClassVar[...]` at the top level is a class variable:
+with an assigned value it is kept in the class dict and excluded from
+the field plan, the constructor, and every metadata table; without one
+it is refused, because a class variable is a constant. `InitVar[...]`
+is refused, and so is a `ClassVar` nested inside another annotation.
+The check walks the annotation's forms when it arrives as an object;
+when it arrives as source text — a quoted string, or a
+future-annotations string — the check is a spelling match, and the two
+paths differ at the edges:
 
 | annotation | object path | text path |
 | --- | --- | --- |
-| an alias of `ClassVar`, `CV[int]` | refused | accepted; the field swallows a positional |
+| an alias of `ClassVar`, `CV[int]` | the form itself: kept with a value, refused without | accepted; the field swallows a positional |
 | a type of yours actually named `ClassVar` | accepted | refused |
 | `Annotated[int, ClassVar]` | accepted | refused, though the `ClassVar` there is metadata |
 | `Optional[Annotated[ClassVar[int], "m"]]` | refused | refused |
 
-3.14 evaluates resolvable annotations by default, so there the alias reaches
-the object path and is refused; a name whose root cannot be resolved
-arrives as an object too, and is accepted, while any other evaluation
-error still fails the class. Every row is pinned in
+3.14 evaluates resolvable annotations by default, so there the alias
+reaches the object path and is the form; a name whose root cannot be
+resolved arrives as an object too, and is accepted, while any other
+evaluation error still fails the class. Every row is pinned in
 `tests/test_classvar_paths.py`.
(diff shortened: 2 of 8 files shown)


JPHutchins:
closes #134

src/fields.h
```diff
@@ -6,14 +6,15 @@
...
-	PyObject * new_names;     /* list[str] */
-	PyObject * defaults;      /* tuple */
-	PyObject * annotations;   /* tuple[Any], aligned with all_names */
-	PyObject * metadata;      /* tuple[tuple[Any, ...]], aligned with all_names */
+	PyObject * all_names;        /* list[str] */
+	PyObject * new_names;        /* list[str] */
+	PyObject * class_var_names;  /* list[str] */
+	PyObject * defaults;         /* tuple */
+	PyObject * annotations;      /* tuple[Any], aligned with all_names */
+	PyObject * metadata;         /* tuple[tuple[Any, ...]], aligned with all_names */
```
JPHutchins, line 14: this aligned woth nonsesne should be strucutral

JPHutchins:
Convergence disposition (round 6, score 0.95, converged):

**Systemic `text-path-spelling-heuristic` — decided.** The spelling heuristic is the design: CPython exposes no public C tokenizer, and feeding the text through Python's tokenize costs a stdlib import on the text path, which the zero-startup budget exists to avoid. Its boundaries are the documented ones: the README table rows and the pinned holes (renamed import, attribute-qualified spelling).

**Findings recorded, not chased:**
- `text-path-qualified-attribute-classvar-kept` — the r5 commit message overclaimed; an attribute-qualified spelling (`x.ClassVar[int]`) is kept with a value, the mirror of the renamed-import hole. Corrected here.
- `dunder-classvar-over-refusal` — the broad dunder rule is round 5's own recommendation ("every __dunder__ binding shadows a slot or protocol lookup"); the empirical narrowing is a future residue.
- `mixin-classvar-branch-unreachable` — recorded cleanup: with the dunder shape test, the class_var_names half of refuse_mixin_method_fields is unreachable today; drop it in a future residue.
- `inherited-required-classvar-silent-default` — the inheritance rule outranks the ClassVar annotation (documented in the README); the required-becomes-defaulted nuance is that rule's consequence.
- `nested-classvar-mutable-diagnosis-lost` — a nested ClassVar with a shared-mutable default now reports the unsupported message rather than the mutable diagnosis; recorded as a diagnosis-order note.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was the code-review convergence protocol: decide the systemic, record the round-6 findings, and merge.

PR #150, opened 2026-08-29
chore: fold the recorded nits from the #145/#147 convergence rounds

JPHutchins:
The convergence dispositions from #145 and #147, folded:

- **Dropped the unreachable mixin ClassVar branch** — the dunder shape test in the plan loop fires before the mixin guard for every mixin method name, so `class_var_names` (plan member, parameter, the ClassVar message branch) was dead. If a non-dunder mixin method ever appears, the guard needs restoring for that name — recorded in the commit rather than in dead code.
- **Dropped the dead NULL-without-error `continue`** on `refuse_colliding_methods`' refusal-path refetch — a belonging verdict implies the helper held the binding, so a NULL refetch can only be a failure.
- **Restored the shared-mutable diagnosis** for a nested ClassVar with a shared-mutable default (the message it had before the flip), pinned.

Full gate green: 3.10–3.15, free-threading, c_test, analyzers, all type checkers.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was the review-loop queue: fold the nits recorded at the converged rounds' dispositions into a residue PR.

### DIFF
--- a/src/fields.c
+++ b/src/fields.c
@@ -53,7 +53,6 @@ static enum result append_declared(
 	PyObject * namespace,
 	PyObject * all_names,
 	PyObject * new_names,
-	PyObject * class_var_names,
 	PyObject * default_by_name,
 	PyObject * annotation_values,
 	PyObject * metadata_values,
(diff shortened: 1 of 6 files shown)


JPHutchins:
Convergence disposition (round 1, score 0, converged):

- `error-return-without-exception` — recorded: the NULL-refetch `return RESULT_ERROR` can return without an exception in the (unreachable) NULL-without-error case. The pre-existing shape (separate the PyErr check) is the codebase's error discipline; folded into the next residue as a three-line restore rather than held here.
- `side-effect-hash-on-refused-path` — recorded: the restored shared-mutable diagnosis hashes the default on the nested-refusal path, and a foreign exception propagates exactly as it did before the ClassVar flip; the diagnosis is the pre-PR behavior, side effect included.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was the code-review convergence protocol: record the round-1 nits and merge.

PR #151, opened 2026-08-29
fix: frozen writes raise AttributeError, matching every frozen thing (#24)

JPHutchins:
**#24** — writing to a frozen struct now raises `AttributeError`, matching a frozen dataclass, a read-only property, a namedtuple field, and a slotted class — the exception an `except AttributeError:` is written for. The message also quotes the type name the way CPython does everywhere else.

## Scope

- `Struct_set_attribute`'s refusal: `PyExc_TypeError` → `PyExc_AttributeError` for both assignment and deletion; message becomes `'P' object does not support attribute assignment`.
- The `__orig_class__` carve-out is untouched.
- Twenty pins flip in the suite. Where 3.10–3.12 already split the escaped-shape tests (the slot wrapper's own `can't apply this __setattr__` TypeError for the body-hook half), the split is per operation now: the half salix answers is the new `AttributeError` on every version, the slot half stays `TypeError`.
- The base-combinations classifier (`observe`) catches `AttributeError` for the frozen column.

## Noted, not this PR

`test_base_combinations.py` has a `strict=False` xfail (`test_a_single_base_asking_to_drop_an_inherited_weakref_slot_is_the_same_bug`) that xpasses — its marker says to flip it when the behavior changes; the weakref refusal is class-creation machinery this PR never touches, so it stays as-is.

Full gate green: 3.10–3.15, free-threading, c_test, analyzers, all type checkers.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was the review-loop queue: take #24 from the open-issue list, implement its table's ruling, and drive it through the loop.

### DIFF
--- a/src/mixin.c
+++ b/src/mixin.c
@@ -702,8 +702,8 @@ static int Struct_set_attribute(
 	}
 
 	PyErr_Format(
-		PyExc_TypeError,
-		"%.200s object does not support attribute %s",
+		PyExc_AttributeError,
+		"'%.200s' object does not support attribute %s",
 		Py_TYPE(self)->tp_name,
 		value == NULL ? "deletion" : "assignment"
 	);
--- a/tests/test_base_combinations.py
+++ b/tests/test_base_combinations.py
@@ -160,7 +160,7 @@ def observe(cls: type) -> Behaviour:
     try:
         setattr(instance(cls), cls.__struct_fields__[0], 99)
         frozen = False
-    except (TypeError, IndexError):
+    except (AttributeError, IndexError):
         frozen = True
 
     return Behaviour(
--- a/tests/test_custom_init.py
+++ b/tests/test_custom_init.py
@@ -209,7 +209,7 @@ def test_a_frozen_struct_with_a_body_init_cannot_write_its_fields():
         def __init__(self) -> None:
             self.x = 1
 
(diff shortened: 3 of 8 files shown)


JPHutchins:
Convergence disposition (round 1, score 0, converged):

- `classifier-catch-removed` — **flagged for the next residue**: observe()'s write probe should mirror deletes() with `(TypeError, AttributeError, IndexError)` so a future combination hitting the 3.10–3.12 slot wrapper cannot misclassify frozen as mutable. The suite passes today, so it is not load-bearing yet.
- `unvalidated-name-slot-arg` — **flagged for the next residue**: guard the `__orig_class__` carve-out with `PyUnicode_Check` + `PyUnicode_EqualToUTF8`; the unchecked compare on a non-string name is UB-shaped, pre-dating this PR (shipped in #139).
- `untested-exception-contract`, `duplicated-version-gate`, `unpinned-message-format` — recorded: C-test exception pins, the gate's collapse, and a single quoted-format pin are polish; the Python suite already pins the contract end-to-end.

> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins. The prompt was the code-review convergence protocol: record the round-1 nits at convergence and merge without a nit-fix push.

Issue #160, opened 2026-08-30
Proof experiment: replace dataclass with salix.Struct in transformers and benchmark the fork


JPHutchins:
## Executive summary

Prove the dataclass drop-in claim at scale: fork a popular library, mechanically replace every `@dataclass` with `salix.Struct` behind a small shim, run the library's own test suite on the fork, and A/B-benchmark import and runtime against stock.

**The candidate is [huggingface/transformers](https://github.com/huggingface/transformers):** 947 `@dataclass` definitions measured (864 in shipped source across 430 files), a ~90% salix-compat envelope, and it is the canonical slow-import library — hundreds of ModelOutput/config classes created at import. **Repo shape: a standalone fork pulled in as a nix flake input, not a git submodule.**

<details>
<summary>Measured census (shallow clones, AST census cross-checked with ripgrep; every sha cited)</summary>

| repo | git sha | py files | @dataclass defs | src | frozen | slots | kw_only | `__post_init__` | InitVar | ClassVar | cached_prop |
| --- | --- | --: | --: | --: | --: | --: | --: | --: | --: | --: | --: |
| huggingface/transformers | `42ca970` | 4809 | **947** | 864 | 10 | 0 | 0 | 37 | 0 | 11 | 1 |
| dagster-io/dagster | `fc0d15b` | 6234 | 225 | 150 | 44 | 0 | 0 | 4 | 1 | 0 | 35 |
| google/flax | `01854da` | 353 | 173 (+26 struct) | 113 | 54 | 33 | 0 | 38 | 0 | 0 | 0 |
| apache/airflow | `b072197` | 7808 | 138 | 112 | 28 | 4 | 0 | 13 | 1 | 9 | 24 |
| facebookresearch/hydra | `d1e07c8` | 457 | 226 | 97 | 3 | 0 | 0 | 5 | 2 | 0 | 0 |
| omry/omegaconf | `a8bcf1f` | 151 | 417 | 2 | 6 | 0 | 0 | 4 | 0 | 0 | 0 |
| brentyi/tyro | `d0c9877` | 313 | 2749 | 51 | 399 | 0 | 0 | 16 | 0 | 14 | 1 |

[...]

PR #162, opened 2026-09-01
bench: dataclass-compat harness — omegaconf patch tier for #160

JPHutchins:
bench: dataclass-compat harness — omegaconf patch tier for #160

> [!WARNING]
> **LLM Disclosure**
>
> This PR was authored by deepseek-v4-pro on behalf of @JPHutchins, following her #160 plan to prove out dataclasses replacement on consumer suites before the release.

Executive summary: a reproducible consumer-tier harness for #160. `bench/dataclass-compat/_shim.py` patches the `dataclasses` module so `@dataclass` builds salix Structs behind the dataclasses API, and `run_omegaconf.sh` runs omegaconf's own test suite (unmodified, pinned at `a8bcf1f`) against it.

Measured tally: **8473 passed, 40 failed, 362 skipped, 1 module excluded** (python 3.13, salix 0.1.0 wheel built from this repo). Every failure and the exclusion are documented in `known_gaps.md`.

<details>
<summary>Known gaps (salix-level, not shim-level)</summary>

- 37 pickle failures — salix refuses pickle by design (#13, lowest priority per JPH)
- 3 failures — structs have no instance `__dict__`, so arbitrary attribute assignment is impossible (omegaconf's `to_container` setattr path)
- 1 module excluded — `Child(Left, Right)` with two unrelated Struct bases raises CPython's layout conflict at statement time, before any decorator can intercept
</details>

<details>
<summary>What the suite forced out of the shim</summary>

- statement-time StructMeta binding precedes the decorator → struct subclasses rebuild with namespace recovery
- `field(default_factory=...)` → `_PLACEHOLDER` default + injected `__post_init__` (per-instance fill, dataclasses-exact)
- `field(init=False)` → exec-synthesized `__init__` (salix's body-init handling disables the auto `__post_init__` call)
- unset member descriptors fall back to class attributes (the factory design leans on this)
- namespace `__setattr__` overrides the C slot and enforces frozen
- static-type bases (e.g. `dict`) segfault salix's layout in GC → stock-dataclasses fallback
</details>

Reproducibility: the runner pins the omegaconf revision, the frozen requirement set (`requirements.txt`), and the salix wheel; ANTLR generation uses omegaconf's own `setup.py antlr` (committed 4.11.1 jar). Verified end-to-end from a clean workdir with an identical tally.

Part of the #160 plan: patch tier (this PR) + source tier (omegaconf fork with salix-native rewrites, to follow).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

## Systemic disposition (streak 3: benchmark-harness-untested)

The reviewer's systemic finding — the shim ships with no self-tests and the harness is not wired into CI — is acknowledged and **deliberately accepted** at this stage, with a plan:

- The shim's test suite IS three real consumer suites: omegaconf (8473 passing), tyro (5062), hydra (3264 at stock parity). Each is reproduced by a pinned runner in this PR (`run_omegaconf.sh`, `run_tyro.sh`, `run_hydra.sh`) plus the transformers import smoke with fields() parity on all 782 dataclasses.
- Those suites found 30+ real shim defects over the three review rounds, which is the evidence that the coverage is real, not a snapshot.
- CI wiring (a camas task running the runners on a schedule) is planned once the harness stabilizes; it is not part of this PR because the consumer checkouts are large and the CI emit deserves its own change.

If this disposition is not acceptable, the alternative is a dedicated `tests/test_shim.py` covering the shim's public surface; that is a well-scoped follow-up issue.

> [!WARNING]
> **LLM Disclosure**
>
> This update was authored by deepseek-v4-pro on behalf of @JPHutchins, disposing of the round-3 streak-3 systemic finding in the PR body per the review protocol.


JPHutchins:
> [!WARNING]
> **LLM Disclosure**
>
> This post was authored by deepseek-v4-pro on behalf of @JPHutchins, disposing of three round-2 findings by measurement before the round-3 push.

Three findings were re-checked against the measured behavior of stock CPython 3.13 and salix 0.1.0:

**`post-init-bypassed-by-own-init` — the stock premise is inverted.** Measured on stock 3.13:

```python
@dataclass
class StockBodyInit:
    x: int = 1
    def __init__(self, x=2):
        calls.append("init"); self.x = x
    def __post_init__(self):
        calls.append("post_init")

StockBodyInit()  # calls == ['init'], x == 2 — the BODY init ran, post_init did not
```

Stock keeps a body `__init__` (its `init` parameter is documented as ignored in that case) and never calls `__post_init__`; the shim matches. The known_gaps.md bullet stands.

**`asdict-deepcopy-nested-struct` — deepcopy does not raise** (salix implements `__deepcopy__`; the omegaconf suite's deepcopy tests pass with struct metadata inside the copied graph). The nested-divergence half was real and is fixed anyway: `asdict` now recurses into nested struct values like stock.

**`init-false-without-default-unconstructable` — the fill_defaults claim doesn't hold for injected-init classes.** Measured: a salix struct with a required field and a body `__init__` constructs fine and `getattr` raises `AttributeError` — exactly stock's `field(init=False)`-without-default behavior. The shim's exec-synthesized init is a body init, so no placeholder default is needed (adding one made the attribute `hasattr`-visible and broke omegaconf's `test_to_container` assertion; reverted).

The remaining round-2 findings are fixed in the round-3 commit: empty-mutable defaults factory-ized, inherited-factory placeholder on the rebuild, early-return option comparison, registry harvest includes the class itself, `__match_args__`/weakref alignment, `__dataclass_self__` receiver rename, per-field repr/compare/hash raises, `replace()` reconstructs via the constructor (post_init re-runs, init=False changes rejected), and the harness now also runs tyro's suite (5062 passed / 30 failed, all documented).

